PR #96, opened 2014-12-23
Revert "Do not re-buffer sources and sinks."

JakeWharton:
This reverts commit 383e5c6903ac8c19fb3651f89cc5b1c2fb7be05e.


JakeWharton:
Yep. And ultimately you can do the `instanceof` check yourself if you *really* want to avoid the extra layering. In practice the multiple-buffering only happens with certain usage patterns. Reverting this change restores the principle of least surprise with very little actual cost.

Issue #97, opened 2014-12-24
Native read/write primitive to strings.


JakeWharton:
`int` and `long` with radix. Any others? Naming convention?

PR #98, opened 2014-12-25
Add support for reading and writing longs as stringy values.

JakeWharton:
Refs #97.

Use it. Or don't. Was good fun writing/exploring the many ways to do these things.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -45,6 +45,9 @@ import static okio.Util.reverseBytesLong;
  * This class avoids zero-fill and GC churn by pooling byte arrays.
  */
 public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
+  private static final byte[] DIGITS =
+      { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
+
   Segment head;
   long size;
 
@@ -405,6 +408,119 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public long readLongLe() {
     return Util.reverseBytesLong(readLong());
+  }
+
+  @Override public long readDecimalLong() {
+    if (size == 0) throw new IllegalStateException("size == 0");
+
+    // This value is always built negatively in order to accommodate Long.MIN_VALUE.
+    long value = 0;
+    int seen = 0;
+    boolean negative = false;
+    boolean done = false;
+
+    long overflowZone = Long.MIN_VALUE / 10;
+    long overflowDigit = (Long.MIN_VALUE % 10) + 1;
+
+    do {
+      Segment segment = head;
+
+      byte[] data = segment.data;
+      int pos = segment.pos;
+      int limit = segment.limit;
+
+      for (; pos < limit; pos++, seen++) {
+        byte b = data[pos];
+        if (b >= '0' && b <= '9') {
+          int digit = '0' - b;
+
+          // Detect when the digit would cause an overflow.
+          if (value < overflowZone || value == overflowZone && digit < overflowDigit) {
+            Buffer buffer = new Buffer().writeDecimalLong(value).writeByte(b);
+            if (!negative) buffer.readByte(); // Skip negative sign.
+            throw new NumberFormatException("Number too large: " + buffer.readUtf8());
+          }
+          value *= 10;
+          value += digit;
+        } else if (b == '-' && seen == 0) {
+          negative = true;
+          overflowDigit -= 1;
+        } else {
+          if (seen == 0) {
+            throw new NumberFormatException(
+                "Expected leading [0-9] or '-' character but was 0x" + Integer.toHexString(b));
+          }
+          // Set a flag to stop iteration. We still need to run through segment updating below.
+          done = true;
+          break;
+        }
+      }
+
+      if (pos == limit) {
+        head = segment.pop();
+        SegmentPool.INSTANCE.recycle(segment);
+      } else {
+        segment.pos = pos;
+      }
+    } while (!done && head != null);
+
+    size -= seen;
+    return negative ? value : -value;
+  }
+
+  @Override public long readHexadecimalUnsignedLong() {
+    if (size == 0) throw new IllegalStateException("size == 0");
+
+    long value = 0;
+    int seen = 0;
+    boolean done = false;
+
+    do {
+      Segment segment = head;
+
+      byte[] data = segment.data;
+      int pos = segment.pos;
+      int limit = segment.limit;
+
+      for (; pos < limit; pos++) {
+        int digit;
+
+        byte b = data[pos];
+        if (b >= '0' && b <= '9') {
+          digit = b - '0';
+        } else if (b >= 'a' && b <= 'f') {
+          digit = b - 'a' + 10;
+        } else if (b >= 'A' && b <= 'F') {
+          digit = b - 'A' + 10; // We never write uppercase, but we support reading it.
+        } else {
+          if (seen == 0) {
+            throw new NumberFormatException(
+                "Expected leading [0-9a-fA-F] character but was 0x" + Integer.toHexString(b));
+          }
+          // Set a flag to stop iteration. We still need to run through segment updating below.
+          done = true;
+          break;
+        }
+
+        if (++seen > 16) {
+          Buffer buffer = new Buffer().writeHexadecimalUnsignedLong(value).writeByte(b);
+          throw new NumberFormatException("Number too large: " + buffer.readUtf8());
+        }
+
+        value <<= 4;
+        value |= digit;
+      }
+
+      if (pos == limit) {
+        head = segment.pop();
+        SegmentPool.INSTANCE.recycle(segment);
+      } else {
+        segment.pos = pos;
+      }
+    } while (!done && head != null);
+
+    size -= seen;
+    return value;
   }
 
   @Override public ByteString readByteString() {
@@ -773,6 +889,81 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public Buffer writeLongLe(long v) {
     return writeLong(reverseBytesLong(v));
+  }
+
+  @Override public Buffer writeDecimalLong(long v) {
+    if (v == 0) {
+      // Both a shortcut and required since the following code can't handle zero.
+      return writeByte('0');
+    }
+
+    boolean negative = false;
+    if (v < 0) {
+      v = -v;
+      if (v < 0) { // Only true for Long.MIN_VALUE.
+        return writeUtf8("-9223372036854775808");
+      }
+      negative = true;
+    }
+
+    // Binary search for character width which favors matching lower numbers.
+    int width = //
+          v < 100000000L
+        ? v < 10000L
+        ? v < 100L
+        ? v < 10L ? 1 : 2
+        : v < 1000L ? 3 : 4
+        : v < 1000000L
+        ? v < 100000L ? 5 : 6
+        : v < 10000000L ? 7 : 8
+        : v < 1000000000000L
+        ? v < 10000000000L
+        ? v < 1000000000L ? 9 : 10
+        : v < 100000000000L ? 11 : 12
+        : v < 1000000000000000L
+        ? v < 10000000000000L ? 13
+        : v < 100000000000000L ? 14 : 15
+        : v < 100000000000000000L
+        ? v < 10000000000000000L ? 16 : 17
+        : v < 1000000000000000000L ? 18 : 19;
+    if (negative) {
+      ++width;
+    }
+
+    Segment tail = writableSegment(width);
+    byte[] data = tail.data;
+    int pos = tail.limit + width; // We write backwards from right to left.
+    while (v != 0) {
+      int digit = (int) (v % 10);
+      data[--pos] = DIGITS[digit];
+      v /= 10;
+    }
+    if (negative) {
+      data[--pos] = '-';
+    }
+
+    tail.limit += width;
+    this.size += width;
+    return this;
+  }
+
+  @Override public Buffer writeHexadecimalUnsignedLong(long v) {
+    if (v == 0) {
+      // Both a shortcut and required since the following code can't handle zero.
+      return writeByte('0');
+    }
+
+    int width = Long.numberOfTrailingZeros(Long.highestOneBit(v)) / 4 + 1;
+
+    Segment tail = writableSegment(width);
+    byte[] data = tail.data;
+    for (int pos = tail.limit + width - 1, start = tail.limit; pos >= start; pos--) {
+      data[pos] = DIGITS[(int) (v & 0xF)];
+      v >>>= 4;
+    }
+    tail.limit += width;
+    size += width;
+    return this;
   }
 
   /**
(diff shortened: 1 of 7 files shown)


JakeWharton:
Hopefully getting back to this soon.

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -233,6 +233,16 @@ public RealBufferedSource(Source source) {
     return buffer.readLongLe();
   }
 
+  @Override public long readDecimalLong() throws IOException {
+    request(20);
```
JakeWharton: @swankjesse Is this the right way to do this? I have to request the maximum width that a decimal long could be.

swankjesse: This isn't particularly good. Imagine you're reading from a blocking stream that offers the following data immediately:

```
1 abc
```

Subsequent data isn't available immediately. But you don't need it; you only need 2 chars. With the `request(20)` call, you'll block needlessly. For interactive protocols, that unnecessary blocking could even cause you to hang.

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,68 @@ public byte getByte(long pos) {
...
+      for (; pos < limit; pos++) {
+        int digit;
+
+        byte b = data[pos];
+        if (b >= '0' && b <= '9') {
+          digit = b - '0';
+        } else if (b >= 'a' && b <= 'f') {
+          digit = b - 'a' + 10;
+        } else if (b >= 'A' && b <= 'F') {
+          digit = b - 'A' + 10; // We never write uppercase, but we support reading it.
```
swankjesse: Nice comment.

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,68 @@ public byte getByte(long pos) {
...
+        } else {
+          done = true;
+          break;
+        }
+
+        if (++seen > 16) {
+          Buffer buffer = new Buffer().writeHexadecimalUnsignedLong(value).writeByte(b);
```
swankjesse: Cute.

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,68 @@ public byte getByte(long pos) {
...
+        byte b = data[pos];
+        if (b >= '0' && b <= '9') {
+          digit = b - '0';
+        } else if (b >= 'a' && b <= 'f') {
+          digit = b - 'a' + 10;
+        } else if (b >= 'A' && b <= 'F') {
+          digit = b - 'A' + 10; // We never write uppercase, but we support reading it.
+        } else {
+          done = true;
+          break;
```
swankjesse: use a labeled break, and drop the `done` nonesense!

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,68 @@ public byte getByte(long pos) {
...
+
+      if (pos == limit) {
+        head = segment.pop();
+        SegmentPool.INSTANCE.recycle(segment);
+      } else {
+        segment.pos = pos;
+      }
+    } while (!done && head != null);
+
+    if (seen == 0) {
```
swankjesse: Move this check up to line 444?

swankjesse: (... then you already have `b` handy; you don't need to extract it from head again)

okio/src/main/java/okio/Buffer.java
```diff
@@ -775,6 +840,93 @@ public void clear() {
...
+      v = -v;
+      if (v < 0) { // Only true for Long.MIN_VALUE.
+        return writeUtf8("-9223372036854775808");
+      }
+      negative = true;
+    }
+
+    // Binary search for character width which favors matching lower numbers.
+    int width;
+    if (v < 100000000L) {
```
swankjesse: Tempting to see what this looks like as the world's most obnoxious ternary statement

swankjesse: (Perhaps with `_` chars afforded by Java 7, you could have something rather geometric?)

okio/src/main/java/okio/Buffer.java
```diff
@@ -775,6 +840,93 @@ public void clear() {
...
+    if (negative) {
+      data[--pos] = '-';
+    }
+
+    tail.limit += width;
+    size += width;
+    return this;
+  }
+
+  @Override public Buffer writeHexadecimalUnsignedLong(long v) {
```
swankjesse: Writing the leading `0` chars makes this lame for HTTP chunked encoding. See `HttpConnection.writeHex()`

okio/src/main/java/okio/RealBufferedSource.java
swankjesse: I think we need something like `RealBufferedSource.indexOfElement()`, but actually inverted. Maybe a private method, since otherwise the name will be really awkward?

okio/src/test/java/okio/BufferedSinkTest.java
```diff
@@ -211,4 +211,46 @@
...
+    assertLongDecimalString(Long.MIN_VALUE + 1);
+    assertLongDecimalString(Integer.MIN_VALUE - 1L);
+    assertLongDecimalString(Integer.MIN_VALUE);
+    assertLongDecimalString(-1);
+    assertLongDecimalString(0);
+    assertLongDecimalString(1);
+    assertLongDecimalString(Integer.MAX_VALUE);
+    assertLongDecimalString(Integer.MAX_VALUE + 1L);
+    assertLongDecimalString(Long.MAX_VALUE);
+    assertLongDecimalString(Long.MAX_VALUE - 1);
```
swankjesse: nit: swap this & previous line? Weird for these to be in _almost_ increasing order.

okio/src/test/java/okio/BufferedSinkTest.java
```diff
@@ -211,4 +211,46 @@
...
+    assertLongHexString(Long.MIN_VALUE + 1);
+    assertLongHexString(Integer.MIN_VALUE - 1L);
+    assertLongHexString(Integer.MIN_VALUE);
+    assertLongHexString(-1);
+    assertLongHexString(0);
+    assertLongHexString(1);
+    assertLongHexString(Integer.MAX_VALUE);
+    assertLongHexString(Integer.MAX_VALUE + 1L);
+    assertLongHexString(Long.MAX_VALUE);
+    assertLongHexString(Long.MAX_VALUE - 1);
```
swankjesse: same

okio/src/test/java/okio/BufferedSinkTest.java
```diff
@@ -177,4 +177,46 @@
...
+  }
+
+  private void assertLongHexString(long value) throws IOException {
+    sink.writeHexadecimalUnsignedLong(value).flush();
+    String expected = String.format("%16x", value).replace(' ', '0');
```
swankjesse: Glad you showed me your source code, 'cause it's wrong!
```
    System.out.println(String.format("%016x", 0xb0bL));
```

Merry Christmas, by the way!

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -468,4 +469,115 @@
...
+    }
+  }
+
+  @Test public void longHexStringTooShortThrows() throws IOException {
+    try {
+      data.writeUtf8(" ");
+      source.readHexadecimalUnsignedLong();
+      fail();
+    } catch (NumberFormatException e) {
+      assertEquals("Expected leading hex character but was 0x20", e.getMessage());
```
swankjesse: this error message is confusing, though I don't have a better suggestion

(complaining that 0x20 isn't hex)

swankjesse:
Really excited to see this land. Two big requests though:

Drop the leading zeros on hex literals. Or make 'em optional? For chunked encoding we don't want 'em, and as far as I know that's the most important use case.

Fix the `request(20)` call to work the same way that `readUtf8Line()` works. Somehow. That's going to be tricky to do nicely.

swankjesse:
(One actually plausible option is to just have have a custom loop in `RealBufferedSource` that knows about the ranges of hex & decimal values: `0-9, a-f, A-F`. That wouldn't be so bad.

okio/src/main/java/okio/Buffer.java
JakeWharton:
```java
    len = v < 100000000L
        ? v < 10000L
        ? v < 100L
        ? v < 10L ? 1 : 2
        : v < 1000L ? 3 : 4
        : v < 1000000L
        ? v < 100000L ? 5 : 6
        : v < 10000000L ? 7 : 8
        : v < 1000000000000L
        ? v < 10000000000L
        ? v < 1000000000L ? 9 : 10
        : v < 100000000000L ? 11 : 12
        : v < 1000000000000000L
        ? v < 10000000000000L ? 13
        : v < 100000000000000L ? 14 : 15
        : v < 100000000000000000L
        ? v < 10000000000000000L ? 16 : 17
        : v < 1000000000000000000L ? 18 : 19;
```

okio/src/main/java/okio/RealBufferedSource.java
JakeWharton: request != require, so it would take the 5 chars and then move on to parsing and succeed. I was worried about the other case where you have 5 but there's data on the wire that hasn't come across yet. In general, it's clear this needs to go byte-by-byte with request(1) looking for matching chars and then parse later, like you said.

JakeWharton:
PTAL

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,113 @@ public byte getByte(long pos) {
...
+          }
+          break outer;
+        }
+      }
+
+      if (pos == limit) {
+        head = segment.pop();
+        SegmentPool.INSTANCE.recycle(segment);
+      } else {
+        segment.pos = pos;
```
swankjesse: Do you need to make this call when you `break outer` ?

swankjesse: Actually you might need both this and the `pop` thing.

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,113 @@ public byte getByte(long pos) {
...
+      int pos = segment.pos;
+      int limit = segment.limit;
+
+      for (; pos < limit; pos++, seen++) {
+        byte b = data[pos];
+        if (b >= '0' && b <= '9') {
+          int digit = '0' - b;
+
+          // Detect when the digit would cause an overflow.
+          if (value < overflowZone || value == overflowZone && digit < overflowDigit) {
```
swankjesse: Does this handle asymmetry of the last digit? Long.MIN_VALUE has a different last digit than MAX_VALUE.

swankjesse: Ahhh, I see that below. Neat!

okio/src/main/java/okio/Buffer.java
```diff
@@ -407,6 +410,113 @@ public byte getByte(long pos) {
...
+        if (++seen > 16) {
+          Buffer buffer = new Buffer().writeHexadecimalUnsignedLong(value).writeByte(b);
+          throw new NumberFormatException("Number too large: " + buffer.readUtf8());
+        }
+
+        value <<= 4;
+        value |= digit;
+      }
+
+      if (pos == limit) {
```
swankjesse: Similarly?

okio/src/main/java/okio/Buffer.java
```diff
@@ -733,6 +785,93 @@ public void clear() {
...
+      return writeByte('0');
+    }
+
+    boolean negative = false;
+    if (v < 0) {
+      v = -v;
+      if (v < 0) { // Only true for Long.MIN_VALUE.
+        return writeUtf8("-9223372036854775808");
+      }
+      negative = true;
```
swankjesse: Yup! I like what you have.

okio/src/main/java/okio/Buffer.java
```diff
@@ -775,6 +885,97 @@ public void clear() {
...
+  }
+
+  @Override public Buffer writeHexadecimalUnsignedLong(long v) {
+    if (v == 0) {
+      // Both a shortcut and required since the following code can't handle zero.
+      return writeByte('0');
+    }
+
+    // Binary search for character width which favors matching lower numbers.
+    int width =  //
```
swankjesse: You want Long.highestOneBit or similar here. Computing digit count in hex doesn't require a binary search.

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -233,6 +233,52 @@ public RealBufferedSource(Source source) {
...
+      if (!request(pos + 1)) {
+        break; // No more data.
+      }
+      byte b = buffer.getByte(pos);
+      if ((b < '0' || b > '9') && (pos != 0 || b != '-')) {
+        break; // Non-digit, or non-leading negative sign.
+      }
+      pos += 1;
+      if (pos > 20) {
+        throw new NumberFormatException("Number too large: " + buffer.readUtf8());
```
swankjesse: Break and let the other method throw here?

swankjesse:
Thoughts on leading zeroes? I think JSON wants us to support `00000000000000000000001`.

okio/src/test/java/okio/BufferedSinkTest.java
```diff
@@ -211,4 +212,41 @@
...
+    assertLongDecimalString(Long.MAX_VALUE);
+
+    for (int i = 1; i < 20; i++) {
+      long value = BigInteger.valueOf(10L).pow(i).longValue();
+      assertLongDecimalString(value - 1);
+      assertLongDecimalString(value);
+    }
+  }
+
+  private void assertLongDecimalString(long value) throws IOException {
```
swankjesse: Confirm the buffer is left in the expected state? Maybe write a trailing `abc` and read it back?

JakeWharton: Done

JakeWharton:
RE: leading zero, already supported

okio/src/main/java/okio/Buffer.java
JakeWharton: Ah, yeah. That's why I had used the `done` boolean flag before.

okio/src/main/java/okio/Buffer.java
swankjesse: Does this break too many leading 0s? (Not likely to be a problem in practice)

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -233,6 +233,51 @@ public RealBufferedSource(Source source) {
...
+    while (true) {
+      if (!request(pos + 1)) {
+        break; // No more data.
+      }
+      byte b = buffer.getByte(pos);
+      if ((b < '0' || b > '9') && (pos != 0 || b != '-')) {
+        break; // Non-digit, or non-leading negative sign.
+      }
+      if (++pos > 20) {
+        break; // Exceeded valid digit count.
```
swankjesse: I think this might be incorrect for input like `0000000000000000000000000000001` because we might not block enough

swankjesse: (returning `0` instead of `1` as we should)

swankjesse:
Love this.

swankjesse:
https://github.com/square/okio/pull/118/files

Issue #94, opened 2014-12-21
Different buffer pool implementations suitable for server use?


tinkerware:
Hi,

Would you be willing to accept changes for a pluggable segment pool to make Okio more suitable for use in server environments with lots of concurrent buffer use?

swankjesse:
Though I haven't done much on this lately, I think we're going to need it. Here's more benchmarks, this time from Guava. https://github.com/google/guava/pull/2112

swankjesse:
Fixed recently.
https://github.com/square/okio/pull/717

PR #106, opened 2015-01-06
Add JMH benchmarks for measuring buffer performance.

tinkerware:
As discussed in #94 here are some benchmarks so that we can discuss segment pool alternatives in server environments.

These microbenchmarks try to measure performance of common buffer operations in a request/response handling scenario and a multi-threaded server environment. There are two benchmarks:

- Hot: This is a "pedal-to-the-metal" test. It is not a realistic use of the buffers but it may be useful in finding out performance regressions that are otherwise too subtle to show up any other way.

- Cold: This test simulates request handling with a configurable think time between responses. It assumes request data is saved and queued as they arrive before they are rate-limited/expired and that response workers outnumber request handlers. I think these are reasonable assumptions based on existing high-performance servers (for example, Cassandra); we give preference to recently arrived requests instead of refusing service and assume that timed-out requests are expired from their queue without ever seeing a worker.

Maximum number of bytes to read from and write to a buffer at one time, the request size and how large responses are compared to requests (`responseFactor`) are configurable parameters.

### DIFF
--- a/benchmarks/src/main/java/com/squareup/okio/benchmarks/BufferPerformanceBench.java
+++ b/benchmarks/src/main/java/com/squareup/okio/benchmarks/BufferPerformanceBench.java
@@ -1,0 +1,330 @@
+/*
+ * Copyright (C) 2014 Square, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.squareup.okio.benchmarks;
+
+import java.io.EOFException;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.TimeUnit;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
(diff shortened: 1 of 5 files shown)


benchmarks/src/main/java/com/squareup/okio/benchmarks/BufferPerformanceBench.java
```diff
@@ -0,0 +1,330 @@
...
+
+@Fork(1)
+@Warmup(iterations = 10, time = 10)
+@Measurement(iterations = 10, time = 10)
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class BufferPerformanceBench {
+
+  public static final File OriginPath =
```
JakeWharton: nit: should be named `originPath` or `ORIGIN_PATH`. Title case isn't normal convention.

JakeWharton:
This looks cool! Don't have time to review or test it right now.

tinkerware:
I've run these benchmarks on an EC2 c3.8xlarge instance over the weekend:

https://plot.ly/~tinkerware/97 and https://plot.ly/~tinkerware/77 have a few charts showing that the segment pool very quickly becomes a bottleneck. The JMH logs of the cold benchmark run are at https://gist.github.com/tinkerware/f410f139b775f83a6039.

tinkerware:
So, looking at the results for the "cold" benchmark, we can utilize at most two cores according to Linux `perf`, and the stack profiler shows that most of that is spent recycling segments and contending. After falling though a cliff at 8 cores, throughput levels off and slowly declines once we are past 12 cores (3 request reader and 9 response writer threads). Above that, context switches level off around 25K/sec and threads keep spending more and more time being blocked and migrated between cores.

A segment pool using `j.u.c.ConcurrentLinkedQueue` would deal with multi-threaded use better. I'm planning to experiment with a pluggable segment pool mechanism soon; I'll try that out and see how much improvement we can get.

swankjesse:
@tinkerware awesome.

swankjesse:
I'm merging right now 'cause I want to start playing with this. I'm also eager to see what happens with better segment pooling.

PR #108, opened 2015-01-12
Make a Sink also a Flushable.

JakeWharton:


swankjesse:
Will this cause problems on AOSP?

swankjesse:
Short answer: no.

Issue #112, opened 2015-02-02
Slice methods on ByteString


swankjesse:
To extract a prefix, substring or suffix.

f2prateek:
+1, we have wrappers around it currently, but would be nice to have it natively. I'll send a PR shortly.

JakeWharton:
I think this can be closed. @swankjesse are you satisfied with the two substring overloads?

PR #113, opened 2015-02-02
Add ByteString#substring methods

f2prateek:
Ref #112

### DIFF
--- a/okio/src/main/java/okio/ByteString.java
+++ b/okio/src/main/java/okio/ByteString.java
@@ -211,6 +211,37 @@ public final class ByteString implements Serializable {
     return this;
   }
 
+  /**
+   * Returns a byte string that is a substring of this byte string, beginning at the specified 
+   * index until the end of this string. Returns this byte string if {@code beginIndex} is 0.
+   */
+  public ByteString substring(int beginIndex) {
+    return substring(beginIndex, data.length);
+  }
+
+  /**
+   * Returns a byte string that is a substring of this byte string, beginning at the specified 
+   * {@code beginIndex} and ends at the specified {@code endIndex}. Returns this byte string if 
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/ByteString.java
```diff
@@ -211,6 +211,37 @@ public ByteString toAsciiUppercase() {
...
+    
+    int subLen = endIndex - beginIndex;
+    if (subLen < 0) throw new IllegalArgumentException("endIndex < beginIndex");
+    
+    if ((beginIndex == 0) && (endIndex == data.length)) {
+      return this;
+    }
+
+    byte[] copy = new byte[subLen];
+    System.arraycopy(data, beginIndex, copy, 0, subLen);
```
JakeWharton: Since these are immutable we should just return a view of the original data rather than require a copy.

f2prateek: That could cause a memory leak by keeping the potentially huge and unwanted original array around.

JakeWharton: The inverse is also true: this can cause massive amounts of useless objects by copying data that we know will never change.

f2prateek: True, this will be a much bigger change in that case, since we'll have to keep track of the length and offset on the underlying array manually.

JakeWharton: I think there's advantages in re-using the underlying storage. For example, we read headers using a `ByteString` and it would be advantageous to do a zero-copy "split" it into name and value. Perhaps we document the fact that it's a view and require you to call `.clone()` if you want a defensive copy?

swankjesse: For what it's worth, I think the copy is less likely to surprise.

swankjesse:
Thanks!

PR #114, opened 2015-02-06
This PR fixes #107

### LINKED ISSUE #107: Watchdog clean shut down
Okio version 1.2.0.

Currently there does not seem to be a way to cleanly shut down the watchdog thread "Okio Watchdog" created by Okio AsyncTimeout. This results in the following warning when shutting down an application.
```
org.apache.catalina.loader.WebappClassLoader clearReferencesThreads
WARNING: The web application [/] appears to have started a thread named [Okio Watchdog] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread:
 java.lang.Object.wait(Native Method)
 java.lang.Object.wait(Object.java:460)
 okio.AsyncTimeout.awaitTimeout(AsyncTimeout.java:309)
 okio.AsyncTimeout.access$000(AsyncTimeout.java:40)
 okio.AsyncTimeout$Watchdog.run(AsyncTimeout.java:272)
```

lukiano:
There is a flag "running" to break the loop, but it will consume the queue first.

### DIFF
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -284,7 +284,6 @@ public class AsyncTimeout extends Timeout {
           // Close the timed out node.
           timedOut.timedOut();
         } catch (InterruptedException ignored) {
-            //Thread.interrupted();
         }
       }
       releaseHead();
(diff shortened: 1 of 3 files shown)


swankjesse:
https://github.com/square/okio/issues/107

swankjesse:
I'm uncomfortable with this.

I don't want callers to need to know about AsyncTimeout, nor for them to manually need to call `close()`. I also dislike that this new method is `static synchronized`. Whose responsibility is it to call `close()`? This isn't composable.

Do the app-reloading containers offer hooks to detect unloading? Ideally we can register with some central authority that we're creating a thread, and that authority will notify us when we're being unloaded.

dyshen:
I would recommend to automatically shutdown the thread in the `close()` method of `Source`.

swankjesse:
The fix I’d like to implement is to have the watchdog exit after say, 60 seconds of inactivity. This change is awkward because the application developer has to know it exists.

PR #116, opened 2015-02-18
Segment sharing.

swankjesse:
We now have three different types of segments:
 * Private. This was the only type of segment previously offered.
   This is shared=false, owner=true.
 * Appendable. Segments of this type can be written, but only
   beyond segment.limit. Bytes preceding segment.limit are shared
   and must not be written. This is shared=true, owner=true.
 * Unwritable. Segments of this type cannot be written at all.
   They are owned by another buffer. This is shared=true, owner=false.

Note that there are no segments of type 'shared=false, owner=false',
since the only way a buffer doesn't own its segment is if that segment
is shared.

This introduces a new subclass of ByteString, SegmentedByteString.
These byte strings share backing arrays with buffers. SegmentedByteString
instances are immutable and support fast random access. Reading a random
byte from a SegmentedByteString is O(log N); reading a random byte from a
Buffer is O(N).

Segment sharing isn't necessarily more efficient in all cases. For example,
shared segments are not pooled; this harms efficiency of segment allocation.
Shared segments may also be wasteful of memory: previously segments were
consistently > 50% full. With this change, we have no such minimum.

This change should have significant benefits for Buffer.copyTo(), which
is an important operation of OkHttp. Right now whenever a single buffer
needs to be written to multiple destinations, we manually copy the buffer
so that each destination gets its own private copy of the buffer's segments.
By sharing segments, the work required by copyTo() is cut down to the
segment bookkeeping. No byte arrays are copied.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -86,7 +86,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return this; // Nowhere to emit to!
   }
 
-  @Override public BufferedSink emit() throws IOException {
+  @Override public BufferedSink emit() {
     return this; // Nowhere to emit to!
   }
 
@@ -98,7 +98,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     if (size < byteCount) throw new EOFException();
   }
 
-  @Override public boolean request(long byteCount) throws IOException {
+  @Override public boolean request(long byteCount) {
     return size >= byteCount;
   }
 
@@ -164,30 +164,26 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     checkOffsetAndCount(size, offset, byteCount);
     if (byteCount == 0) return this;
 
-    Segment source = head;
-    Segment target = out.writableSegment(1);
     out.size += byteCount;
 
-    while (byteCount > 0) {
-      // If necessary, advance to a readable source segment. This won't repeat after the first copy.
-      while (offset >= source.limit - source.pos) {
-        offset -= (source.limit - source.pos);
-        source = source.next;
-      }
+    // Skip segments that we aren't copying from.
+    Segment s = head;
+    for (; offset >= (s.limit - s.pos); s = s.next) {
+      offset -= (s.limit - s.pos);
+    }
 
-      // If necessary, append another target segment.
-      if (target.limit == Segment.SIZE) {
-        target = target.push(SegmentPool.INSTANCE.take());
+    // Copy one segment at a time.
+    for (; byteCount > 0; s = s.next) {
+      Segment copy = new Segment(s);
+      copy.pos += offset;
+      copy.limit = Math.min(copy.pos + (int) byteCount, copy.limit);
+      if (out.head == null) {
+        out.head = copy.next = copy.prev = copy;
+      } else {
+        out.head.prev.push(copy);
       }
-
-      // Copy bytes from the source segment to the target segment.
-      long sourceReadable = Math.min(source.limit - (source.pos + offset), byteCount);
-      long targetWritable = Segment.SIZE - target.limit;
-      int toCopy = (int) Math.min(sourceReadable, targetWritable);
-      System.arraycopy(source.data, source.pos + (int) offset, target.data, target.limit, toCopy);
-      offset += toCopy;
-      target.limit += toCopy;
-      byteCount -= toCopy;
+      byteCount -= copy.limit - copy.pos;
+      offset = 0;
     }
 
     return this;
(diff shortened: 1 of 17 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -164,30 +164,26 @@ public Buffer copyTo(Buffer out, long offset, long byteCount) {
...
+        out.head = copy.next = copy.prev = copy;
+      } else {
+        out.head.prev.push(copy);
       }
-
-      // Copy bytes from the source segment to the target segment.
-      long sourceReadable = Math.min(source.limit - (source.pos + offset), byteCount);
-      long targetWritable = Segment.SIZE - target.limit;
-      int toCopy = (int) Math.min(sourceReadable, targetWritable);
-      System.arraycopy(source.data, source.pos + (int) offset, target.data, target.limit, toCopy);
```
adriancole: I think you forgot to do some array copies? Jk nice work!

okio/src/main/java/okio/Buffer.java
```diff
@@ -164,30 +164,26 @@ public Buffer copyTo(Buffer out, long offset, long byteCount) {
...
+    for (; offset >= (s.limit - s.pos); s = s.next) {
+      offset -= (s.limit - s.pos);
+    }
 
-      // If necessary, append another target segment.
-      if (target.limit == Segment.SIZE) {
-        target = target.push(SegmentPool.INSTANCE.take());
+    // Copy one segment at a time.
+    for (; byteCount > 0; s = s.next) {
+      Segment copy = new Segment(s);
```
adriancole: I think this is the most subtle part of the code. It isn't really the concern of this method whether it is a copy of the input data, or a view of it. I think those new to okio post-merge might not assume copy meant copy of data. For these reasons, and the fact that I can't think of a better word than copy, LG

okio/src/main/java/okio/Buffer.java
```diff
@@ -589,14 +585,15 @@ public void clear() {
...
-        SegmentPool.INSTANCE.recycle(toRecycle);
+        SegmentPool.recycle(toRecycle);
       }
     }
   }
 
   @Override public Buffer write(ByteString byteString) {
     if (byteString == null) throw new IllegalArgumentException("byteString == null");
-    return write(byteString.data, 0, byteString.data.length);
+    byteString.write(this);
```
adriancole: the evolution of buffer and bytestring together are starting to make this method look deprecate worthy.. except for its chaining virtue.

okio/src/main/java/okio/DeflaterSink.java
```diff
@@ -99,7 +99,7 @@ private void deflate(boolean syncFlush) throws IOException {
         buffer.size += deflated;
         sink.emitCompleteSegments();
       } else if (deflater.needsInput()) {
-        return;
+        return; // TODO(jwilson): do we have a dangling empty tail segment here?
```
adriancole: add an assert near the TODO?

okio/src/main/java/okio/SegmentedByteString.java
```diff
@@ -0,0 +1,238 @@
...
+
+import static okio.Util.arrayRangeEquals;
+import static okio.Util.checkOffsetAndCount;
+
+/**
+ * An immutable byte string composed of segments of byte arrays. This class exists to implement
+ * efficient snapshots of buffers. It is implemented as an array of segments, plus a directory in
+ * two halves that describes how the segments compose this byte string.
+ *
+ * <p>The first half of the directory is the cumulative byte count covered by each segment. The
```
adriancole: very clever!

adriancole:
LGTM Agreed that this should help in okhttp where the same payload is written to multiple times, for example, lowering the impact of a retry.

okio/src/main/java/okio/Segment.java
```diff
@@ -79,28 +107,12 @@ public Segment push(Segment segment) {
...
-    } else {
-      // Create a new segment of size 'bSize' after this segment.
-      Segment after = SegmentPool.INSTANCE.take();
-      System.arraycopy(data, pos + aSize, after.data, after.pos, bSize);
-      limit -= bSize;
-      after.limit += bSize;
-      push(after);
-      return this;
-    }
+    if (byteCount <= 0 || byteCount > limit - pos) throw new IllegalArgumentException();
```
JakeWharton: This looks like `Util.checkOffsetAndCount`.

swankjesse:
The problem with pooling segment instances is that I'd need to reference count them. And that's easy, except that in a SegmentedByteString, they'd never be released.

JakeWharton:
@swankjesse what's the what here? I don't want all this work to become stale.

swankjesse:
Merged! I suspect there will be other optimizations to follow, mostly around putting a ceiling on the amount of memory used. (Policy will be to only share when there's a non-negligible benefit to doing so; say at least 256 bytes.)

PR #121, opened 2015-03-07
Add convenience methods for MD5 and SHA-256 hashes.

### LINKED ISSUE #110: Make computing hashes easier
Maybe `ByteString.md5()` and `ByteString.sha256()` would be sufficient.

JakeWharton:
Closes #110.

### DIFF
--- a/okio/src/main/java/okio/ByteString.java
+++ b/okio/src/main/java/okio/ByteString.java
@@ -101,6 +101,24 @@ public final class ByteString implements Serializable {
     return Base64.encode(data);
   }
 
+  /** Returns the MD5 hash of this byte string. */
+  public String md5() {
+    return digest("MD5");
+  }
+
+  /** Returns the SHA-256 hash of this byte string. */
+  public String sha256() {
+    return digest("SHA-256");
+  }
+
+  private String digest(String digest) {
+    try {
+      return ByteString.of(MessageDigest.getInstance(digest).digest(data)).hex();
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/ByteString.java
```diff
@@ -101,6 +101,24 @@ public String base64() {
     return Base64.encode(data);
   }
 
+  /** Returns the MD5 hash of this byte string. */
+  public String md5() {
```
swankjesse: Hmmmm . . . I was thinking this should return a ByteString, and make it the caller's responsibility to call `hex()`.

(On the Cash server, we store binary hashes directly in our database, which reduces the storage requirement by 50%).

okio/src/main/java/okio/ByteString.java
```diff
@@ -285,12 +303,7 @@ public void write(OutputStream out) throws IOException {
       return String.format("ByteString[size=%s data=%s]", data.length, hex());
     }
 
-    try {
-      return String.format("ByteString[size=%s md5=%s]", data.length,
-          ByteString.of(MessageDigest.getInstance("MD5").digest(data)).hex());
-    } catch (NoSuchAlgorithmException e) {
-      throw new AssertionError();
-    }
+    return String.format("ByteString[size=%s md5=%s]", data.length, md5());
```
swankjesse: Nice.

PR #126, opened 2015-03-11
allow base64-decoding byte arrays, not just strings

arnt:
This is just a convenience function; it helps to avoid having to convert data from the network from bytes to String before decodeBase64() converts it back and decodes it. If you accept this sort of overload I'll provide some more, e.g. decodeBase64(final ByteString in) and an RFC-compliant encoding variant.

I see the diff removes some spurious whitespace at EOL. Oh well.

### DIFF
--- a/okio/src/main/java/okio/Base64.java
+++ b/okio/src/main/java/okio/Base64.java
@@ -22,15 +22,24 @@ package okio;
 
 import java.io.UnsupportedEncodingException;
 
+import java.nio.charset.Charset;
+
+
 final class Base64 {
+  static private Charset utf8 = Charset.forName("utf8");
+
   private Base64() {
   }
 
-  public static byte[] decode(String in) {
+  public static byte[] decode(final String in) {
+      return decode(in.getBytes(utf8));
+  }
+
+  public static byte[] decode(final byte[] in) {
     // Ignore trailing '=' padding and whitespace from the input.
-    int limit = in.length();
+    int limit = in.length;
     for (; limit > 0; limit--) {
-      char c = in.charAt(limit - 1);
+      byte c = in[limit - 1];
       if (c != '=' && c != '\n' && c != '\r' && c != ' ' && c != '\t') {
         break;
       }
@@ -43,7 +52,7 @@ final class Base64 {
 
     int word = 0;
     for (int pos = 0; pos < limit; pos++) {
-      char c = in.charAt(pos);
+      byte c = in[pos];
 
       int bits;
       if (c >= 'A' && c <= 'Z') {
--- a/okio/src/main/java/okio/ByteString.java
+++ b/okio/src/main/java/okio/ByteString.java
@@ -105,7 +105,17 @@ public final class ByteString implements Serializable {
    * Decodes the Base64-encoded bytes and returns their value as a byte string.
    * Returns null if {@code base64} is not a Base64-encoded sequence of bytes.
    */
-  public static ByteString decodeBase64(String base64) {
+  public static ByteString decodeBase64(final String base64) {
+    if (base64 == null) throw new IllegalArgumentException("base64 == null");
+    byte[] decoded = Base64.decode(base64);
+    return decoded != null ? new ByteString(decoded) : null;
+  }
+
+  /**
+   * Decodes the Base64-encoded bytes and returns their value as a byte string.
+   * Returns null if {@code base64} is not a Base64-encoded sequence of bytes.
+   */
+  public static ByteString decodeBase64(final byte[] base64) {
     if (base64 == null) throw new IllegalArgumentException("base64 == null");
     byte[] decoded = Base64.decode(base64);
     return decoded != null ? new ByteString(decoded) : null;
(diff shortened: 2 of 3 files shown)


arnt:
Oh, and the reason why Travis complains is that HEAD won't compile, and this diff doesn't solve HEAD's problems. The problem seems to be an earlier mismerge, which you might want to attend to before looking at this.

swankjesse:
I'm really reluctant to accept this. Why would a reasonable application have base64 data as a `byte[]`?

arnt:
Because the origin (a TCP connection and an ABNF parser) delivers byte strings and the code has not yet been converted to use ByteString.

For example, SMTP and SASL are defined in terms of byte strings. An SMTP server code reads bytes from a TCP connection until it sees CR LF, then (depending on mode) it either interprets the bytes it has as ASCII, as a base64-encoded SASL blob, as UTF8, or as opaque data.

Most Java programs hold that byte string in a String object. so you'll occasionally see String objects that actually contain ISO-8859-6 (Arabic). ByteString is a much better choice now that it exists. The code I am working on generally uses byte[] for raw network data, and is slowly being migrated to ByteString.

FWIW, fitting binary data into printable ASCII is the very purpose of base64, so I think being able to decode from byte[] and from ByteString is natural, and decodeBase64(final String in) is the unnatural variant.

swankjesse:
Alright, I'm convinced!

swankjesse:
More questions:
1. Should this accept a `byte[]` or a `ByteString`?
2. Should we do likewise for encoding to base64?
3. Should we do likewise for hex?

(Not necessarily to be addressed in this pull request, but I want to make sure the API is consistent.)

arnt:
1/3. You'll get a ByteString variant too, I just wanted to send something small before I type my fingers to stubs. Encoding to doesn't seem to need these variants, but of(final byte[]) is handy. For example I have code that should rightly look like this: MessageDigest foo = ... ; ByteString.of(d.digest()).base64().toByteArray().

swankjesse:
I'm thinking about all of the possible APIs we could write (but shouldn't):
 * Input data type can be byte[], ByteString, or String.
 * Input encoding can be raw, base64-encoded, base64url encoded, or hex-encoded.
 * Output data type can be byte[], ByteString, or String.
 * Output encoding can be raw, base64-encoded, base64url encoded or hex-encoded.

Right now we offer the following 5 APIs:
 * decodeHex() input is a hex String, output is a raw ByteString
 * hex() input is a raw ByteString, output is a hex String
 * decodeBase64() input is base64 or base64url String, output is a raw ByteString
 * base64() input is a raw ByteString, output is a base64 String
 * base64url() input is a raw ByteString, output is a base64url String

We definitely don't want to offer everything:
 * We can eliminate anything that does an identity transformation.
 * We can eliminate anything that is doesn't have raw as input or output type.
 * I don't want to encourage anyone to use a ByteString whose content is not raw. You shouldn't ever have a ByteString with hex or base64-encoded bytes in it; that's weird. (Just as Java Strings should never have arbitrary binary data in them). So let's avoid those.

After scratching that stuff, what's still interesting, and still missing is hex-as-byte[], base64-as-byte[], base64url-as-byte[] for encoding and decoding.

swankjesse:
I'm managing to get myself unconvinced that we want these new APIs. I like making normal every day things easy. But I wonder whether this is really what we want . . .

If this is an optimization, it's an imbalanced one. Most of the time when we're dealing with hex and base64, we're dealing with it as a string because infrastructure like JSON and XML and HTTP make strings easy (and binary hard). This optimization changes the performance characteristic for strings so that they now do an allocation (`toByteArray()`) where they didn't previously. So we make that common case slower in order to make the `byte[]` case fast.

Perhaps we really want something more aggressively optimized? Like a Base64EncodingSink and a Base64DecodingSource? Those things could decode values too-large for memory!

arnt:
I don't know what you want. What I want is to read ByteString from TCP and 64-decode it without having to detour via String, do a few similar chores, and to help with transition from old code to ByteString-using code. And I want all that with pretty code.

Perhaps decoding from String is common, but I think it's just *wrong* if ByteString.decodeBase64() is optimized for working on String; its prime goal should be to make code that uses ByteString consistently readable and fast. Supporting decodeBase64(final String in) and decodeBase64(final byte[] in) is desirable because there's so much old code that cannot be changed (at least not at once), but the bestest version should be decodeBase64(final ByteString in).

swankjesse:
Yeah, I don't know what I want either!

Better question: what are you using to read from TCP? Are you using an Okio source? That's why I suggested that a Base64-encoding Source could provide the best option: it could decode base64 without any unnecessary copying.

arnt:
I expect my case is fairly typical for network protocol work... I read byte-oriented protocols that contain base64 fields now and then.  The example on http://tools.ietf.org/html/rfc2554#page-4 looks like much of what I do, although it comes in different guises. The copying isn't a major problem. Undesirable of course, but phones are neat that way: If the app crashes because it would need to copy 1TB, that's better than sending 1TB and draining the battery ;)

I'll just update the pull request with all five functions I think should be there. Should be all of ten lines of code plus tests and javadoc. Yes it will hurt people who work with base64-encoded blobs XML, but a little bird tells me that they won't even notice ;)

swankjesse:
I'd prefer to not add another allocation to the string case. Some of this code gets executed in tight loops, such as in databinding.

swankjesse:
(and the nature of Android is that allocations cause jank)

swankjesse:
I think my options are all bad:

1. Merge this and degrade performance for strings.
2. Duplicate the base64 implementation for byte[] and increase our library size.
3. Decline this, and punish you unfairly!

I'm going with option #3, which is bad and I feel bad. But this isn't a base64 library, and I don't think I'd like the code more with options 1 and 2. Sorry about this! If it's a performance concern in your code, please copy the Base64 class. If it's an API concern, adding a helper method to convert shouldn't be too bad either.

PR #135, opened 2015-03-30
Add behavior decorators for emulating speeds and failures.

JakeWharton:


JakeWharton:
Giant (nearly) unsolicited code drop commence!

I chose to emulate `Timeout` with `Behavior` and `ByteRate` in having them be mutable objects owned by the `BehaviorSink` and `BehaviorSource`. Originally these were immutable objects, and I'd be happy to switch back to it. The mutable ones are nice because it allows you to easily alter the behavior over time. And most essentially, it does not require re-wrapping the `Sink` or `Source`.

okio-behavior/src/main/java/okio/test/Behavior.java
```diff
@@ -0,0 +1,84 @@
+package okio.test;
```
swankjesse: copyright

JakeWharton: just added them all!

okio-behavior/src/main/java/okio/test/Behavior.java
```diff
@@ -0,0 +1,84 @@
...
+
+import java.io.IOException;
+import okio.Buffer;
+import okio.Sink;
+import okio.Source;
+
+public final class Behavior {
```
swankjesse: very interesting name. I don't hate it!

okio-behavior/src/main/java/okio/test/Behavior.java
```diff
@@ -0,0 +1,84 @@
...
+  public interface Callback {
+    void apply(Behavior behavior) throws IOException;
```
swankjesse: apply?

okio-behavior/src/main/java/okio/test/Behavior.java
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  private volatile Callback beforeFlushCallback;
+  private volatile Callback afterFlushCallback;
+  private volatile Callback beforeCloseCallback;
+  private volatile Callback afterCloseCallback;
+  private volatile ByteRate byteRate = new ByteRate();
```
JakeWharton: Should be `final` now that this is mutable rather than immutable.

okio-behavior/src/main/java/okio/test/Behavior.java
```diff
@@ -0,0 +1,84 @@
...
+  void beforeFlush() throws IOException {
+    Callback beforeFlushCallback = this.beforeFlushCallback;
+    if (beforeFlushCallback != null) {
+      beforeFlushCallback.apply(this);
+    }
+  }
+
+  void afterFlush() throws IOException {
+    Callback afterFlushCallback = this.afterFlushCallback;
+    if (afterFlushCallback != null) {
```
swankjesse: nice null checks!

okio-behavior/src/main/java/okio/test/BehaviorSink.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public BehaviorSink(Sink delegate) {
+    super(delegate);
+  }
+
+  public Behavior behavior() {
+    return behavior;
+  }
+
+  @Override public void write(Buffer source, long byteCount) throws IOException {
```
swankjesse: This isn't right. You can defeat the rate by manipulating the frequency of calls to `write()`. You need to keep a `long written` field member that starts out.

okio-behavior/src/main/java/okio/test/BehaviorSink.java
```diff
@@ -0,0 +1,62 @@
...
+    ByteRate rate = behavior.byteRate();
+
+    long written = 0L;
+    while (written < byteCount) {
```
swankjesse: I think this is a for loop

okio-behavior/src/main/java/okio/test/BehaviorSink.java
```diff
@@ -0,0 +1,62 @@
...
+          return;
+        }
+      }
+    }
+  }
+
+  @Override public void flush() throws IOException {
+    behavior.beforeFlush();
+    super.flush();
+    behavior.afterFlush();
```
swankjesse: try/finally ?

okio-behavior/src/main/java/okio/test/ByteRate.java
```diff
@@ -0,0 +1,59 @@
+package okio.test;
+
+import java.util.concurrent.TimeUnit;
+
+/** A byte count expressed over a period of time for determining transfer rates. */
+public final class ByteRate {
```
swankjesse: should this have helper functions ?

```
ByteRate byteRate = new ByteRate();
byteRate.kBps(28);
byteRate.mBps(28);
```

The bytes/bits nonsense burns us a bit here here.

JakeWharton: We could add a dependency to [ByteUnits][1] :grinning: 
```java
rate.count(28, KIBIBYTES).per(5, SECONDS);
rate.count(28, MEGABYTES).per(5, SECONDS);
```
[1]: https://github.com/JakeWharton/byteunits

okio-behavior/src/main/java/okio/test/BehaviorSink.java
JakeWharton: How do you mean? No matter how many of how few calls you make we chunk them and sleep accordingly. And if you make many calls we adjust the sleep based on the partial amount.

okio-behavior/src/main/java/okio/test/BehaviorSink.java
JakeWharton: I added documentation to the fact that exceptions will result in later methods not being called. The main point of them is to throw exceptions, and I don't want to deal with the exception in `finally` case unless you feel very strongly.

swankjesse: Works for me.

JakeWharton:
Working on a v2 of this. Closing for now.

PR #138, opened 2015-04-12
Add native substring writing to BufferedSink.

JakeWharton:


JakeWharton:
There are instances in Moshi where we write substrings for replacing characters. This will avoid a `String` allocation, a `char[]` allocation for the substring, and the data copy from the original underlying `char[]` for that operation and move it to happen natively.

okio/src/main/java/okio/Buffer.java
```diff
@@ -782,10 +794,23 @@ public void clear() {
...
+      throw new IllegalArgumentException("endIndex < beginIndex: " + endIndex + " < " + beginIndex);
+    }
+    if (endIndex > string.length()) {
+      throw new IllegalArgumentException(
+          "endIndex > string.length: " + endIndex + " > " + string.length());
+    }
     if (charset == null) throw new IllegalArgumentException("charset == null");
     if (charset.equals(Util.UTF_8)) return writeUtf8(string);
-    byte[] data = string.getBytes(charset);
+    byte[] data = string.substring(beginIndex, endIndex).getBytes(charset);
```
swankjesse: This doesn't allocate, right?

swankjesse: (... on the case where we're writing the full string.)

JakeWharton: It does on JDK 8+. Not sure about Android. I originally had a "TODO see if we can get a subset of bytes rather than bytes of substring" but deleted it before making the PR.

swankjesse:
Good find!

okio/src/main/java/okio/Buffer.java
JakeWharton: Also, it does not allocate when `beginIndex == 0` and `endIndex == string.length()`.

PR #140, opened 2015-04-16
Add isEmpty() convenience method.

JakeWharton:
`size() == 0` is so 2014

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -59,6 +59,11 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return size;
   }
 
+  /** True if there are no bytes currently in this buffer. */
+  public boolean isEmpty() {
+    return size == 0;
+  }
+
   @Override public Buffer buffer() {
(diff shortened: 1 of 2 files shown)


swankjesse:
I know I suggested this. Do we want both this and `exhausted()` ?

swankjesse:
LGTM

JakeWharton:
No. I think we just keep forgetting to use it!

JakeWharton:
Although perhaps the semantics of `exhausted()` when dealing with a buffer justify this method because `exhausted()` sounds like something you check while reading whereas `isEmpty()` is a check that can be made after conditionally writing.

JakeWharton:
Examples where `isEmpty` would probably read better:
```java
if (content.exhausted()) {
  throw new IllegalStateException("Form encoded body must have at least one part.");
}
```
```java
while (readBuffer.exhausted() && !finished && !closed && errorCode == null) {
  waitForIo(); // <-- maybe writes into readBuffer
}
```
This one isn't so bad, but would read even better:
```java
boolean wasEmpty = readBuffer.exhausted();
```

PR #142, opened 2015-04-19
Throw EOFException reading digits from an empty source.

swankjesse:
Currently we crash while composing the NumberFormatException on the call
to get(0).

Closes: https://github.com/square/okhttp/issues/1540

### DIFF
--- a/okio/src/main/java/okio/RealBufferedSource.java
+++ b/okio/src/main/java/okio/RealBufferedSource.java
@@ -234,40 +234,36 @@ final class RealBufferedSource implements BufferedSource {
   }
 
   @Override public long readDecimalLong() throws IOException {
-    int pos = 0;
-    while (true) {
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -234,40 +234,36 @@ public RealBufferedSource(Source source) {
...
+          throw new NumberFormatException(String.format(
+              "Expected leading [0-9] or '-' character but was %#x", buffer.getByte(0)));
+        }
+        break;
       }
-      pos++;
-    }
-    if (pos == 0) {
-      throw new NumberFormatException("Expected leading [0-9] or '-' character but was 0x"
-          + Integer.toHexString(buffer.getByte(0)));
```
swankjesse: We were crashing here 'cause the buffer was empty.

PR #143, opened 2015-04-20
Ensure empty sources work with gzip and inflater wrappers.

JakeWharton:
Closes https://github.com/square/retrofit/issues/804.

### DIFF
--- a/okio/src/main/java/okio/GzipSource.java
+++ b/okio/src/main/java/okio/GzipSource.java
@@ -68,6 +68,10 @@ public final class GzipSource implements Source {
     if (byteCount < 0) throw new IllegalArgumentException("byteCount < 0: " + byteCount);
     if (byteCount == 0) return 0;
 
+    if (source.exhausted()) {
+      return -1;
+    }
+
     // If we haven't consumed the header, we must consume it before anything else.
     if (section == SECTION_HEADER) {
       consumeHeader();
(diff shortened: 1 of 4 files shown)


okio/src/main/java/okio/GzipSource.java
```diff
@@ -68,6 +68,10 @@ public GzipSource(Source source) {
     if (byteCount < 0) throw new IllegalArgumentException("byteCount < 0: " + byteCount);
     if (byteCount == 0) return 0;
 
+    if (source.exhausted()) {
```
swankjesse: Yuck. Do we really need to support malformed gzip responses?

swankjesse:
I'm pushing back on the retrofit bug...

okio/src/main/java/okio/GzipSource.java
JakeWharton: While it may be malformed in the context of HTTP, I'm not sure passing something empty through an `GzipSource` that happens to be empty is malformed (e.g., a `File`) and we should rather adhere to the `Source` spec.

JakeWharton: Full disclosure: I agree with both sides of the argument here. If it were in OkHttp I'd fall on the not support side but only in a pure Okio sense do I think special casing empty might be passable.

I should find out whether you can actually have an empty gzip payload (with header + footer but no data) because if you can then that would probably sway me against this change.

swankjesse: You can have a gzip payload with header + footer but no data.

JakeWharton: Works for me.

PR #145, opened 2015-04-25
APIs to read and write code points.

swankjesse:
The String APIs transcode UTF-16 to UTF-8 and back.

These APIs avoid the UTF-16 intermediate form altogether, and
go right from UTF-8 to a codepoint and back.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -47,6 +47,7 @@ import static okio.Util.reverseBytesLong;
 public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
   private static final byte[] DIGITS =
       { '0', '1', '2', '3', '4', '5', '6', '7', '8', '9', 'a', 'b', 'c', 'd', 'e', 'f' };
+  static final int REPLACEMENT_CHARACTER = '\ufffd';
 
   Segment head;
   long size;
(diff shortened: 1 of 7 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -793,6 +869,42 @@ public void clear() {
...
+      writeByte(codePoint);
+
+    } else if (codePoint < 0x800) {
+      // Emit a 11-bit code point with 2 bytes.
+      writeByte(codePoint >>  6        | 0xc0); // 110xxxxx
+      writeByte(codePoint       & 0x3f | 0x80); // 10xxxxxx
+
+    } else if (codePoint < 0x10000) {
+      if (codePoint >= 0xd800 && codePoint <= 0xdfff) {
+        throw new IllegalArgumentException(
```
swankjesse: Interesting design decision here. What to do when the input's not quite right:
 - we could throw
 - we could encode it (it's still recoverable on the other end, though encoders might return the replacement char)
 - we could encode the replacement character ourselves

Thoughts?

swankjesse:
`HttpUrl` wants this. It's defined in terms of code points, and wants to parse codepoint-by-codepoint. I'm currently doing it Java char-by-char, but it's awkward.
https://github.com/square/okhttp/issues/1486

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -173,6 +173,20 @@
   String readUtf8LineStrict() throws IOException;
 
   /**
+   * Removes and returns a single UTF-8 code point, reading between 1 and 4 bytes as necessary.
+   *
+   * <p>If this source is exhausted before a complete code point can be read, this throws an {@link
+   * java.io.EOFException} and consumes no input.
```
JakeWharton: Nice doc. I don't think we're very good about documenting the behavior when methods throw (related to the contents of the underlying buffer).

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -203,6 +203,10 @@ public RealBufferedSource(Source source) {
     return buffer.readUtf8Line(newline);
   }
 
+  @Override public int readUtf8CodePoint() throws IOException {
+    return buffer.readUtf8CodePoint();
```
JakeWharton: You need `request`/`require` logic here.
```java
Source s = new Buffer().writeByte('a');
BufferedSource bs = Okio.buffer(source);
bs.readUtf8CodePoint(); // <-- throws EOF
```

okio/src/main/java/okio/Buffer.java
```diff
@@ -626,6 +627,81 @@ String readUtf8Line(long newline) throws EOFException {
...
+    int min;
+
+    if ((b0 & 0x80) == 0) {
+      // 0xxxxxxx.
+      codePoint = b0 & 0x7f;
+      byteCount = 1; // 7 bits (ASCII).
+      min = 0x0;
+
+    } else if ((b0 & 0xe0) == 0xc0) {
+      // 0x110xxxxx
```
JakeWharton: You can use binary literals if you want:
```java
((b0 & 0b1110000) == 0b1100000
```

okio/src/main/java/okio/Buffer.java
```diff
@@ -793,6 +869,42 @@ public void clear() {
...
+      // Emit a 7-bit code point with 1 byte.
+      writeByte(codePoint);
+
+    } else if (codePoint < 0x800) {
+      // Emit a 11-bit code point with 2 bytes.
+      writeByte(codePoint >>  6        | 0xc0); // 110xxxxx
+      writeByte(codePoint       & 0x3f | 0x80); // 10xxxxxx
+
+    } else if (codePoint < 0x10000) {
+      if (codePoint >= 0xd800 && codePoint <= 0xdfff) {
```
JakeWharton: You use hex here for the values but chars above.

okio/src/main/java/okio/Buffer.java
JakeWharton: Tough. Option 2 doesn't seem reasonable to me. I'm inclined to lean towards 1 (as implemented), but I see the value in 3. I think it's hard to be _too_ forgiving at this level. A higher-level API could handle the replacement character writing, but when you're at this level I think throwing is correct.

JakeWharton:
LGTM

okio/src/main/java/okio/RealBufferedSource.java
JakeWharton: Note: `BufferedSourceTest` would catch this!

swankjesse: Done.

Also changed BufferedSourceTest to support a special buffered source that receives data one byte at a time. That found a bug in readFully's exception handling.

okio/src/main/java/okio/Buffer.java
swankjesse: Oooh, that's interesting. I might circle back on that. I'm going to leave it as-is for now though.

okio/src/main/java/okio/Buffer.java
swankjesse: Good catch. Fixed to use integers in both places.

okio/src/main/java/okio/RealBufferedSource.java
swankjesse: (strangely, I had it the right way initially, but got confused by the fact that I'm throwing `EOFException` anyway when the buffer is exhausted. Forgot about refilling the buffer.)

PR #148, opened 2015-05-16
Change Timeout.throwIfReached() to throw InterruptedIOException

swankjesse:
Previously this was throwing IOException, but nothing was anticipated
that. This is slightly semantically incorrect; the thread wasn't
interrupted. But it's much more convenient to use a single exception
type for both timeouts.

Also add a new type, ForwardingTimeout.

### DIFF
--- a/okio/src/main/java/okio/ForwardingTimeout.java
+++ b/okio/src/main/java/okio/ForwardingTimeout.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright (C) 2015 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/** A {@link Timeout} which forwards calls to another. Useful for subclassing. */
+public class ForwardingTimeout extends Timeout {
+  private Timeout delegate;
+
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/ForwardingTimeout.java
```diff
@@ -0,0 +1,72 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+
+/** A {@link Timeout} which forwards calls to another. Useful for subclassing. */
+public class ForwardingTimeout extends Timeout {
```
JakeWharton: The others are abstract

okio/src/main/java/okio/ForwardingTimeout.java
```diff
@@ -0,0 +1,72 @@
...
+  private Timeout delegate;
```
JakeWharton: Do we need to be volatile? Not sure what the thread guarantees are here

okio/src/main/java/okio/ForwardingTimeout.java
```diff
@@ -0,0 +1,72 @@
...
+    if (delegate == null) throw new IllegalArgumentException("delegate == null");
+    this.delegate = delegate;
+  }
+
+  /** {@link Timeout} instance to which this instance is currently delegating. */
+  public final Timeout delegate() {
+    return delegate;
+  }
+
+  public ForwardingTimeout setDelegate(Timeout delegate) {
```
JakeWharton: final to match delegate()? I don't see any value in being able to override this.

JakeWharton:
LGTM

okio/src/main/java/okio/ForwardingTimeout.java
swankjesse: Turns out I can accomplish all I need with just `delegate()` and `setDelegate()`...

okio/src/main/java/okio/ForwardingTimeout.java
swankjesse: These aren't threadsafe

okio/src/main/java/okio/ForwardingTimeout.java
swankjesse: Done.

PR #151, opened 2015-06-02
Implement comparable in ByteString.

swankjesse:


okio/src/main/java/okio/ByteString.java
```diff
@@ -331,6 +333,14 @@ public boolean rangeEquals(int offset, byte[] other, int otherOffset, int byteCo
     return result != 0 ? result : (hashCode = Arrays.hashCode(data));
   }
 
+  @Override public int compareTo(ByteString byteString) {
+    for (int i = 0, size = Math.min(size(), byteString.size()); i < size; i++) {
+      int compare = Integer.compare(getByte(i) & 0xff, byteString.getByte(i) & 0xff);
```
JakeWharton: `Integer.compare` is JDK 8

JakeWharton:
LGTM

okio/src/main/java/okio/ByteString.java
swankjesse: Lame. Fixed!

PR #154, opened 2015-06-08
Make Okio Socket-related methods throw SocketTimeoutException

nfuller:
...instead of InterruptedIOException. This is a backwards compatible
change because SocketTimeoutException extends
InterruptedIOException.

This change is to address:
https://github.com/square/okhttp/issues/1676

### DIFF
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -21,8 +21,10 @@ import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InterruptedIOException;
 import java.io.OutputStream;
 import java.net.Socket;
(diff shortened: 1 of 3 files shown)


okio/src/main/java/okio/Okio.java
```diff
@@ -201,6 +203,14 @@ public static Source source(final Socket socket) throws IOException {
 
   private static AsyncTimeout timeout(final Socket socket) {
     return new AsyncTimeout() {
+      @Override protected IOException newTimeoutException(IOException cause) {
+        InterruptedIOException ioe = new SocketTimeoutException("timeout");
+        if (ioe != null) {
```
swankjesse: I think you intend to be null checking `cause`, not `ioe`. (ioe is never null)

swankjesse:
LGTM. One slight fix required.

nfuller:
Fixed. Good spot.

PR #158, opened 2015-07-07
Add methods for writing a ByteString to a byte[].

JakeWharton:


JakeWharton:
Actually I just realized the code that I want this for is only temporary. It'll eventually switch to writing a `ByteString` to a `BufferedSink`... so feel free to close if you don't think these are of value.

swankjesse:
I think I'd rather a Sink that writes to a user-supplied `byte[]`.

PR #162, opened 2015-07-30
indexOf overloads for matching a series of bytes.

JakeWharton:


okio/src/main/java/okio/Buffer.java
```diff
@@ -1253,6 +1253,28 @@ Segment writableSegment(int minimumCapacity) {
...
   }
 
+  @Override public long indexOf(ByteString bytes) throws IOException {
+    return indexOf(bytes, 0);
+  }
+
+  @Override public long indexOf(ByteString bytes, long fromIndex) throws IOException {
+    int bytesSize = bytes.size();
+    loop:
+    while ((fromIndex = indexOf(bytes.getByte(0), fromIndex)) != -1) {
```
JakeWharton: If you think there's too much going on in this line I can switch to a `while (true)` loop and move it to an `if` check. That was in my original implementation.

chrisjenx: IMO, I think it is easier to read the previous way.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1253,6 +1253,28 @@ Segment writableSegment(int minimumCapacity) {
     return -1L;
   }
 
+  @Override public long indexOf(ByteString bytes) throws IOException {
+    return indexOf(bytes, 0);
+  }
+
+  @Override public long indexOf(ByteString bytes, long fromIndex) throws IOException {
```
swankjesse: crashes if `bytes.size() == 0`

okio/src/main/java/okio/Buffer.java
```diff
@@ -1253,6 +1253,28 @@ Segment writableSegment(int minimumCapacity) {
...
+    int bytesSize = bytes.size();
+    loop:
+    while ((fromIndex = indexOf(bytes.getByte(0), fromIndex)) != -1) {
+      if (!request(fromIndex + bytesSize)) {
+        return -1; // Not enough bytes to match.
+      }
+      for (int i = 1; i < bytesSize; i++) {
```
swankjesse: I'm guessing we want a helper method like this:
```
public boolean rangeMatches(int offset, ByteString other, int otherOffset, int byteCount);
```
That would be this loop!

JakeWharton: Yeah. I thought the same in doing this.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1253,6 +1253,28 @@ Segment writableSegment(int minimumCapacity) {
...
 
+  @Override public long indexOf(ByteString bytes) throws IOException {
+    return indexOf(bytes, 0);
+  }
+
+  @Override public long indexOf(ByteString bytes, long fromIndex) throws IOException {
+    int bytesSize = bytes.size();
+    loop:
+    while ((fromIndex = indexOf(bytes.getByte(0), fromIndex)) != -1) {
+      if (!request(fromIndex + bytesSize)) {
```
swankjesse: If you have a helper regionMatches method, that should take care of this check

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -215,6 +215,21 @@
   long indexOf(byte b, long fromIndex) throws IOException;
 
   /**
+   * Returns the index of the first match for {@code bytes} in the buffer. This expands the buffer
```
swankjesse: Awesome.

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -463,6 +463,26 @@
     assertEquals(15, source.indexOf((byte) 'b', 15));
   }
 
+  @Test public void indexOfByteString() throws IOException {
+    sink.writeUtf8("abacad");
+    assertEquals(2, source.indexOf(ByteString.encodeUtf8("ac")));
+    source.readUtf8(); // Clear stream.
+
+    // Make sure we backtrack and resume searching after partial match.
```
swankjesse: Nice test.

swankjesse:
Nice. Ping me when I should take another look!

okio/src/main/java/okio/Buffer.java
JakeWharton: Turns out we have this on `ByteString` for matching others called `rangeEquals`.

JakeWharton:
PTAL. I'm holding off on adding a public `rangeEquals` because it's complicated to do efficiently on the underlying `byte[]`s. Plus I'd like the operator additions to be atomic in the git history.

swankjesse:
Nice!

chrisjenx:
:+1:

PR #166, opened 2015-08-11
catch exception

pguilbot:
An issue that impact android version up to 4.2.2 cause an exception to be thrown:
java.lang.AssertionError: libcore.io.ErrnoException: getsockname failed: EBADF (Bad file number)
caused by libcore.io.IoBridge.getSocketLocalPort(IoBridge.java:649)
 https://code.google.com/p/android/issues/detail?id=54072

I use okHttp on my application and I get some crashes regarding this issue. 
I cannot catch this exception on my code because this exception is thrown at a deeper level. 
But this exception can be catched by the okio code.
Most of my crashes can be fixed by this pull request.

This fix is related to this discussion : https://github.com/square/okhttp/issues/1684#issuecomment-117123290

### DIFF
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -216,6 +216,9 @@ public final class Okio {
           socket.close();
         } catch (Exception e) {
           logger.log(Level.WARNING, "Failed to close timed out socket " + socket, e);
+        } catch (AssertionError are) {
+          //Catch this exception due to a Firmware issue up to android 4.2.2
+          logger.log(Level.WARNING, "Failed to close timed out socket " + socket, are);
         }
       }
     };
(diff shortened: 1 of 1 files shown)


swankjesse:
Nice.

Two things:
 * I'm anxious about catching the wrong `AssertionError`. Could you change the catch clause to confirm that the cause is not null, and that its message starts with "getsockname failed" ?
 * Could you link to the Android issue in a common the code? It's impossible to write a test case for this so we should document why it exists. That way it won't accidentally get refactored away.

CLA is good.

swankjesse:
Merged manually.

swankjesse:
Also fixed some style nits and used http://rebaseandsqua.sh.

swankjesse:
Yikes, how embarrassing. For for that problem is here: https://github.com/square/okio/pull/167

PR #173, opened 2015-09-22
Try 8 KiB segments.

swankjesse:
We've heard reports that this dramatically increases throughput
for some applications.


JakeWharton:
LGTM

PR #177, opened 2015-12-21
Remove two unused constructor overloads.

JakeWharton:


JakeWharton:
This removes 2 methods bringing the total down to 610.

PR #178, opened 2015-12-21
Remove synthetic accessors for private members in different scopes.

JakeWharton:
This removes 5 methods taking the count to 607 from 612.


swankjesse:
That’s unfortunate ’cause it wants to regress.

JakeWharton:
I'm working on that! https://youtrack.jetbrains.com/issue/IDEA-149645

PR #183, opened 2016-01-15
Revert "Add OSGi metadata to Okio"

swankjesse:
This reverts commit d4da5cce2cd21383d75f171188575d272a353e6c.

### DIFF
--- a/okio/pom.xml
+++ b/okio/pom.xml
@@ -9,7 +9,6 @@
     <version>1.7.0-SNAPSHOT</version>
   </parent>
 
-  <packaging>bundle</packaging>
   <artifactId>okio</artifactId>
   <name>Okio</name>
 
@@ -19,6 +18,7 @@
       <artifactId>animal-sniffer-annotations</artifactId>
       <optional>true</optional>
     </dependency>
+
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
@@ -46,18 +46,6 @@
             <artifactId>java16</artifactId>
             <version>1.1</version>
           </signature>
-        </configuration>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.felix</groupId>
-        <artifactId>maven-bundle-plugin</artifactId>
-        <extensions>true</extensions>
-        <configuration>
-          <instructions>
-            <Bundle-SymbolicName>${project.groupId}.${project.artifactId}</Bundle-SymbolicName>
-            <Bundle-Version>${project.version}</Bundle-Version>
-            <Export-Package>okio.*</Export-Package>
-          </instructions>
         </configuration>
       </plugin>
     </plugins>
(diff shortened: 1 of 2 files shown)


mattprecious:
lgtm

JakeWharton:
Amaze

lburgazzoli:
For my understanding, why it has been removed ?

swankjesse:
We saw people's builds break because okhttp-urlconnection was including a copy of its dependencies.

lburgazzoli:
And how it is related to osgi ? really for my understanding only

swankjesse:
The bundle build was causing more class files to be included in the .jar file than we had expected. Changing from bundle back to the default fixed the problem.

lburgazzoli:
Do you mean the bundle of okhttp-urlconnection? can you share more details
so I can have a look ?

swankjesse:
Yeah. Take a look at the reverted commit in OkHttp, and what .class files are included in the .jar before and after that change.

lburgazzoli:
Ok I see, the problem is that there are different modules with the same package and bnd merges the classes. Any chance to have something like an uberjar (i.e. okhttp-all) with all the modules included ?

swankjesse:
I think that’s more appropriate for applications than libraries. I don’t know which OkHttp modules you’re using, and actually think it’s a bad idea to need all of ’em.

lburgazzoli:
I do not see any problem for okio but for okhttp and retrofit having different jar "exporting" the same package is definitively an issue for OSGi, would you be ok if I make a proper OSGi bundle for okio, okhttp, retrofit as [ServiceMix](https://github.com/apache/servicemix-bundles) bundle ? so if you need an OSGi-fied version you need to use the ServiceMix one

swankjesse:
Sure? As long as the ongoing costs to this project will be minimal.

Is there any way for you to use OSGi without changing all of the libraries you’re using? I think OSGi as carrying unnecessary complexity, and OSGi users are spreading its virus.

lburgazzoli:
It won't touch square's jar/code as new jars will be provided and published under servicemix group, i.e.: 

``` 
org.apache.servicemix.bundles:org.apache.servicemix.bundles.okio:1.6.0_1
```

Sometimes it is possible to automatically inject entries in MANIFEST at run-time but it is OSGi implementation-dependent and far from optimal, best would be to have OSGi support from ground-up but it is not always possible I know :-)

Usually it is enough to use the maven-bundle-plugin or something similar but unfortunately there are edge cases like same package used among different jars and in that case the maintenance cost for project not using OSGi could be a little high.

swankjesse:
Sounds good to me.

lburgazzoli:
FYI,  okio, okhttp and retrofit have been repackaged and available as part of servicemix's budle collection on mavencentral:

http://search.maven.org/#search%7Cga%7C1%7Corg.apache.servicemix.bundles.okio
http://search.maven.org/#search%7Cga%7C1%7Corg.apache.servicemix.bundles.okhttp
http://search.maven.org/#search%7Cga%7C1%7Corg.apache.servicemix.bundles.retrofit

PR #184, opened 2016-01-16
Added ByteString.asByteBuffer().

crazybob:
This avoids an unnecessary copy.


swankjesse:
Two catches:

 * is the buffer immutable? Not sure there's much we can do there?
 * we have a subclass of ByteString that's internally a list of segments. Unfortunately that one will require a copy...

swankjesse:
LGTM!

Issue #186, opened 2016-01-27
Consider primitive to copy data to a second sink when source is read


digitalbuddha:
I'm struggling with reading from a BufferedSource twice. What I'm trying to do is take an okhttp response body and stream it to both a parser and a file.  First I created 2 ```Buffer```instances from same ```BufferedSource``` next I wrap a ```Buffer``` in an ```InputStreamReader``` and gracefully (or so I thought) pass it to gson.  What I get is ```MalformedJsonException``` 

If I however simply do ```body.source().readUtf8Line()``` I get the full raw json.   

full example:
```
        Request request = new Request.Builder()
                .url("https://www.reddit.com/r/aww/new/.json")
                .build();
        response =new OkHttpClient().newCall(request).execute();

                        ResponseBody body = response.body();
                        BufferedSource sourceforFile = body.source();
                        sourceforFile.require(8014);
                        Buffer buffer = sourceforFile.buffer();
                        Buffer sourceForParser = buffer.clone();

                        InputStreamReader inputStreamReader = new InputStreamReader(sourceForParser.inputStream());
                        result= new GsonBuilder()
                                .registerTypeAdapterFactory(new GsonAdaptersModel()).create()
                                .fromJson(inputStreamReader, RedditData.class);

                        File file = new File(getContext().getCacheDir(), "file");
                        BufferedSink sink = Okio.buffer(Okio.sink(file));
                        sink.writeAll(sourceforFile.buffer());
                        diskValue = Okio.buffer(Okio.source(file)).readString(Charset.defaultCharset());
```

 It seems like Buffers are perfectly suited for taking a single source and streaming in multiple direction, would love to finally bring it all together.  Thank you for your help.

digitalbuddha:
I rearranged to instead use the original BufferedSource to create the ```InputStream``` and the parsing step works, but the copied Buffer still only gives me a partial data saved to disk.  Do I need to synchronize or wait for the originalSource to be exhausted prior to using a cloned Buffer?

```
ResponseBody body = response.body();
BufferedSource originalSource = body.source();
originalSource.require(8000);
Buffer copiedSource = originalSource.buffer().clone();

InputStreamReader inputStreamReader = new InputStreamReader(originalSource.inputStream());
result= new GsonBuilder()
        .fromJson(inputStreamReader, RedditData.class);

File file = new File(getContext().getCacheDir(), "file");
BufferedSink sink = Okio.buffer(Okio.sink(file));
sink.writeAll(copiedSource);
[...]

PR #187, opened 2016-01-27
Feature/tee

digitalbuddha:
Would appreciate suggestion of how to test this. Addresses https://github.com/square/okio/issues/186

### DIFF
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -243,9 +243,7 @@ public final class Okio {
         && e.getMessage().contains("getsockname failed");
   }
 
-  public static Source tee(final Source source, final BufferedSink cacheBody)
-  {
+  public static Source tee(final Source source, final BufferedSink cacheBody) {
     return new TeeSource(source, cacheBody);
   }
-
 }
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -243,7 +243,7 @@ public final class Okio {
         && e.getMessage().contains("getsockname failed");
   }
 
-  public static Source tee(final Source source, final BufferedSink cacheBody) {
-    return new TeeSource(source, cacheBody);
+  public static Source tee(final Source source, final BufferedSink copySink) {
+    return new TeeSource(source, copySink);
   }
 }
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -243,7 +243,7 @@ public final class Okio {
         && e.getMessage().contains("getsockname failed");
   }
 
-  public static Source tee(final Source source, final BufferedSink copySink) {
+  public static Source tee(final Source source, final Sink copySink) {
     return new TeeSource(source, copySink);
   }
 }
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -242,4 +242,10 @@ public final class Okio {
     return e.getCause() != null && e.getMessage() != null
         && e.getMessage().contains("getsockname failed");
   }
+
+  public static Source tee(final Source source, final BufferedSink cacheBody)
+  {
+    return new TeeSource(source, cacheBody);
+  }
+
 }
--- a/okio/src/main/java/okio/TeeSource.java
+++ b/okio/src/main/java/okio/TeeSource.java
@@ -6,11 +6,11 @@ import java.io.IOException;
 class TeeSource implements Source {
 
   private final Source source;
-  private final BufferedSink cacheBody;
+  private final BufferedSink sink;
 
-  public TeeSource(Source source, BufferedSink cacheBody) {
+  public TeeSource(Source source, BufferedSink copySink) {
     this.source = source;
-    this.cacheBody = cacheBody;
+    this.sink = copySink;
   }
 
   @Override
@@ -19,12 +19,12 @@ class TeeSource implements Source {
     bytesRead = source.read(sink, byteCount);
 
     if (bytesRead == -1) {
-      cacheBody.close(); // The cache response is complete!
+      this.sink.close(); // The cache response is complete!
       return -1;
     }
 
-    sink.copyTo(cacheBody.buffer(), sink.size() - bytesRead, bytesRead);
-    cacheBody.emitCompleteSegments();
+    sink.copyTo(this.sink.buffer(), sink.size() - bytesRead, bytesRead);
+    this.sink.emitCompleteSegments();
     return bytesRead;
   }
 
@@ -35,7 +35,7 @@ class TeeSource implements Source {
 
   @Override
   public void close() throws IOException {
-    cacheBody.close();
+    sink.close();
     source.close();
   }
 }
--- a/okio/src/main/java/okio/TeeSource.java
+++ b/okio/src/main/java/okio/TeeSource.java
@@ -16,8 +16,12 @@ class TeeSource implements Source {
   @Override
   public long read(Buffer sink, long byteCount) throws IOException {
     long bytesRead = source.read(sink, byteCount);
-    sink.copyTo(copySink.buffer(), sink.size() - bytesRead, bytesRead);
-    copySink.emitCompleteSegments();
+
+    if (bytesRead > 0) {
+      sink.copyTo(copySink.buffer(), sink.size() - bytesRead, bytesRead);
+      copySink.emitCompleteSegments();
+    }
+
     return bytesRead;
   }
 
--- a/okio/src/main/java/okio/TeeSource.java
+++ b/okio/src/main/java/okio/TeeSource.java
@@ -1,0 +1,41 @@
+package okio;
+
+import java.io.IOException;
+
+
+class TeeSource implements Source {
+
+  private final Source source;
+  private final BufferedSink cacheBody;
+
+  public TeeSource(Source source, BufferedSink cacheBody) {
+    this.source = source;
+    this.cacheBody = cacheBody;
+  }
+
+  @Override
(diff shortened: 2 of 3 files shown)


okio/src/main/java/okio/Okio.java
```diff
@@ -242,4 +245,50 @@ static boolean isAndroidGetsocknameError(AssertionError e) {
...
+
+      @Override
+      public Timeout timeout() {
+        return source.timeout();
+      }
+
+      @Override
+      public void close() throws IOException {
+        if (!cacheRequestClosed
+                && !discard(this, DISCARD_STREAM_TIMEOUT_MILLIS, MILLISECONDS)) {
```
JakeWharton: I'm not sure this should be part of the feature itself. If you fail to exhaust the stream then I don't think it's Okio's job to try and continue it for you. The 100ms policy is of OkHttp requests, and doesn't make sense everywhere.

okio/src/main/java/okio/Util.java
```diff
@@ -71,10 +76,96 @@ public static void sneakyRethrow(Throwable t) {
...
+    } finally {
+      if (originalDuration == Long.MAX_VALUE) {
+        source.timeout().clearDeadline();
+      } else {
+        source.timeout().deadlineNanoTime(now + originalDuration);
+      }
+    }
+  }
+
+  public static Source tree(final Source source, final BufferedSink cacheBody)
```
JakeWharton: Delete?

okio/src/main/java/okio/Okio.java
digitalbuddha: removed second part of if statement

okio/src/main/java/okio/Util.java
digitalbuddha: oops copy pasta

okio/src/main/java/okio/Util.java
```diff
@@ -27,33 +31,33 @@ private Util() {
   public static void checkOffsetAndCount(long size, long offset, long byteCount) {
     if ((offset | byteCount) < 0 || offset > size || size - offset < byteCount) {
       throw new ArrayIndexOutOfBoundsException(
-          String.format("size=%s offset=%s byteCount=%s", size, offset, byteCount));
+              String.format("size=%s offset=%s byteCount=%s", size, offset, byteCount));
```
swankjesse: Please don’t reformat our code!

okio/src/main/java/okio/Util.java
```diff
@@ -71,10 +75,50 @@ public static void sneakyRethrow(Throwable t) {
...
 
   public static boolean arrayRangeEquals(
-      byte[] a, int aOffset, byte[] b, int bOffset, int byteCount) {
+          byte[] a, int aOffset, byte[] b, int bOffset, int byteCount) {
     for (int i = 0; i < byteCount; i++) {
       if (a[i + aOffset] != b[i + bOffset]) return false;
     }
     return true;
   }
+
```
swankjesse: Are these new methods unused? I don’t see the motivation for ’em

okio/src/main/java/okio/Okio.java
```diff
@@ -242,4 +244,49 @@ static boolean isAndroidGetsocknameError(AssertionError e) {
...
+      private boolean cacheRequestClosed;
+
+      @Override
+      public long read(Buffer sink, long byteCount) throws IOException {
+        long bytesRead;
+        try {
+          bytesRead = source.read(sink, byteCount);
+        } catch (IOException e) {
+          if (!cacheRequestClosed) {
+            cacheRequestClosed = true;
```
swankjesse: Seems like this prevents the sink from being closed?

okio/src/main/java/okio/Okio.java
```diff
@@ -242,4 +244,49 @@ static boolean isAndroidGetsocknameError(AssertionError e) {
...
+
+      @Override
+      public Timeout timeout() {
+        return source.timeout();
+      }
+
+      @Override
+      public void close() throws IOException {
+        if (!cacheRequestClosed) {
+          cacheRequestClosed = true;
```
swankjesse: this should actually close the cacheBody, right?

digitalbuddha:
```OKIO.tee``` seemed like a bloated method, create ```TeeSource``` better or worse? Also removed all the fluff (I think).  Seems to work for my use case at least.

File: okio/src/main/java/okio/Okio.java [hunk 7]
```diff
@@ -242,4 +242,8 @@ static boolean isAndroidGetsocknameError(AssertionError e) {
     return e.getCause() != null && e.getMessage() != null
         && e.getMessage().contains("getsockname failed");
   }
+
+  public static Source tee(final Source source, final BufferedSink cacheBody) {
```
JakeWharton: This should be named something other than `cacheBody`. Maybe `copySink`?

Also does it need to be a `BufferedSink`? I think ideally this is just a normal `Sink` (which we may buffer internally).

okio/src/main/java/okio/TeeSource.java
```diff
@@ -0,0 +1,41 @@
...
+    this.cacheBody = cacheBody;
+  }
+
+  @Override
+  public long read(Buffer sink, long byteCount) throws IOException {
+    long bytesRead;
+    bytesRead = source.read(sink, byteCount);
+
+    if (bytesRead == -1) {
+      cacheBody.close(); // The cache response is complete!
```
JakeWharton: Not sure we want to retain this close here from the cache code for the general case.

okio/src/main/java/okio/TeeSource.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.io.IOException;
+
+
+class TeeSource implements Source {
+
+  private final Source source;
+  private final BufferedSink cacheBody;
+
+  public TeeSource(Source source, BufferedSink cacheBody) {
```
digitalbuddha: is a Factory method or builder preferred here?

okio/src/main/java/okio/TeeSource.java
swankjesse: Hmmm . . . my guess is that we probably do? Could also add a boolean configuration option. Tough question.

okio/src/main/java/okio/Okio.java
digitalbuddha: renaming to ```copySink```, as for type `, we won't have access to ```copySink.buffer()``` with ``Sink```

digitalbuddha:
@JakeWharton renamed, not sure what to do about ```if (bytesRead == -1)``` piece, let me know how you want me to proceed and thank you for making my code less embarrassing  :-)

digitalbuddha:
Anything besides squash still needed?

okio/src/main/java/okio/Okio.java
JakeWharton: You can use `Okio.buffer` to wrap a `Sink` with a `Buffer`. There doesn't seem to be a reason that one would need to be passed in.

okio/src/main/java/okio/TeeSource.java
JakeWharton: You don't think it's surprising behavior to close the tee sink when application code never called close on this source? Not calling close on the source would be a bug (and potential leak).

For OkHttp's use, we can implement this early auto-close behavior on top of this Okio feature.

digitalbuddha: make sense removing sink close from feature.  thank you.

okio/src/main/java/okio/Okio.java
```diff
@@ -242,4 +242,8 @@ static boolean isAndroidGetsocknameError(AssertionError e) {
     return e.getCause() != null && e.getMessage() != null
         && e.getMessage().contains("getsockname failed");
   }
+
+  public static Source tee(final Source source, final Sink copySink) {
```
artem-zinnatullin: Let's please test resulting `Source`, since `TeeSource` is internal we can cover `Okio.tee()` instead of `TeeSource`!

artem-zinnatullin: Also, needs javadoc

digitalbuddha:
added ```if (bytesRead > 0) ``` around copy and @artem-zinnatullin added tests. Anything else needed?

digitalbuddha:
I hit a bit of a wall implementing.  

 ```
Source cacheWritingSource = TeeSource.of(response.body().source(), sink);
            teedResponse = response.newBuilder()
                    .body(new RealResponseBody(response.headers(), Okio.buffer(cacheWritingSource)))
                    .build();
...
//somewhere else in code
 BufferedSource source = teedResponse.body().source();
            InputStream in = source.inputStream();
            reader = new InputStreamReader(in);
            Foo foo = new Gson().fromJson(response.body().charStream(),Foo.class);
```

Everything seems to work great except I can't figure out how to close the response's underlying ```Source```.  I know that it is a ```RealBufferedSource``` but I can't cast to that package local final class.

swankjesse:
My apologies for sitting on this. I think our next step is to build out a _recipes_ directory in Okio and make this into our first recipe. I would love to make this a public API but unfortunately I don’t think we can come up with an error handling strategy that’ll work universally.

digitalbuddha:
Do you have any suggestions of how to close the original source? That seems to be only issue I see.

swankjesse:
That plus what to do when writing to the sink throws.

digitalbuddha:
Wanted to touch base again on whether there is a convenient way to close the original source when wrapping in a `TeeSource`.

swankjesse:
Hey, yeah sorry for the radio silence on this. I think we want this functionality, but I’m wrestling with some ideas on how to make it really efficient.

swankjesse:
We’re going to write this as a recipe instead. The core problem is that each tee use case may have different requirements for handling source errors, source exhausted, and source closing.

PR #194, opened 2016-02-17
Limit snapshot byte strings to requested size.

### LINKED ISSUE #192: Buffer.snapshot(size) can return a ByteString longer than size
It seems this is due to reusing the internal segments of the buffer. I think the behaviour is fine, should just be documented accordingly.

JakeWharton:
Closes #192. #Closes #193.


JakeWharton:
@swankjesse ping!

Issue #185, opened 2016-01-26
Watchdog thread is eating CPU


swankjesse:
http://stackoverflow.com/questions/34997596/abnormal-cpu-usage-okio-watchdog

DavidEdwards:
I am able to reproduce this consistently in my current project. I haven't used OkHttp (okio) before this project. I cannot guarantee that it isn't a problem introduced by me.

I'll post an example of how I use OkHttp (okio) shortly.

DavidEdwards:
How I am starting the trace, `onPause`:

```java
Debug.startMethodTracing("debug", 80 * 1024 * 1024); // 80 MiB trace file
new Handler(Looper.getMainLooper()).postDelayed(new Runnable() {
	@Override
	public void run() {
		Debug.stopMethodTracing();
	}
}, 30000);
```

A barebones example of how I use the library:

```java
public class Api {

	private static Api mApi = null;
	private OkHttpClient mClient = null;
	private WeakReference<Context> mContext = null;

	private final boolean mTestDelay = false;

	private static final String API_GET_USER = "http://example.com/api/userinfo.php";

	private static String VERSION_CODE = "0";
	private static String VERSION_VERSION = "1.0";

	@NonNull
	public static Api getInstance() {
		return getInstance(OTRApplication.sContext.get());
	}

	@NonNull
	public static Api getInstance(@NonNull Context context) {
		if (mApi == null) {
			mApi = new Api(context);
		}
		if(mApi.mClient == null) {
			OkHttpClient.Builder builder = new OkHttpClient.Builder();

			mApi.mClient = new OkHttpClient();

			File cacheDir = new File(context.getCacheDir(), "okhttp");
			if(!cacheDir.mkdirs()) {
				cacheDir = context.getCacheDir();
			}
			int cacheSize = 1 * 1024 * 1024; // 1 MiB
			Cache cache = new Cache(cacheDir, cacheSize);

			builder.connectTimeout(15, TimeUnit.SECONDS);
			builder.cookieJar(new NotPersistentCookieJar());
			builder.cache(cache);

			mApi.mClient = builder.build();
		}
		mApi.mContext = new WeakReference<Context>(context);

		return mApi;
	}

	public Api(@NonNull Context context) {
		try {
			PackageInfo pInfo = context.getPackageManager().getPackageInfo(context.getPackageName(), 0);
			VERSION_CODE = String.valueOf(pInfo.versionCode);
			VERSION_VERSION = pInfo.versionName;
[...]

PR #198, opened 2016-02-24
Simplifies the Watchdog thread impl by using a DelayQueue.

jnlopar:
Also lowers thread priority and adds an explicit yield between processed items.
I'm hoping this can ameliorate or at least help narrow down CPU issues related to #185 . I wasn't able to repro as severely as some of the original reporters, but this does seem to hit the watchdog less. Any thoughts on better ways to verify this would be awesome.

### DIFF
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -74,18 +74,17 @@ public class AsyncTimeout extends Timeout implements Delayed {
       return; // No timeout and no deadline? Don't bother with the queue.
     }
     inQueue = true;
-    scheduleTimeout(this, timeoutNanos, hasDeadline);
+    scheduleTimeout(timeoutNanos, hasDeadline);
   }
 
-  private void scheduleTimeout(
-      AsyncTimeout node, long timeoutNanos, boolean hasDeadline) {
+  private void scheduleTimeout(long timeoutNanos, boolean hasDeadline) {
     // Start the watchdog thread and create the head node when the first timeout is scheduled.
     // Double-check lock to initialize the watchdog thread.
     if (queue == null) {
       synchronized (watchdogLock) {
         if (queue == null) {
           queue = new DelayQueue<AsyncTimeout>();
-          final Watchdog watchdog = new Watchdog();
+          Watchdog watchdog = new Watchdog();
           watchdog.setPriority(Thread.MIN_PRIORITY);
           watchdog.start();
         }
@@ -96,31 +95,31 @@ public class AsyncTimeout extends Timeout implements Delayed {
     if (timeoutNanos != 0 && hasDeadline) {
       // Compute the earliest event; either timeout or deadline. Because nanoTime can wrap around,
       // Math.min() is undefined for absolute values, but meaningful for relative ones.
-      node.timeoutAt = now + Math.min(timeoutNanos, node.deadlineNanoTime() - now);
+      timeoutAt = now + Math.min(timeoutNanos, deadlineNanoTime() - now);
     } else if (timeoutNanos != 0) {
-      node.timeoutAt = now + timeoutNanos;
+      timeoutAt = now + timeoutNanos;
     } else if (hasDeadline) {
-      node.timeoutAt = node.deadlineNanoTime();
+      timeoutAt = deadlineNanoTime();
     } else {
       throw new AssertionError();
     }
 
     // Insert the node in timeout order.
-    queue.offer(node);
+    queue.offer(this);
   }
 
   /** Returns true if the timeout occurred. */
   public final boolean exit() {
     if (!inQueue) return false;
     inQueue = false;
-    return cancelScheduledTimeout(this);
+    return cancelScheduledTimeout();
   }
 
   /** Returns true if the timeout occurred. */
-  private boolean cancelScheduledTimeout(AsyncTimeout node) {
-    // Remove the node from the linked list.  If the node wasn't found in the queue, it must have
+  private boolean cancelScheduledTimeout() {
+    // Remove the node from the queue.  If the node wasn't found in the queue, it must have
     // timed out!
-    return !queue.remove(node);
+    return !queue.remove(this);
   }
 
   /**
@@ -137,10 +136,7 @@ public class AsyncTimeout extends Timeout implements Delayed {
 
   @Override
   public final int compareTo(Delayed that) {
-    if (that instanceof AsyncTimeout) {
-      return (int) (timeoutAt - ((AsyncTimeout) that).timeoutAt);
-    }
-    return (int) (getDelay(TimeUnit.NANOSECONDS) - that.getDelay(TimeUnit.NANOSECONDS));
+    return (int) (timeoutAt - ((AsyncTimeout) that).timeoutAt);
   }
 
   /**
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -54,12 +54,8 @@ public class AsyncTimeout extends Timeout implements Delayed {
   /**
    * The watchdog thread processes a queue of pending timeouts, sorted in the order to be triggered.
    */
-  private static class QueueHolder {
-    private static final DelayQueue<AsyncTimeout> INSTANCE = new DelayQueue<AsyncTimeout>();
-  }
-
   private static final Object watchdogLock = new Object();
-  private static volatile Watchdog watchdog;
+  private static volatile DelayQueue<AsyncTimeout> queue;
 
   /** True if this node is currently in the queue. */
   private boolean inQueue;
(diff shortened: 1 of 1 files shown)


okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -49,15 +52,14 @@
...
-   * the order to be triggered. This class synchronizes on AsyncTimeout.class.
-   * This lock guards the queue.
-   *
-   * <p>Head's 'next' points to the first element of the linked list. The first
-   * element is the next node to time out, or null if the queue is empty. The
-   * head is null until the watchdog thread is started.
+   * The watchdog thread processes a queue of pending timeouts, sorted in the order to be triggered.
    */
-  private static AsyncTimeout head;
+  private static class QueueHolder {
```
JakeWharton: Why the indirection here instead of just having the queue exist as a static object on `AsyncTimeout`?

jnlopar: good point. I didn't want to initialize it if the watchdog was never used. I added it to the watchdog initialization.

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -151,6 +130,24 @@ private long remainingNanos(long now) {
...
   }
 
+  @Override
+  public final long getDelay(TimeUnit unit) {
+    return unit.convert(timeoutAt - System.nanoTime(), TimeUnit.NANOSECONDS);
+  }
+
+  @Override
+  public final int compareTo(Delayed that) {
+    long thisDelay = getDelay(TimeUnit.NANOSECONDS);
```
ejona86: Hmm... It'd be nice to use `timeoutAt` instead, to avoid calls to `System.nanoTime()`. It does cause some problem with `AsyncTimeout` being public, since it would be more convenient to assume `that instanceof AsyncTimeout`.

jnlopar: good call... i made it prefer we're an `AsyncTimeout` but fall back to comparing getDelay

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -79,12 +77,19 @@ public final void enter() {
...
-    if (head == null) {
-      head = new AsyncTimeout();
-      new Watchdog().start();
+    // Double-check lock to initialize the watchdog thread.
+    if (queue == null) {
+      synchronized (watchdogLock) {
+        if (queue == null) {
+          queue = new DelayQueue<AsyncTimeout>();
+          final Watchdog watchdog = new Watchdog();
+          watchdog.setPriority(Thread.MIN_PRIORITY);
```
swankjesse: Why MIN PRIORITY ?

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -79,12 +77,19 @@ public final void enter() {
     scheduleTimeout(this, timeoutNanos, hasDeadline);
   }
 
-  private static synchronized void scheduleTimeout(
+  private void scheduleTimeout(
       AsyncTimeout node, long timeoutNanos, boolean hasDeadline) {
```
swankjesse: is `node` ever not `this` ?

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -151,6 +130,19 @@ private long remainingNanos(long now) {
...
+  public final long getDelay(TimeUnit unit) {
+    return unit.convert(timeoutAt - System.nanoTime(), TimeUnit.NANOSECONDS);
+  }
+
+  @Override
+  public final int compareTo(Delayed that) {
+    if (that instanceof AsyncTimeout) {
+      return (int) (timeoutAt - ((AsyncTimeout) that).timeoutAt);
+    }
+    return (int) (getDelay(TimeUnit.NANOSECONDS) - that.getDelay(TimeUnit.NANOSECONDS));
```
swankjesse: Not sure how I feel about this. I think it might be simplest just to throw a ClassCastException in this case.

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -310,50 +302,13 @@ public Watchdog() {
...
-
-          // Didn't find a node to interrupt. Try again.
-          if (timedOut == null) continue;
-
+          final AsyncTimeout timedOut = queue.take();
           // Close the timed out node.
           timedOut.timedOut();
         } catch (InterruptedException ignored) {
         }
+        Thread.yield();
```
swankjesse: Why?

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -49,15 +52,10 @@
...
-   * This lock guards the queue.
-   *
-   * <p>Head's 'next' points to the first element of the linked list. The first
-   * element is the next node to time out, or null if the queue is empty. The
-   * head is null until the watchdog thread is started.
+   * The watchdog thread processes a queue of pending timeouts, sorted in the order to be triggered.
    */
-  private static AsyncTimeout head;
+  private static final Object watchdogLock = new Object();
+  private static volatile DelayQueue<AsyncTimeout> queue;
```
swankjesse: is volatile helpful or harmful here?

okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -79,12 +77,19 @@ public final void enter() {
...
     // Start the watchdog thread and create the head node when the first timeout is scheduled.
-    if (head == null) {
-      head = new AsyncTimeout();
-      new Watchdog().start();
+    // Double-check lock to initialize the watchdog thread.
+    if (queue == null) {
+      synchronized (watchdogLock) {
+        if (queue == null) {
+          queue = new DelayQueue<AsyncTimeout>();
+          final Watchdog watchdog = new Watchdog();
```
swankjesse: nit: why final ?

swankjesse:
I’d really like to confirm what the problem was before making big structural changes. This is a nice simplification in the code, but it actually increases the overhead of taking a timeout. DelayQueue is a lot more powerful than we need because it supports multiple concurrent calls to `take()`, that means it needs _another_ set of locks with a reentrant lock, which increases the book-keeping substantially.

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: So my thought was, if there are multiple items immediately on the queue to take, this could be executing for awhile with the loop. If we yield out the thread we would block other threads less.

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: meh, i prefer it on my code, but happy to adjust since this is your codebase :)

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: there's an explanation here: http://stackoverflow.com/a/7855774
basically it's possible for the JVM to reorder things if volatile isn't used for the DCL here. the volatile on the queue should still be much cheaper than the static synchronized lock on the whole class monitor lock.

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: nope. same deal on the cancel call, so i took it out of there too. good catch!

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: sure, that simplifies it anyway

jnlopar:
@swankjesse I totally agree, I'm trying to find the best way to consistently repo it. I've been using systrace and this does contend less, but I'll try to do a forensic reconstruction of the degenerate state using some of the sample code in the original SO post and then post back my findings. My theory is that this should still be less overhead because there will is less overall blocking and contention.

jnlopar:
@swankjesse I think I realized what's going on with the original code, at least, and possibly contributing to/causing the initial bug.
https://github.com/square/okio/blob/c8ca861739c0a69ad0ecb43e1c0a51546fddf3cd/okio/src/main/java/okio/AsyncTimeout.java#L82

so that's synchronized on the class, but it's also the mechanism to enqueue a new timeout via `enter()`... the code is supposed to notify if something is newly added to the head of the queue to run, but if something is added with a timeout earlier than the current head of the queue, the `wait` call from that item is going to block it at least until that one has timed out.

Also, I think the only reason the main notify is working at all is because head is set on startup, and notified, which seems to balance out the eventual 'wait for head to be enqueued' here: https://github.com/square/okio/blob/c8ca861739c0a69ad0ecb43e1c0a51546fddf3cd/okio/src/main/java/okio/AsyncTimeout.java#L337

So, it seems like `DelayQueue` is a straightforward implementation of what you're trying to do here. If you're still concerned about the lock book-keeping I could try to rewrite it with a custom lock. Wdyt?

jnlopar:
Oh, this also seems to indicate that any call to cancelTimeout, being also synchronized, would also block until the class monitor is unlocked, which would be the next thing expiring or notifying and addition at head.

okio/src/main/java/okio/AsyncTimeout.java
jnlopar: On Android the watchdog is running at the same priority as the main/UI thread, while it really should be running as a background thread. Is there a value between min/default that I should use? Otherwise min seems to be the best choice here.

jnlopar:
I'm going to try writing up some unit tests based on the ordering of timeouts to test some of this. I realized I might have been misinterpreting the lock wait/notify flow here, so test is best. :)

swankjesse:
I think the delay queue is a higher, more powerful level of abstraction than what we need here. I’d like to close this PR and go back to figuring out what’s causing issue #185. At the very least I’ll learn something.

jnlopar:
sounds good, i'll go ahead and close. thanks for looking into this!

PR #203, opened 2016-04-03
Make Source.indexOf(ByteString) significantly faster.

swankjesse:
Previously the algorithm that did this was extremely inefficient, and had worst
case runtime of O(N * S * S) for N is size of the bytestring and S is the number
of segments.

The new code runs in O(N * S). It accomplishes this by not starting each
search at the first segment, which could occur many times when called by
RealBufferedSource.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -1259,16 +1259,52 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public long indexOf(ByteString bytes, long fromIndex) throws IOException {
     if (bytes.size() == 0) throw new IllegalArgumentException("bytes is empty");
-    while (true) {
-      fromIndex = indexOf(bytes.getByte(0), fromIndex);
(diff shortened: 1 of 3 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -1259,16 +1259,52 @@ Segment writableSegment(int minimumCapacity) {
...
+        s = s.next;
+        offset = nextOffset;
+      }
+    }
+
+    // Scan through the segments, searching for the lead byte. Each time that is found, delegate to
+    // rangeEquals() to check for a complete match.
+    byte b0 = bytes.getByte(0);
+    int bytesSize = bytes.size();
+    long resultLimit = (int) (size - bytesSize) + 1;
```
JakeWharton: Is the `int` cast leftover from something else?

JakeWharton:
LGTM

okio/src/main/java/okio/Buffer.java
swankjesse: Yeah, nice catch.

PR #205, opened 2016-04-06
Optimize indexOfElement.

swankjesse:


okio/src/main/java/okio/Buffer.java
```diff
@@ -1314,27 +1344,73 @@ Segment writableSegment(int minimumCapacity) {
   @Override public long indexOfElement(ByteString targetBytes, long fromIndex) {
     if (fromIndex < 0) throw new IllegalArgumentException("fromIndex < 0");
 
-    Segment s = head;
-    if (s == null) return -1L;
-    long offset = 0L;
-    byte[] toFind = targetBytes.toByteArray();
```
swankjesse: interesting fact! This used to allocate! I bet that wasn’t fast.

JakeWharton:
Looks good :+1:

Issue #195, opened 2016-02-17
Efficient Message digest for buffers...


Lakedaemon:
Could we have efficient message digest  for Buffer ? (like updating a sha256 with the content of a buffer)

If I'm not mistaken, sha256 can be computed for a buffer by doing 
buffer.snapshot().sha256()  <-- lots of work in there
while  In the toString implementation of buffer, we have something that looks like way more efficient : 

try {
      MessageDigest md5 = MessageDigest.getInstance("MD5");
      md5.update(head.data, head.pos, head.limit - head.pos);
      for (Segment s = head.next; s != head; s = s.next) {
        md5.update(s.data, s.pos, s.limit - s.pos);
      }
      return String.format("Buffer[size=%s md5=%s]",
          size, ByteString.of(md5.digest()).hex());
    } catch (NoSuchAlgorithmException e) {
      throw new AssertionError();
    }

swankjesse:
Yeah, good idea. I'd like something that'll work on any source or sink.

swankjesse:
This is for sources and sinks:
https://github.com/square/okio/pull/204

Still to do is adding md5(), sha1(), and sha256() methods to Buffer.

PR #207, opened 2016-04-11
Add md5(), sha1(), and sha256() methods to Buffer.

swankjesse:
Make it a little more extravagant.

Closes https://github.com/square/okio/issues/195

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -1400,6 +1400,34 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return result;
   }
 
+  /** Returns the MD5 hash of this buffer. */
+  public ByteString md5() {
+    return digest("MD5");
+  }
+
+  /** Returns the SHA-1 hash of this buffer. */
+  public ByteString sha1() {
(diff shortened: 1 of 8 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -1456,17 +1484,7 @@ private boolean rangeEquals(
...
-      md5.update(head.data, head.pos, head.limit - head.pos);
-      for (Segment s = head.next; s != head; s = s.next) {
-        md5.update(s.data, s.pos, s.limit - s.pos);
-      }
-      return String.format("Buffer[size=%s md5=%s]",
-          size, ByteString.of(md5.digest()).hex());
-    } catch (NoSuchAlgorithmException e) {
-      throw new AssertionError();
-    }
+    return String.format("Buffer[size=%s md5=%s]", size, digest("MD5").hex());
```
JakeWharton: `md5().hex()`

JakeWharton:
LGTM!

okio/src/main/java/okio/Buffer.java
swankjesse: Fixed! Whoops.

PR #211, opened 2016-04-16
Implement BufferedSource.select(List<ByteString>).

swankjesse:
I spent a lot of time working on an alternate version that took a
Selection object. It modeled a 'compiled' variant of the List<String>
that had certain optimizations applied. Unfortunately the benefits
of these optimizations weren't worth the complexity: they improved
performance only when the choices were very long strings.

This API is not particularly extensible. If we ever found optimizations
that require the bytestrings to be sorted, or converted into a trie,
or something else, this API prevents that. From my investigation this
is a reasonable tradeoff; those approaches have been explored and aren't
that compelling!

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -529,6 +529,41 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return new ByteString(readByteArray(byteCount));
   }
 
+  @Override public int select(List<ByteString> byteStrings) {
+    Segment s = head;
+    if (s == null) return byteStrings.indexOf(ByteString.EMPTY);
+
+    for (int i = 0, listSize = byteStrings.size(); i < listSize; i++) {
+      ByteString b = byteStrings.get(i);
+      if (size >= b.size() && rangeEquals(s, s.pos, b, 0, b.size())) {
+        try {
+          skip(b.size());
+          return i;
+        } catch (EOFException e) {
+          throw new AssertionError(e);
+        }
+      }
+    }
+    return -1;
+  }
+
+  /**
+   * Returns the index of a value in {@code byteStrings} that is either the prefix of this buffer,
+   * or that this buffer is a prefix of. Unlike {@link #select} this never consumes the value, even
+   * if it is found in full.
+   */
+  int selectPrefix(List<ByteString> byteStrings) {
+    Segment s = head;
+    for (int i = 0, listSize = byteStrings.size(); i < listSize; i++) {
+      ByteString b = byteStrings.get(i);
+      int bytesLimit = (int) Math.min(size, b.size());
+      if (bytesLimit == 0 || rangeEquals(s, s.pos, b, 0, bytesLimit)) {
+        return i;
+      }
+    }
+    return -1;
+  }
+
   @Override public void readFully(Buffer sink, long byteCount) throws EOFException {
     if (size < byteCount) {
       sink.write(this, size); // Exhaust ourselves.
--- a/okio/src/test/java/okio/BufferedSourceTest.java
+++ b/okio/src/test/java/okio/BufferedSourceTest.java
@@ -810,4 +810,109 @@ public class BufferedSourceTest {
     assertLongDecimalString("-00000000000000009223372036854775808", -9223372036854775808L);
     assertLongDecimalString(TestUtil.repeat('0', Segment.SIZE + 1) + "1", 1);
   }
+
+  @Test public void select() throws IOException {
+    List<ByteString> selection = Arrays.asList(
+        ByteString.encodeUtf8("ROCK"),
+        ByteString.encodeUtf8("SCISSORS"),
+        ByteString.encodeUtf8("PAPER"));
+
+    sink.writeUtf8("PAPER,SCISSORS,ROCK");
+    assertEquals(2, source.select(selection));
+    assertEquals(',', source.readByte());
+    assertEquals(1, source.select(selection));
+    assertEquals(',', source.readByte());
+    assertEquals(0, source.select(selection));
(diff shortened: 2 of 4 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -529,6 +529,41 @@ public byte getByte(long pos) {
     return new ByteString(readByteArray(byteCount));
   }
 
+  @Override public int select(List<ByteString> byteStrings) {
+    Segment s = head;
+    if (s == null) return byteStrings.indexOf(ByteString.EMPTY);
```
JakeWharton: Interesting behavior. Haven't made it to the tests yet, but I would have just expected -1 here.

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -810,4 +810,109 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+        ByteString.encodeUtf8("SCISSORS"),
+        ByteString.encodeUtf8("PAPER"));
+
+    sink.writeUtf8("PAPER,SCISSORS,ROCK");
+    assertEquals(2, source.select(selection));
+    assertEquals(',', source.readByte());
+    assertEquals(1, source.select(selection));
+    assertEquals(',', source.readByte());
+    assertEquals(0, source.select(selection));
+    assertEquals(true, source.exhausted());
```
JakeWharton: nit: could be assertTrue

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -810,4 +810,109 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+    List<ByteString> selection = Arrays.asList(a, bc, bd);
+
+    sink.write(bd);
+    sink.write(a);
+    sink.write(bc);
+
+    assertEquals(2, source.select(selection));
+    assertEquals(0, source.select(selection));
+    assertEquals(1, source.select(selection));
+    assertEquals(true, source.exhausted());
```
JakeWharton: assertTrue

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -810,4 +810,109 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+  @Test public void selectEmptyByteString() throws IOException {
+    List<ByteString> selection = Arrays.asList(ByteString.of());
+    sink.writeUtf8("abc");
+    assertEquals(0, source.select(selection));
+    assertEquals("abc", source.readUtf8());
+  }
+
+  @Test public void selectEmptyByteStringFromEmptySource() throws IOException {
+    List<ByteString> selection = Arrays.asList(ByteString.of());
+    assertEquals(0, source.select(selection));
```
JakeWharton: What motivates this behavior? I'm not opposed to it, but it's different than I would have thought is desired.

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -810,4 +810,109 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+    assertEquals("abc", source.readUtf8());
+  }
+
+  @Test public void selectReturnsFirstByteStringThatMatches() throws IOException {
+    List<ByteString> selection = Arrays.asList(
+        ByteString.encodeUtf8("abcd"),
+        ByteString.encodeUtf8("abc"),
+        ByteString.encodeUtf8("abcde"));
+    sink.writeUtf8("abcdef");
+    assertEquals(0, source.select(selection));
```
JakeWharton: Won't this be problematic for how we want to use this API?

JakeWharton:
LGTM but some questions

okio/src/main/java/okio/Buffer.java
swankjesse: I guess it comes down to whether `"".starts with("")` is true or not. I'd prefer to not make this a special case, but when the Buffer is empty it doesn't have any segments so we can't use the regular flow.

okio/src/test/java/okio/BufferedSourceTest.java
swankjesse: We're going to be searching for strings that have a double quote delimiter, like `ROCK"`, `SCISSORS"`, and `PAPER"`. Otherwise we'd risk ambiguity on prefixes like `FAILED` vs. FAILED_PERMANENTLY`. For none of our use cases will there be options that are prefixes of other options.

okio/src/test/java/okio/BufferedSourceTest.java
swankjesse: It's cause "". start With("") is true.

okio/src/test/java/okio/BufferedSourceTest.java
JakeWharton: Ah, good call. I wasn't thinking that we'd include the trailing quote. Clever.

JakeWharton: That being said, doesn't it still make sense to match the most specific case if possible for the general case?

swankjesse: Tough to say. Attempting to match a longer string could cause the read to block on loading more data. Right now the caller can get whatever behavior she wants: sort the choices for shortest match, reverse sort for linear match, and arbitrary order if that's preferred.

JakeWharton: Yeah. Doing so would favor a trie or equivalent data structure which knew
that it needed to disambiguate.

okio/src/test/java/okio/BufferedSourceTest.java
swankjesse: Fixed x2

PR #215, opened 2016-04-21
Change BufferedSource.select to take an Options type.

swankjesse:
This gives us flexibility going forward in case we decide to more
strict validation on options (no empty options?) or if we want to
precompute a trie or similar.

### DIFF
--- a/okio/src/main/java/okio/Options.java
+++ b/okio/src/main/java/okio/Options.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2016 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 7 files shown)


okio/src/main/java/okio/Options.java
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio;
+
+import java.util.AbstractList;
+
+/** An indexed set of values that may be read with {@link BufferedSource#select}. */
+public final class Options extends AbstractList<ByteString> {
```
JakeWharton: nit: `implements RandomAccess`

JakeWharton:
lgtm!

okio/src/main/java/okio/Options.java
swankjesse: Fixed!

PR #219, opened 2016-05-01
Friendlier toString for ByteStrings and Buffers.

swankjesse:
If the data looks like text, this returns text. Otherwise it returns hex.
MD5 checksums are no longer used.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -644,7 +644,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
       Buffer data = new Buffer();
       copyTo(data, 0, Math.min(32, size));
       throw new EOFException("\\n not found: size=" + size()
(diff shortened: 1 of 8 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -644,7 +644,7 @@ int selectPrefix(Options options) {
       Buffer data = new Buffer();
       copyTo(data, 0, Math.min(32, size));
       throw new EOFException("\\n not found: size=" + size()
-          + " content=" + data.readByteString().hex() + "...");
+          + " content=" + data.readByteString().hex() + "…");
```
JakeWharton: Guava recently moved away from non-ASCII in toString. I forget the reasons, will have to look it up.

JakeWharton:
LGTM. Let me find that Guava thing and see if we care.

swankjesse:
https://github.com/google/guava/issues/2376

PR #227, opened 2016-06-18
Fix a major bug where HashingSource is incorrect for large writes.

swankjesse:
There was a bug where it wasn't traversing through the segments of
the buffer being hashed. This means that HashingSource was returning
incorrect answers for any writes that spanned multiple segment
boundaries.

The problem wasn't detected by our test cases because although
we have tests for non-uniform segments, and for buffers that
span multiple segments, we don't have tests for non-uniform
buffers that span multiple segments. This has been fixed too.

### DIFF
--- a/okio/src/main/java/okio/HashingSource.java
+++ b/okio/src/main/java/okio/HashingSource.java
@@ -71,7 +71,7 @@ public final class HashingSource extends ForwardingSource {
       // Find the first segment that has new bytes.
       long offset = sink.size;
       Segment s = sink.head;
-      while (offset > sink.size - result) {
+      while (offset > start) {
         s = s.prev;
         offset -= (s.limit - s.pos);
       }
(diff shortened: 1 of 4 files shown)


okio/src/main/java/okio/HashingSource.java
```diff
@@ -71,7 +71,7 @@ private HashingSource(Source source, String algorithm) {
       // Find the first segment that has new bytes.
       long offset = sink.size;
       Segment s = sink.head;
-      while (offset > sink.size - result) {
+      while (offset > start) {
```
swankjesse: (just a micro-optimization)

JakeWharton:
Yowza

okio/src/main/java/okio/HashingSource.java
pawatzaza: ok

Issue #218, opened 2016-04-25
Pipe


swankjesse:
It’d be handy to have a class that implements the producer/consumer recipe. Possibly with a maximum size, so the producer can block if the consumer isn't keeping up.

PR #230, opened 2016-06-26
Promote OkHttp's Pipe into Okio.

swankjesse:
Also make Timeout.waitUntilNotified() a public API.


okio/src/main/java/okio/Pipe.java
```diff
@@ -0,0 +1,141 @@
...
+  final Source source = new PipeSource();
+
+  public Pipe(long maxBufferSize) {
+    if (maxBufferSize < 1L) {
+      throw new IllegalArgumentException("maxBufferSize < 1: " + maxBufferSize);
+    }
+    this.maxBufferSize = maxBufferSize;
+  }
+
+  public Source source() {
```
swankjesse: New accessors so that `sink`, `source` fields don’t need to be public.

swankjesse: . . . thoughts? The public members had a very simple struct feel to it.

okio/src/main/java/okio/Timeout.java
```diff
@@ -149,4 +150,48 @@ public void throwIfReached() throws IOException {
       throw new InterruptedIOException("deadline reached");
     }
   }
+
+  /**
+   * Waits on {@code monitor} until it is notified. Throws {@link SocketTimeoutException} if either
+   * the thread is interrupted or if this timeout elapses before {@code monitor} is notified. The
+   * caller must be synchronized on {@code monitor}.
+   */
+  public final void waitUntilNotified(Object monitor) throws IOException {
```
swankjesse: (In OkHttp this is a static method that takes a `Timeout` as a parameter.)

okio/src/main/java/okio/Pipe.java
JakeWharton: Doesn't bother me. Feels more appropriate here anyway.

okio/src/main/java/okio/Timeout.java
```diff
@@ -149,4 +150,79 @@ public void throwIfReached() throws IOException {
...
+        monitor.wait(waitMillis, (int) (waitNanos - waitMillis * 1000000L));
+        elapsedNanos = System.nanoTime() - start;
+      }
+
+      // Throw if the timeout elapsed before the monitor was notified.
+      if (elapsedNanos >= waitNanos) {
+        throw new SocketTimeoutException("timeout");
+      }
+    } catch (InterruptedException e) {
+      throw new SocketTimeoutException("interrupted");
```
JakeWharton: The use of `SocketTimeoutException` in here is unfortunate. Our desired use is leaking. How about letting `InterruptedException` propagate and returning a `boolean` as to whether timeout occurred?

okio/src/main/java/okio/Pipe.java
```diff
@@ -0,0 +1,140 @@
...
+ * been exhausted. At that point reads will return -1, indicating the end of the stream. But if the
+ * source is closed first, writes to the sink will immediately fail with an {@link IOException}.
+ */
+public final class Pipe {
+  final long maxBufferSize;
+  final Buffer buffer = new Buffer();
+  boolean sinkClosed;
+  boolean sourceClosed;
+  final Sink sink = new PipeSink();
+  final Source source = new PipeSource();
```
JakeWharton: Make these two `private`. Force tests to use more of the public API.

JakeWharton:
lgtm. i only scanned the tests since i reviewed them in detail going into OkHttp and I assume they're the same.

okio/src/main/java/okio/Timeout.java
swankjesse: Yeah, this isn’t very good. I was cheating and treating `SocketTimeoutException ` as a `TimeoutException` which is a small lie. Still, I don’t like returning a boolean either here. This thing is trying to package up a timeout abstraction and throwing when it elapses is part of that.

What about keeping the SocketTimeoutException above, and trading this for an InterruptedIOException?

okio/src/main/java/okio/Pipe.java
swankjesse: fixed!

okio/src/main/java/okio/Timeout.java
JakeWharton: `throwIfReached` uses `InterruptedIOException` for both cases. Should this then? 

As to your suggestions specifically, I'm not sure if two checked exceptions is strictly worse than one whose name isn't quite perfect.

swankjesse:
https://github.com/square/okio/issues/218

okio/src/main/java/okio/Timeout.java
swankjesse: InterruptedIOException works for me. Note that SocketTimeoutException is a subclass of that!

okio/src/main/java/okio/Timeout.java
```diff
@@ -149,4 +150,79 @@ public void throwIfReached() throws IOException {
       throw new InterruptedIOException("deadline reached");
     }
   }
+
+  /**
+   * Waits on {@code monitor} until it is notified. Throws {@link SocketTimeoutException} if either
```
JakeWharton: Don't forget to change this javadoc

PR #231, opened 2016-06-27
Add rangeEquals to BufferedSource.

### LINKED ISSUE #165: BufferedSource.rangeEquals
```java
boolean rangeEquals(long offset, ByteString bytes, long bytesOffset, long count);
```
Mentioned in #162.

JakeWharton:
Closes #165.

### DIFF
--- a/okio/src/main/java/okio/RealBufferedSource.java
+++ b/okio/src/main/java/okio/RealBufferedSource.java
@@ -367,6 +367,22 @@ final class RealBufferedSource implements BufferedSource {
     }
   }
 
+  @Override public boolean rangeEquals(long offset, ByteString bytes) throws IOException {
+    return rangeEquals(offset, bytes, 0, bytes.size());
+  }
+
+  @Override public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int count)
+      throws IOException {
+    if (closed) throw new IllegalStateException("closed");
+
+    for (int i = 0; i < count; i++) {
+      long bufferOffset = offset + i;
+      if (!request(bufferOffset + 1)) return false;
+      if (buffer.getByte(bufferOffset) != bytes.getByte(bytesOffset + i)) return false;
+    }
+    return true;
+  }
+
   @Override public InputStream inputStream() {
     return new InputStream() {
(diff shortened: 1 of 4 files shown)


okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -368,6 +368,22 @@
...
+    return rangeEquals(offset, bytes, 0, bytes.size());
+  }
+
+  @Override public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int count)
+      throws IOException {
+    if (closed) throw new IllegalStateException("closed");
+
+    for (int i = 0; i < count; i++) {
+      long bufferOffset = offset + i;
+      if (!request(bufferOffset + 1)) return false;
```
JakeWharton: Not sure if this is worth optimizing for. The alternative is just `request(offset + count) && buffer.rangeEquals(..)`.

swankjesse: Yeah, After thinking about it for 3 seconds I think this is actually necessary for correctness. Otherwise this method could block when it isn’t supposed to.

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -916,4 +917,27 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+    assertTrue(source.rangeEquals(0 , ByteString.encodeUtf8("AAA mannn"), 2, 5));
+    assertTrue(source.rangeEquals(24, ByteString.encodeUtf8("PPPanamaaa"), 2, 6));
+  }
+
+  @Test public void rangeEqualsOnlyReadsUntilMismatch() throws IOException {
+    assumeTrue(factory == Factory.ONE_BYTE_AT_A_TIME); // Other sources read in chunks anyway.
+
+    sink.writeUtf8("A man, a plan, a canal. Panama.");
+    assertFalse(source.rangeEquals(0, ByteString.encodeUtf8("A man.")));
+    assertEquals("A man,", source.buffer().readUtf8());
```
swankjesse: Nice test!

swankjesse:
I really like this.

PR #232, opened 2016-06-27
Add argument validation to rangeEquals.

JakeWharton:


okio/src/main/java/okio/Buffer.java
```diff
@@ -1456,6 +1456,18 @@ Segment writableSegment(int minimumCapacity) {
   }
 
   @Override public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int count) {
+    if (offset < 0) throw new IllegalArgumentException("offset < 0: " + offset);
```
swankjesse: `Util.checkOffsetAndCount()` ?

okio/src/main/java/okio/Buffer.java
```diff
@@ -1456,6 +1456,18 @@ Segment writableSegment(int minimumCapacity) {
...
+    if (bytesOffset < 0) throw new IllegalArgumentException("bytesOffset < 0: " + bytesOffset);
+    if (count < 0) throw new IllegalArgumentException("count < 0: " + count);
+    int bytesSize = bytes.size();
+    if (bytesOffset > bytesSize) {
```
swankjesse: I think you want `(bytesOffset + count > bytesSize)`

okio/src/main/java/okio/Buffer.java
JakeWharton: Gah. It's been so long since I've contributed I forgot all the tricks!

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -940,4 +940,37 @@ private void assertLongDecimalString(String s, long expected) throws IOException
     assertFalse(source.rangeEquals(0, ByteString.encodeUtf8("A man.")));
     assertEquals("A man,", source.buffer().readUtf8());
   }
+
+  @Test public void rangeEqualsArgumentValidation() throws IOException {
+    try {
+      source.rangeEquals(-1, ByteString.encodeUtf8("A"));
+      fail();
+    } catch (IllegalArgumentException e) {
```
swankjesse: e –> expected ?

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -375,6 +375,18 @@
       throws IOException {
     if (closed) throw new IllegalStateException("closed");
 
+    if (offset < 0) throw new IllegalArgumentException("offset < 0: " + offset);
+    if (bytesOffset < 0) throw new IllegalArgumentException("bytesOffset < 0: " + bytesOffset);
+    if (count < 0) throw new IllegalArgumentException("count < 0: " + count);
+    int bytesSize = bytes.size();
+    if (bytesOffset > bytesSize) {
```
swankjesse: Same. You want `(bytesOffset + count > bytesSize)`

swankjesse:
LGTM with comments!

JakeWharton:
Updated. Couldn't use `checkOffsetAndCount` because it would throw when a `Buffer`'s size was insufficient where in this case we just want to return `false`. I also couldn't use it for the `bytes.size()` + `bytesOffset` + `byteCount` because the exception message was misleading. I copied the implementation of the check and adjusted the exception message to match the parameter names.

swankjesse:
One more thing. I just checked ByteString and it's less strict about range bounds. That's consistent with String, and it has the upside that it's a little more ergonomic for the caller.

```
public boolean isPrefixOfHello(Buffer b) {
  return b.rangeMatches(0, HELLO, 0, b.size())
}
```

swankjesse:
(I've certainly  used rangeMatches on strings a few times in ways that take advantage of this.)

PR #233, opened 2016-06-27
Be forgiving in rangeEquals arguments.

JakeWharton:
This matches String.regionMatches and ByteString.rangeEquals behavior.


swankjesse:
Thanks!

PR #235, opened 2016-06-28
Add more test coverage to InputStream.skip.

JakeWharton:


swankjesse:
Is there any behavior change if the skip is longer than the buffer?

JakeWharton:
It was only about 4-6 lines of code to get matching behavior, but probably not worth optimizing then. I updated to expand the test case instead.

PR #237, opened 2016-06-28
Add Reader and Writer facades.

JakeWharton:


okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +173,54 @@ public long size() {
...
+      }
+    };
+  }
+
+  int readChars(Charset charset, char[] cbuf, int off, int len) {
+    Util.checkOffsetAndCount(cbuf.length, off, len);
+
+    Segment s = head;
+    if (s == null) return -1;
+    int toCopy = Math.min(len, s.limit - s.pos);
```
JakeWharton: This isn't right. We need a way to limit byte decoding to only `len` chars.

okio/src/main/java/okio/Buffer.java
```diff
@@ -86,6 +90,45 @@ public long size() {
...
+      @Override public void close() throws IOException {
+      }
+
+      @Override public String toString() {
+        return Buffer.this + ".writer(" + charset.name() + ")";
+      }
+    };
+  }
+
+  void writeChars(Charset charset, char[] cbuf, int off, int len) {
```
swankjesse: One thing that’s pretty awkward is what we do if the last char in `cbuf` is half of a surrogate pair. Do we emit a replacement char? Or does `charset.encode()` do that? It’s weird and broken regardless

swankjesse: (this could happen if say, you’re chunking chars without paying attention to their boundaries)

okio/src/main/java/okio/Buffer.java
```diff
@@ -86,6 +90,45 @@ public long size() {
...
   }
 
+  @Override public Writer writer() {
+    return writer(Util.UTF_8);
+  }
+
+  @Override public Writer writer(final Charset charset) {
+    if (charset == null) throw new NullPointerException("charset == null");
+    return new Writer() {
+      @Override public void write(char[] cbuf, int off, int len) throws IOException {
```
swankjesse: wanna rewrite these parameters using our language? `char[] data`, `int offset`, `int charCount` ?

okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +173,54 @@ public long size() {
...
+
+  @Override public Reader reader(final Charset charset) {
+    if (charset == null) throw new NullPointerException("charset == null");
+    return new Reader() {
+      @Override public int read(char[] cbuf, int off, int len) throws IOException {
+        return readChars(charset, cbuf, off, len);
+      }
+
+      @Override public long skip(long n) throws IOException {
+        Buffer.this.skip(n);
```
swankjesse: Nope! This skips `n` bytes, you want to skip `n` UTF-16 chars.

okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +173,54 @@ public long size() {
...
+  int readChars(Charset charset, char[] cbuf, int off, int len) {
+    Util.checkOffsetAndCount(cbuf.length, off, len);
+
+    Segment s = head;
+    if (s == null) return -1;
+    int toCopy = Math.min(len, s.limit - s.pos);
+
+    ByteBuffer in = ByteBuffer.wrap(s.data, s.pos, toCopy);
+    CharBuffer out = charset.decode(in);
+    int copiedChars = Math.min(len, out.length());
```
swankjesse: This might be zero. You need a loop of sadness.

swankjesse: (Imagine a segment with one trailing byte of a four-byte UTF-8 char)

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -428,6 +429,36 @@ public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int b
...
+
+  @Override public Reader reader(final Charset charset) {
+    if (charset == null) throw new NullPointerException("charset == null");
+    return new Reader() {
+      @Override public int read(char[] cbuf, int off, int len) throws IOException {
+        if (buffer.size == 0) {
+          long read = source.read(buffer, Segment.SIZE);
+          if (read == -1) return -1;
+        }
+        return buffer.readChars(charset, cbuf, off, len);
```
swankjesse: Might be 0, which you definitely don’t want

okio/src/main/java/okio/Buffer.java
JakeWharton: Yeah that wasn't supposed to be in this PR

okio/src/main/java/okio/Buffer.java
pawatzaza: ok

okio/src/main/java/okio/RealBufferedSource.java
pawatzaza: ok

JakeWharton:
I think `InputStreamReader` and `OutputStreamWriter` solve this well enough. I'm not going to pursue.

PR #240, opened 2016-07-08
Add blackhole Sink.

### LINKED ISSUE #229: Blackhole
We have a NULL_SINK in Okio that accepts bytes but doesn’t do anything with ’em. Seems like something that would be generally useful. Maybe `Okio.blackhole()` ?

cketti:
This adds `Okio.blackhole()` which returns a Sink that writes to the void.

Fixes #229


swankjesse:
Thanks!

Issue #239, opened 2016-07-08
Expose Reader/Writer convenience factories.


JakeWharton:
```java
Reader reader(); // Defaults to UTF-8
Reader reader(Charset charset);
```
```java
Writer writer(); // Defaults to UTF-8
Writer writer(Charset charset);
```
The reader should implement BOM handling and do the right thing in that case.

PR #241, opened 2016-07-10
Add Reader convenience factory.

cketti:
The `Reader` supports the Unicode byte order mark (BOM) and based on it will select the right `Charset` for decoding.

See #239.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -18,7 +18,9 @@ package okio;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
+import java.io.Reader;
 import java.nio.charset.Charset;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
@@ -26,6 +28,13 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
+import static okio.ByteOrderMark.BOM_UTF_16_BE;
+import static okio.ByteOrderMark.BOM_UTF_16_LE;
+import static okio.ByteOrderMark.BOM_UTF_32_BE;
+import static okio.ByteOrderMark.BOM_UTF_32_LE;
+import static okio.ByteOrderMark.BOM_UTF_8;
+import static okio.ByteOrderMark.getByteOrderMark;
+import static okio.Util.UTF_8;
 import static okio.Util.checkOffsetAndCount;
 import static okio.Util.reverseBytesLong;
 
@@ -128,6 +137,26 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
         return Buffer.this + ".inputStream()";
       }
     };
+  }
+
+  @Override public Reader reader() {
+    return new BomAwareReader(
+        null, BOM_UTF_8, BOM_UTF_32_LE, BOM_UTF_32_BE, BOM_UTF_16_LE, BOM_UTF_16_BE);
+  }
+
+  @Override public Reader reader(Charset charset) {
+    if (charset == null) throw new IllegalArgumentException("charset == null");
+
+    String charsetName = charset.name();
+    if (charsetName.equals("UTF-8")) {
+      return new BomAwareReader(charset, BOM_UTF_8);
+    } else if (charsetName.startsWith("UTF-16")) {
+      return new BomAwareReader(charset, BOM_UTF_16_BE, BOM_UTF_16_LE);
+    } else if (charsetName.startsWith("UTF-32")) {
+      return new BomAwareReader(charset, BOM_UTF_32_BE, BOM_UTF_32_LE);
+    } else {
+      return new InputStreamReader(inputStream(), charset);
+    }
   }
 
   /** Copy the contents of this to {@code out}. */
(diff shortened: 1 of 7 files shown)


okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -442,4 +471,54 @@ public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int b
   @Override public String toString() {
     return "buffer(" + source + ")";
   }
+
+  final class BomAwareReader extends Reader {
```
swankjesse: Can this be defined only once?

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -499,4 +500,10 @@ boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int byteCoun
 
   /** Returns an input stream that reads from this source. */
   InputStream inputStream();
+
+  /** Returns a Reader that reads from this source. */
```
swankjesse: I don’t think I like this overload. When would a user not have a charset? As far as I’m concerned not having a charset is a big bug in how you think about your data.

okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +139,26 @@ public long size() {
...
 
+  @Override public Reader reader() {
+    return new BomAwareReader(
+        null, BOM_UTF_8, BOM_UTF_32_LE, BOM_UTF_32_BE, BOM_UTF_16_LE, BOM_UTF_16_BE);
+  }
+
+  @Override public Reader reader(Charset charset) {
+    if (charset == null) throw new IllegalArgumentException("charset == null");
+
+    String charsetName = charset.name();
```
swankjesse: What do you think about making BomAwareReader responsible for knowing which BOMs it can ignore for which charsets?

okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +139,26 @@ public long size() {
...
+    if (charsetName.equals("UTF-8")) {
+      return new BomAwareReader(charset, BOM_UTF_8);
+    } else if (charsetName.startsWith("UTF-16")) {
+      return new BomAwareReader(charset, BOM_UTF_16_BE, BOM_UTF_16_LE);
```
swankjesse: Will InputStreamReader do UTF-16 LE / UTF-16 BE selection for us? (I have no idea)

okio/src/main/java/okio/Buffer.java
```diff
@@ -130,6 +139,26 @@ public long size() {
...
+    } else if (charsetName.startsWith("UTF-32")) {
+      return new BomAwareReader(charset, BOM_UTF_32_BE, BOM_UTF_32_LE);
+    } else {
+      return new InputStreamReader(inputStream(), charset);
```
swankjesse: If I present UTF-32-LE and there’s a BOM, what happens?

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -442,4 +471,54 @@ public boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int b
...
+      this.suppliedCharset = charset;
+      this.byteOrderMarks = byteOrderMarks;
+    }
+
+    @Override public int read(char[] data, int offset, int charCount) throws IOException {
+      if (delegate == null) {
+        if (closed) throw new IOException("closed");
+
+        // Make sure buffer holds all bytes of the BOM.
+        if (buffer.size() < ByteOrderMark.MAX_BOM_LENGTH) {
```
swankjesse: Sadly this is not at all safe. Imagine a socket that returns `AB` and then stalls waiting for more data to be generated. In this code the user doesn’t see `A` or `B` until at least 4 characters show up. That’s a big bug.

(This happens for real if the document is JSON and the content is `{}`.)

swankjesse: Instead you need to peek a single byte, and make a decision about whether to read more characters or not.

okio/src/test/java/okio/RealBufferedSourceTest.java
```diff
@@ -215,4 +232,181 @@
...
+    BufferedSource bufferedSource = new RealBufferedSource(source);
+
+    bufferedSource.readByteString(Segment.SIZE - 1);
+    Reader reader = bufferedSource.reader();
+    String stringFromReader = readerToString(reader);
+
+    assertEquals("utf8", stringFromReader);
+  }
+
+  @Test public void readerUtf8() throws Exception {
```
swankjesse: You can avoid having to write each of these 2x by using BufferedSourceTest which runs each test on a variety of implementations.

okio/src/main/java/okio/Buffer.java
cketti: Sounds good.

okio/src/main/java/okio/Buffer.java
cketti: Hah. I just assumed InputStreamReader has no BOM handling at all. But it looks like at least on JRE 1.8 it works just fine for UTF-16/UTF-32 (BOM is used to detect endianness and is not part of the output). The only issue is UTF-8 + BOM: http://bugs.java.com/view_bug.do?bug_id=4508058

This might make things considerably easier. I'll investigate further.

swankjesse: Cool. We mostly care about Android, which might be special!

cketti: Same behavior on API 9, API 15, and API 23 emulators :tada: 

Closing this pull request. I'll create another one.

okio/src/main/java/okio/BufferedSource.java
cketti: Don't like as in:
a) let's get rid of it
b) rename it to readerUtf8()
c) something else entirely?

swankjesse: I like `readerUtf8()`!

PR #242, opened 2016-07-11
Add Reader convenience factory (second attempt)

cketti:
Because `InputStreamReader` doesn't do it, `BomAwareReader` will strip a UTF-8 BOM if present (see http://bugs.java.com/view_bug.do?bug_id=4508058).

See #239.

### DIFF
--- a/okio/src/main/java/okio/BomAwareReader.java
+++ b/okio/src/main/java/okio/BomAwareReader.java
@@ -1,0 +1,66 @@
+package okio;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.Reader;
+import java.nio.charset.Charset;
+
+import static okio.Util.UTF_8;
+
+final class BomAwareReader extends Reader {
+  private static final char BOM_CHARACTER = '\uFEFF';
+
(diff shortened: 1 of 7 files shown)


swankjesse:
Sorry for sitting on this. Just an update for @cketti and anyone else following along, I’m still quite reluctant to do stuff with the Reader API...

Unlike `inputStream()`, the `reader()` carries an invisible hidden buffer that consumes more bytes than it should. You can see that with a test like this, which fails with an `EOFException`:

```
  @Test public void mixReading() throws Exception {
    Buffer buffer = new Buffer();
    buffer.writeUtf8("xyz");
    Reader reader = buffer.readerUtf8();
    assertEquals('x', reader.read());
    assertEquals('y', buffer.readUtf8CodePoint());
    assertEquals('z', buffer.readUtf8CodePoint());
  }
```

This is different from our `inputStream()` API, which only consumes bytes you ask it for. The ultimate cause is that `InputStreamReader` buffers bytes implicitly.

What to do instead? We could do own UTF-8-only `Reader` that uses our own UTF-8 codec instead of the one in `InputStreamReader`. For now I’m happy just not supporting `Readers` too deeply in Okio.

PR #243, opened 2016-07-22
Allowing graceful shutdown (if JVM lives longer than application)

QwertGold:
Allow for gracefull shutdown of the Watchdog thread. This is important if the code runs inside an application server JBoss, Tomcat etc, where you deploy new versions of the application without restarting the application server (JVM).

If you don't shutdown the thread gracefully you will get an error from Tomcat.
WARNING: The web application [ROOT##000860] appears to have started a thread named [Okio Watchdog] but has failed to stop it. This is very likely to create a memory leak. Stack trace of thread: java.lang.Object.wait(Native Method) java.lang.Object.wait(Object.java:502) okio.AsyncTimeout.awaitTimeout(AsyncTimeout.java:297) okio.AsyncTimeout.access$000(AsyncTimeout.java:40) okio.AsyncTimeout$Watchdog.run(AsyncTimeout.java:272)

Every time we redeploy our application (potentially many times every day) the JVM looses 1MB of stack space associated with the thread, until the JVM finally crashes with an OutOfMemmoryError or infinite GC loop.

### DIFF
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -71,7 +71,7 @@ public class AsyncTimeout extends Timeout {
   private static volatile boolean shutdown = false;
   private static Watchdog watchdog;
 
-  public final void enter() {
+  public synchronized final void enter() {
     if (inQueue) throw new IllegalStateException("Unbalanced enter/exit");
     long timeoutNanos = timeoutNanos();
     boolean hasDeadline = hasDeadline();
@@ -85,7 +85,7 @@ public class AsyncTimeout extends Timeout {
   private static synchronized void scheduleTimeout(
       AsyncTimeout node, long timeoutNanos, boolean hasDeadline) {
     // Start the watchdog thread and create the head node when the first timeout is scheduled.
-    if (head == null) {
+    if (watchdog == null) {
       head = new AsyncTimeout();
       watchdog = new Watchdog();
       watchdog.start();
@@ -140,11 +140,13 @@ public class AsyncTimeout extends Timeout {
     return true;
   }
 
-  public static void shutdown() {
-      shutdown = true;
-      if (watchdog != null) {
-        watchdog.interrupt();
-      }
+  protected synchronized static void shutdown() {
+    shutdown = true;
+    if (watchdog != null) {
+      watchdog.interrupt();
+    }
+    watchdog = null;
+
   }
 
   /**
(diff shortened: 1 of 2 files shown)


QwertGold:
I read the comment on the previously rejected PR (https://github.com/square/okio/pull/114) 
And however little you may like this solution it is the only one that will work, unless you rewrite your library completely.

The problem is that you have built AsyncTimeout as a static singleton without any lifecycle methods, which is a big no-no. Anything that starts a thread should have a stop method which stops the thread. If you had not caught InterruptedException inside while(true), I could have worked around the issue, by finding the thread and interrupting it. 
Currently this problem is causing memory leaks in all web application that uses Okio (and libraries like Segment which depends on Okio).
Application server offer events when the an application is shutdown, but because AsyncTimeout is static, only a static shutdown method will work. 
The correct implementation (which would break backwards compatibility) would be not have no static singletons, so users had to create a new instance of Okio ( which has a non-static AsyncTimeout), and Okio should have a shutdown method that forwards to AsyncTimeout. 
However since this would break backwards compatibility other libraries may not be happy since they would have to change their code, which means some will simply not upgrade. 

Regards,

swankjesse:
The fix I’d like to implement is to have the watchdog exit after say, 60 seconds of inactivity. This change is awkward because the application developer has to know it exists.

QwertGold:
I've looked a little at the code a better solution would be to place the shutdown method on Okio, since this is where the factory methods developers use to create Sources, Sinks etc. are located. This is how Amazone does it in it's Java API which also starts background threads (http://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/http/IdleConnectionReaper.html), and inside a web container you explicitly have to call this method.

QwertGold:
There is also a few  (theoretical) race conditions

AsyncTimeout.enter() should be synchronized, because it reads then writes the inQueue field, concurrent call to this method could end up adding the Timout to the 'LinkedList' of timeouts twice. 

Fields hasDeadline, deadlineNanoTime, timeoutNanos are not volatile. So if you creates a (Socket) Sink using one thread (this will create an AsyncTimeout), then hand the sink to another thread which calls AsyncTimeout.timeout(), then the first thread is no longer guaranteed to read the correct values of the 3 fields, since they were set by a different thread. This use case is far out, but theoretically possible.

swankjesse:
Sources, sinks and timeouts are not threadsafe. So it is an error for an application to use the same timeout from multiple threads without its own concurrency controls.

swankjesse:
Obsoleted by this:
https://github.com/square/okio/pull/256

PR #244, opened 2016-07-27
Configure compress level for gzip sink

### LINKED ISSUE #224: Configure compression level on gzip
Right now we can’t. That is lame.

mblcdr:
Resolves #224

### DIFF
--- a/okio/src/main/java/okio/GzipSink.java
+++ b/okio/src/main/java/okio/GzipSink.java
@@ -110,6 +110,14 @@ public final class GzipSink implements Sink {
     if (thrown != null) Util.sneakyRethrow(thrown);
   }
 
+  /**
+   * Returns the {@link Deflater}.
+   * Use it to access stats, dictionary, compression level, etc.
+   */
+  public Deflater deflater() {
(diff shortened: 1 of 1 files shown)


swankjesse:
I was thinking maybe we’d just expose the `Deflater` with an accessor method (`deflater()`) and that’d allow callers to also access stats (`getTotalIn()`, `getTotalOut()`), the dictionary, etc.

Thoughts?

mblcdr:
Sounds like a better idea. Will update the PR accordingly.

okio/src/main/java/okio/GzipSink.java
```diff
@@ -110,6 +110,14 @@ public GzipSink(Sink sink) {
     if (thrown != null) Util.sneakyRethrow(thrown);
   }
 
+  /**
+   * Returns the {@link Deflater}.
+   * Use it to access stats, dictionary, compression level, etc.
+   */
+  public Deflater deflater(){
```
15characterlimi: missing space before '{'?

swankjesse:
Thanks!

PR #247, opened 2016-08-06
Don't overallocate when base64-encoding.

cketti:
The order of operations matters. Do the division first to ensure we always arrive at the correct multiple of 4.


swankjesse:
Nice.

Issue #250, opened 2016-08-12
GzipSource does not support reading a source larger than 2GB 


joshzana:
We use OKHttp to download files, and recently hit a failure downloading large files, which we traced back to this exception:
```
java.io.IOException: ISIZE: actual 0x7fffffff != expected 0xc0000000
0 = {StackTraceElement@10289} "okio.GzipSource.checkEqual(GzipSource.java:205)"
1 = {StackTraceElement@10290} "okio.GzipSource.consumeTrailer(GzipSource.java:174)"
2 = {StackTraceElement@10291} "okio.GzipSource.read(GzipSource.java:92)"
3 = {StackTraceElement@10292} "okio.RealBufferedSource$1.read(RealBufferedSource.java:371)"
4 = {StackTraceElement@10293} "java.io.FilterInputStream.read(FilterInputStream.java:118)"
5 = {StackTraceElement@10294} "java.io.InputStream.read(InputStream.java:162)"
```

The line in question is:
```
    checkEqual("ISIZE", source.readIntLe(), inflater.getTotalOut());
```

`getTotalOut()` has the following [javadoc](
https://docs.oracle.com/javase/7/docs/api/java/util/zip/Inflater.html#getTotalOut()):
> Since the number of bytes may be greater than Integer.MAX_VALUE, the getBytesWritten() method is now the preferred means of obtaining this information.

The [Gzip RFC](https://www.ietf.org/rfc/rfc1952.txt) has this to say:

>         ISIZE (Input SIZE)
            This contains the size of the original (uncompressed) input
            data modulo 2^32.

`GzipSink` likely has the same issue, as it is using `getTotalIn()` but I don't have a repro of that.

swankjesse:
Any idea which gzip encoder you’re using on the other end? I attempted to reproduce this with `GzipOutputStream` on Java 7 and it encodes the size mod 2^31 in its `ISIZE`. Since both the `Inflater.getTotalOut()` value _and_ the gzip’s `ISIZE` are truncated, it doesn’t fail.

swankjesse:
[RFC 1952](https://tools.ietf.org/html/rfc1952#section-2.2) says that `ISIZE` is modulo 2^32. It looks like when there’s an overflow `Inflater.getTotalOut()` returns `Integer.MAX_VALUE` in some environments and wraps on others.

```
         ISIZE (Input SIZE)
            This contains the size of the original (uncompressed) input
            data modulo 2^32.
```

[...]

PR #252, opened 2016-08-14
Fix GzipSource to handle large files on AOSP.

swankjesse:
Closes: https://github.com/square/okio/issues/250


swankjesse:
The tests were already passing on the JVM. But I ran ’em against libcore and they failed there. This gets ’em working everywhere.

JakeWharton:
Lgtm.

JakeWharton:
What about `GzipSink`?

swankjesse:
It was never broken!

swankjesse:
... though I was thinking about this, and it’s worth changing for symmetry.
https://github.com/square/okio/pull/253

JakeWharton:
Agreed

PR #256, opened 2016-08-22
Exit the watchdog thread after being idle for 60 seconds.

swankjesse:
This should make it possible for people doing class unloading to
unload class loaders containing Okio.

Closes https://github.com/square/okio/issues/107

### DIFF
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -17,45 +17,45 @@ package okio;
 
 import java.io.IOException;
 import java.io.InterruptedIOException;
+import java.util.concurrent.TimeUnit;
 
 import static okio.Util.checkOffsetAndCount;
 
 /**
- * This timeout uses a background thread to take action exactly when the timeout
- * occurs. Use this to implement timeouts where they aren't supported natively,
- * such as to sockets that are blocked on writing.
+ * This timeout uses a background thread to take action exactly when the timeout occurs. Use this to
+ * implement timeouts where they aren't supported natively, such as to sockets that are blocked on
+ * writing.
  *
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/AsyncTimeout.java
```diff
@@ -17,45 +17,45 @@
...
-   * (slow) progress. Without this, writing a single 1 MiB buffer may never
-   * succeed on a sufficiently slow connection.
+   * Don't write more than 64 KiB of data at a time, give or take a segment. Otherwise slow
+   * connections may suffer timeouts even when they're making (slow) progress. Without this, writing
+   * a single 1 MiB buffer may never succeed on a sufficiently slow connection.
    */
   private static final int TIMEOUT_WRITE_SIZE = 64 * 1024;
 
+  /** Duration for the watchdog thread to be idle before it shuts itself down. */
+  private static final long IDLE_TIMEOUT_MILLIS = TimeUnit.SECONDS.toMillis(60);
```
swankjesse: The PR has no new test cases, which is sad. I ran this with various small values and confirmed that it comes back to life okay. Not sure what else to do that isn’t terrible.

JakeWharton:
lgtm

damien-urruty-sonarsource:
Unfortunately this is not sufficient in our use case. We use okhttp (and so okio) in a plugin running in an IDE. When the plugin gets uninstalled, the IDE tries to unload the corresponding class loader. We can see the Watchdog instance is still alive, and it prevents the plugin from unloading correctly. We are allowed a rather short window (5s) to successfully unload everything, so we can't wait 1min here...

Could the recent changes introduced in okhttp about TaskRunner benefit okio ?

Anyway I think I would prefer having an API to explicitly shutdown the watchdog. I understand in 99% of the time it shouldn't be used but there will still be some valid use cases.

Issue #248, opened 2016-08-07
ByteString decode/encode custom charset


djodjoni:
ByteString currently can be created by encoding string using only UTF-8 charset, also decoding underlying byte array can be done using again only UTF-8 charset, which means in order to use another charset for decoding one needs to allocate another clone of byte[] and then decode this array.

Are you willing to extend this functionality or stick to utf-8 only?

swankjesse:
You can do it manually with String. We really like UTF-8 and look forward to a time when that's the only charset you need.

djodjoni:
I am looking forward this bright future as well :)
It is of course possible to use String but this will add more wrapper code and create otherwise unnecessary memory allocations to clients.

So you prefer to keep it simple and not to add something like "getString(java.nio.charset.Charset charset)" and extend its "ergonomic-ability" ?

swankjesse:
Yeah, we could use an API for that. encodeString/decodeString ?

djodjoni:
yea, looks good to me

PR #257, opened 2016-08-28
APIs on ByteString to encode and decode any charset.

swankjesse:
Closes: https://github.com/square/okio/issues/248


JakeWharton:
Lgtm

Issue #258, opened 2016-08-29
I find a problem in the 961 line of  the  Buffer.java ？


wo883721:
At the 961 line of  the  Buffer.java, We call writeUtf8(string) method, I think here should call writeUtf8(string,beginIndex,endIndex)。

swankjesse:
Wow, huge bug. Will fix ASAP.

PR #259, opened 2016-08-29
Fix a major bug where writeString() didn't always respect offsets.

swankjesse:
Closes: https://github.com/square/okio/issues/258


raskasa:
Wanted to quickly double-check, as I'm relying on the GZIP APIs from this library:

- Does this bug affect 1.10 only?  Or does it affect multiple versions?
- Are you looking to cut a release (1.10.1) in the next day or two?

cketti:
That bug has been present since 2015-04-12.

JakeWharton:
The method is unused internally, so only direct external callers would be
affected.

PR #261, opened 2016-09-14
ByteString: support creating from NIO ByteBuffers

sl4mmy:
Add support for creating ByteStrings from NIO ByteBuffers.  Useful for
projects which standardized on okio for their own code but still need to
integrate with other libraries based on NIO.

Signed-off-by: Kent R. Spillner <kspillner@acm.org>


swankjesse:
Awesome.

swankjesse:
CLA looks good too. Thanks!

Issue #254, opened 2016-08-16
Add MacSink and MacSource


tmarsteel:
I currently need a `MacSink` and a `MacSource`. Are these currently planned? If not, i'd create a pull request within the next days containing the code i am about to write...

swankjesse:
Nothing in flight but it makes sense. Could we reuse HashingSink and HashingSource?

swankjesse:
I'm specifically interested in keeping our APIs small. Maybe we make HashingSource abstract, and have two non-public implementations, one that uses MessageDigest, and the other that uses Mac.

tmarsteel:
What about this?

```java
public class AccumulatorCalculationSink<T> extends ForwardingSink {
	
	private Accumulator<? extends T> accumulator;
	
	public AccumulatorCalculationSink(Sink forwardSink, Accumulator<? extends T> accumulator) {
		super(forwardSink);
		
		this.accumulator = accumulator;
	}
	
	@Overwrite
	public void write(Buffer buffer, int nBytes) {
		byte[] data = buffer.readByteArray(nBytes);
		accumulator.add(data, 0, nBytes);
	}
	
	public T result() {
		return accumulator.result();
	}
	
	public interface Accumulator<T> {
		public static void add(byte[] data, int offset, int len);
		public static T result();
	}
}

// vice-versa for Source
```

And then have a HashingAccumulator, MacAccumulator to do the work; maybe add some factory methods to Okio?

swankjesse:
I don't think we need hardly any new public APIs. Maybe just new methods like this?

```
public class HashingSource {
  public static Source sha1Mac(Key key)
  public static Source sha256Mac(Key key)
  public static Source md5Mac(Key key)
  ...
}
```

tmarsteel:
And these would return package-private MacSource objects?

swankjesse:
Exactly!

PR #266, opened 2016-10-07
Implement HMAC SHA-1 and HMAC-SHA-256 throughout Okio.

swankjesse:
This adds APIs to ByteString, Buffer, HashingSource and HashingSink.

I've forgone polymporphism in HashingSource and HashingSink because it
adds more code than it saves. I can be convinced here.

Closes: https://github.com/square/okio/issues/254

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -20,11 +20,14 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.nio.charset.Charset;
+import java.security.InvalidKeyException;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
 
 import static okio.Util.checkOffsetAndCount;
 import static okio.Util.reverseBytesLong;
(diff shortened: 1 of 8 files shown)


adriancole:
Thanks tons! I'd use this right away in Zipkin for its amazon support
https://github.com/openzipkin/zipkin/blob/master/zipkin-autoconfigure/storage-elasticsearch-aws/src/main/java/zipkin/autoconfigure/storage/elasticsearch/aws/AWSSignatureVersion4.java#L175

okio/src/main/java/okio/HashingSource.java
```diff
@@ -79,7 +107,11 @@ private HashingSource(Source source, String algorithm) {
       // Hash that segment and all the rest until the end.
       while (offset < sink.size) {
         int pos = (int) (s.pos + start - offset);
-        messageDigest.update(s.data, pos, s.limit - pos);
+        if (messageDigest != null) {
+          messageDigest.update(s.data, pos, s.limit - pos);
+        } else if (mac != null) {
```
JakeWharton: Any reason you went for `else if` vs. just `else`?

okio/src/main/java/okio/Buffer.java
```diff
@@ -1538,16 +1541,46 @@ public ByteString sha256() {
...
+  public ByteString hmacSha1(ByteString key) {
+    return hmac("HmacSHA1", key);
+  }
+
+  /** Returns the 256-bit SHA-256 HMAC of this buffer. */
+  public ByteString hmacSha256(ByteString key) {
+    return hmac("HmacSHA256", key);
+  }
+
+  private ByteString hmac(String algorithm, ByteString key) {
```
cketti: Missing `null` check for `key`.

okio/src/main/java/okio/ByteString.java
```diff
@@ -150,6 +153,28 @@ private ByteString digest(String algorithm) {
...
+  public ByteString hmacSha1(ByteString key) {
+    return hmac("HmacSHA1", key);
+  }
+
+  /** Returns the 256-bit SHA-256 HMAC of this byte string. */
+  public ByteString hmacSha256(ByteString key) {
+    return hmac("HmacSHA256", key);
+  }
+
+  private ByteString hmac(String algorithm, ByteString key) {
```
cketti: Missing `null` check for `key`.

okio/src/main/java/okio/HashingSink.java
```diff
@@ -55,12 +59,36 @@ public static HashingSink sha256(Sink sink) {
...
     super(sink);
     try {
       this.messageDigest = MessageDigest.getInstance(algorithm);
+      this.mac = null;
+    } catch (NoSuchAlgorithmException e) {
+      throw new AssertionError();
+    }
+  }
+
+  private HashingSink(Sink sink, ByteString key, String algorithm) {
```
cketti: Missing `null` check for `key`.

okio/src/main/java/okio/HashingSource.java
```diff
@@ -53,12 +57,36 @@ public static HashingSource sha256(Source source) {
...
     super(source);
     try {
       this.messageDigest = MessageDigest.getInstance(algorithm);
+      this.mac = null;
+    } catch (NoSuchAlgorithmException e) {
+      throw new AssertionError();
+    }
+  }
+
+  private HashingSource(Source source, ByteString key, String algorithm) {
```
cketti: Missing `null` check for `key`.

okio/src/main/java/okio/Buffer.java
cketti: Right. I wasn't sure whether explicit checks were omitted intentionally. It's not without precedent in the codebase. But I had the impression explicit checks are the norm.
Is there a code style document where preferences for things like this are spelled out?

swankjesse: We don’t have a document. Closest is the Google Java Style Guide and Effective Java.

okio/src/main/java/okio/HashingSource.java
swankjesse: None. Fixed.

PR #267, opened 2016-10-17
Remove creation of synthetic accessor method.

vanniktech:
Saves 1 method. (head is accessed int the private static inner class `Watchdog`)


swankjesse:
Thanks!

PR #274, opened 2016-12-07
endsWith -> suffix

larsgrefer:
Rename the parameter of `ByteString.endsWith()` from prefix to suffix


swankjesse:
Yes, that’s way better!

PR #275, opened 2016-12-08
Add pro guard rules to readme

Jawnnypoo:
This was discussed a bit here https://github.com/square/retrofit/pull/2090

This adds the ProGuard rules for okio. The format is taken from Picasso. I also added the license footer to the README, as it seems most other Square libraries have it


swankjesse:
Thanks!

Issue #153, opened 2015-06-06
Optional limits on methods like readLine()


swankjesse:
To avoid exhausting memory when it would be better to just crash.

Inspired by https://github.com/square/okio/issues/150

eneim:
I'm waiting for this ... Just for an extra: any idea of a delayed readLine()? (so every 100ms I read the next line, without the Thread.sleep() methodology). Thanks and hope this one is added to life soon :D.

swankjesse:
The limits I was imagining would cause the reads to fail with an exception if they were too long. This might not be a good fit.

ericedens:
For `readUtf8LineStrict`, I just submitted a #279 to recast _limit_ as the maximum number of bytes returned, rather than the maximum number of bytes scanned.

I'll start on the remaining methods soon! (`readUtf8Line()`, `indexOf(ByteString)`, and `indexOfElement()`)

ericedens:
OK, I'm looking at `readUtf8Line(long limit)` now. Any preference on the behavior if `limit` is 0? Looking at its analogs:
- `readUtf8LineStrict(0)` throws an `EOFException`
- `readUtf8(0)` returns an empty string
- Calling `readUtf8Line()` on an empty source returns null

Returning an empty string follows the spirit of `readUtf8Line` being a forgiving API, and we're indeed returning a string that has zero bytes. So I'm inclined in that direction.

Thoughts?

swankjesse:
I should apologize I also changed `readUtf8LineStrict()` to permit 0 and return an empty string. I should have mentioned that on that PR.

My original plan to forbid it wasn’t counting on the fact that sometimes callers may wanna just consume a `\n` or `\r\n` from the input.

So I think the same makes sense for non-strict. Really the only difference between the two is that the strict one won’t accept the end of the file as a line terminator.

ericedens:
Good to close this?

swankjesse:
Yes! Thanks @ericedens!

PR #278, opened 2017-01-09
Allow length of readUtf8LineStrict to be limited.

ericedens:
I ran into needing this while using okio to write an HTTP server, and I think it should address issue #153 .

Two API additions:

1. BufferedOutputStream.readUtf8LineStrict accepts a maxByteCount parameter. If the line break isn't found by that limit, an EOFException is thrown.  (Is that the right exception??)
2. BufferedOutputStream.indexOf includes a marker for ending the search. If reached, -1 is returned.

I thought about whether readUtf8Line should get the maxByteCount parameter as well. If added, it would have a contract of questionable utility: return a string at end of line, EOF, or when this count parameter is reached. In any case, if the method is desirable, it should be easy to add.

Additionally, there are indexOf methods that search for ByteStrings. Those aren't covered by this PR either, but should be similarly easy to add if wanted.

Test output: [mvn-clean-verify.txt](https://github.com/square/okio/files/692663/mvn-clean-verify.txt)
Related issue: #153 Optional limits on methods like readLine()

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -642,11 +642,16 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
   }
 
   @Override public String readUtf8LineStrict() throws EOFException {
-    long newline = indexOf((byte) '\n');
+    return readUtf8LineStrict(Long.MAX_VALUE);
+  }
+
+  @Override public String readUtf8LineStrict(long maxRead) throws EOFException {
+    if (maxRead < 1) throw new IllegalArgumentException("maxRead < 1");
+    long newline = indexOf((byte) '\n', 0, maxRead);
     if (newline == -1) {
       Buffer data = new Buffer();
       copyTo(data, 0, Math.min(32, size));
-      throw new EOFException("\\n not found: size=" + size()
+      throw new EOFException("\\n not found: scanLength=" + Math.min(size(), maxRead)
           + " content=" + data.readByteString().hex() + "…");
     }
     return readUtf8Line(newline);
@@ -1263,7 +1268,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
   }
 
   @Override public long indexOf(byte b) {
-    return indexOf(b, 0);
+    return indexOf(b, 0, Long.MAX_VALUE);
   }
 
   /**
@@ -1271,7 +1276,17 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
    * -1 if this buffer does not contain {@code b} in that range.
    */
   @Override public long indexOf(byte b, long fromIndex) {
-    if (fromIndex < 0) throw new IllegalArgumentException("fromIndex < 0");
+    return indexOf(b, fromIndex, Long.MAX_VALUE);
+  }
+
+  @Override public long indexOf(byte b, long fromIndex, long toIndex) {
+    if (fromIndex < 0 || toIndex < fromIndex) {
+      throw new IllegalArgumentException(
+          String.format("size=%s fromIndex=%s toIndex=%s", size, fromIndex, toIndex));
+    }
+
+    if (toIndex > size) toIndex = size;
+    if (fromIndex == toIndex) return -1L;
 
     Segment s;
     long offset;
@@ -1301,9 +1316,11 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     }
 
     // Scan through the segments, searching for b.
-    while (offset < size) {
+    while (offset < toIndex) {
       byte[] data = s.data;
-      for (int pos = (int) (s.pos + fromIndex - offset), limit = s.limit; pos < limit; pos++) {
+      int limit = (int) Math.min(s.limit, s.pos + toIndex - offset);
+      int pos = (int) (s.pos + fromIndex - offset);
+      for (; pos < limit; pos++) {
         if (data[pos] == b) {
           return pos - s.pos + offset;
         }
--- a/okio/src/main/java/okio/BufferedSource.java
+++ b/okio/src/main/java/okio/BufferedSource.java
@@ -392,6 +392,30 @@ public interface BufferedSource extends Source {
   String readUtf8LineStrict() throws IOException;
 
   /**
+   * Like {@link #readUtf8LineStrict()}, but allows the caller to specify the maximum number
+   * of bytes that should be read. If {@code maxRead} is reached without finding a line break,
+   * then an {@link java.io.EOFException} is thrown. A common use case is parsing headers,
+   * to protect against a sender that never sends {@code "\n"} or {@code "\r\n"}.
+   *
+   * <p>This method is safe. No bytes are discarded if the match fails, and the caller is free
+   * to try another match:<pre>{@code
+   *
+   *   Buffer buffer = new Buffer();
+   *   buffer.writeUtf8("12345\r\n");
+   *
+   *   // This will throw! A newline character (\n)
+   *   // must be read within maxRead.
+   *   buffer.readUtf8LineStrict(5);
+   *
+   *   // No bytes have been consumed, so the caller
+   *   // can recover from the exception and
+   *   // try again.
+   *   assertEquals("12345", buffer.readUtf8LineStrict(100));
+   * }</pre>
+   */
+  String readUtf8LineStrict(long maxRead) throws IOException;
+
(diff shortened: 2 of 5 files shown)


swankjesse:
Awesome. I've been dragging my feet thinking about how to expose this and you've just gone and done it. Nice. I'll give it a full review soon!

okio/src/test/java/okio/ReadUtf8LineTest.java
```diff
@@ -94,6 +94,20 @@
...
+    data.writeUtf8("ab\nde\n");
+    assertEquals("ab", source.readUtf8LineStrict(4));
+    assertEquals("de", source.readUtf8LineStrict(4));
+
+    data.writeUtf8("abcdef\n");
+    try {
+      source.readUtf8LineStrict(4);
+      fail();
+    } catch (EOFException expected) {
+      assertEquals("\\n not found: size=4 content=6162636465660a…", expected.getMessage());
```
JakeWharton: This message looks odd. Size 4 but content is more than 4 bytes.

ericedens: It does look odd, now that you mention it. One option: "\n not found in first n bytes of content=6162636465660a…", where n is replaced with the number of bytes scanned (in this case 4).

cketti: Rather than including encoded bytes in the exception message it'd be nice to have a guarantee (JavaDoc and tests) that the BufferedSource is still in a useable state so users could fetch the information themselves if necessary.

JakeWharton: Yes, that would be a good addition.

ericedens: The byte dump is an existing feature, so I'll let Jesse and Jake weigh in on whether they want to keep that.

The previous implementation still follows: the bytes are copied from the source instead of consumed. Regardless, let's definitely add tests and javadocs to protect that behavior.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1263,18 +1267,23 @@ Segment writableSegment(int minimumCapacity) {
...
   @Override public long indexOf(byte b, long fromIndex) {
+    return indexOf(b, fromIndex, Long.MAX_VALUE);
+  }
+
+  @Override public long indexOf(byte b, long fromIndex, long toIndex) {
     if (fromIndex < 0) throw new IllegalArgumentException("fromIndex < 0");
 
     Segment s;
     long offset;
+    if (toIndex > size) toIndex = size;
```
swankjesse: this should probably throw. It’s probably a bug.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1283,9 +1292,9 @@ Segment writableSegment(int minimumCapacity) {
       if (s == null) {
         // No segments to scan!
         return -1L;
-      } else if (size - fromIndex < fromIndex) {
+      } else if (toIndex - fromIndex < fromIndex) {
```
swankjesse: This should still be size. Otherwise incorrect. This is the offset of the scan, not the desired offset of the scan. It’ll be broken for buffers with multiple segments.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1283,9 +1292,9 @@ Segment writableSegment(int minimumCapacity) {
...
         // We're scanning in the back half of this buffer. Find the segment starting at the back.
-        offset = size;
+        offset = toIndex;
```
swankjesse: same

okio/src/main/java/okio/Buffer.java
```diff
@@ -1283,9 +1292,9 @@ Segment writableSegment(int minimumCapacity) {
...
         while (offset > fromIndex) {
```
swankjesse: ... and this code will need to just skip segments until `offset <= toIndex`

okio/src/main/java/okio/Buffer.java
```diff
@@ -1301,9 +1310,10 @@ Segment writableSegment(int minimumCapacity) {
     }
 
     // Scan through the segments, searching for b.
-    while (offset < size) {
+    while (offset < toIndex) {
       byte[] data = s.data;
-      for (int pos = (int) (s.pos + fromIndex - offset), limit = s.limit; pos < limit; pos++) {
+      long limit = Math.min(s.limit, s.pos + toIndex - offset);
```
swankjesse: dig

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -433,6 +440,13 @@
    */
   long indexOf(byte b, long fromIndex) throws IOException;
 
+  /**
+   * Returns the index of {@code b} if it is found in the range of {@code fromIndex} inclusive
```
swankjesse: nice docs

swankjesse:
I think this PR is great.

Next steps:

 * We need some tests that cover multiple-segment buffers to confirm that those cases work. My guess is that if you have, say, 5 segments and it starts from the end in the 3rd segment, it’s going to get the wrong answer.
 * Review other APIs that need this. (`readUtf8Line()`, `indexOf(ByteString)`, `indexOfElement()`). You’re welcome to do this work if you’d like, or I can take it. You might enjoy it!

okio/src/main/java/okio/Buffer.java
ericedens: sgtm

ericedens:
Thanks for the review! In the next few days I'll update those spots, and take a look at the other methods that could use an upper bound.

okio/src/main/java/okio/Buffer.java
ericedens: After some more thought, I don't think throwing is appropriate. The buffer may not be populated with all of the stream's future data yet, even though the caller still knows that they're willing to read up to n bytes. This happens in RealBufferedSource.indexOf, where additional data may be read from the source if the target byte isn't found yet.

okio/src/main/java/okio/Buffer.java
ericedens: yuck, good catch.

okio/src/main/java/okio/Buffer.java
ericedens: fixed

okio/src/main/java/okio/Buffer.java
ericedens: Rolled back to the original version -- we still need to find the segment that contains toIndex, because that's where the scan will start.

okio/src/main/java/okio/BufferedSource.java
ericedens: thanks!

ericedens:
OK! Here it is. I fixed the items from your comments, and added some multi-segment tests. (Travis had a series of out-of-memory failures from multi-segment ONE_BYTE_AT_A_TIME tests, but those are tuned now.)

For the other APIs, my preference would be to do those in another PR so that the change doesn't get too big. Thoughts?

Thanks for taking a look!

swankjesse:
I merged this manually. Really nice PR. Thank you.

swankjesse:
Also I made two changes. I renamed `maxRead` to `limit`. I think that’s slightly more consistent with the names in Okio. And I changed the code to accept `limit = 0` which will throw an `EOFException` unconditionally. Let me know if I’m mistaken here!

A couple follow-ups:

I want limits overloads for `readUtf8Line()`, `indexOf(ByteString)`, and `indexOfElement()`. I’ll let you do these unless you’d rather not!

We might want to reinterpret the limit argument for `readUtf8Line()`. I think that as a caller I’m probably thinking in “line lengths” and not “delimiter offsets”. For example I might assume 72 bytes per line for an xterm. Or [RFC 2822](https://tools.ietf.org/html/rfc2822#section-3.5) specifies “maximum 998 characters” which came from 1000 characters including CRLF in an earlier spec. It’s nice if I say `readUtf8LineStrict(2048)` and I know the result fits in 2048 UTF-8 bytes. Unfortunately this has one sticky special case. If we make `readUtf8LineStrict(10)` delegate to `indexOf(0, 11)` then we might get `\r` as the byte at 10. So sometimes we’ll need to peek ahead (N+1) bytes and other times (N+2) bytes. Ick.

swankjesse:
PR as merged: 
https://github.com/square/okio/commit/d1d1c7bbba6561860d0e66af0900ca2ab105d814

PR #279, opened 2017-01-17
'limit' in readUtf8LineStrict should specify maximum return length, not the maximum scan length

ericedens:
I liked your idea in PR #278 to change limit to be "the maximum number of bytes returned" instead of "the maximum number of bytes scanned."

I'm not pumped about the duplicate code that this adds in Buffer.readUtf8LineStrict and RealBufferedSource.readUtf8LineStrict, but I didn't see a clean way to avoid that. I'll keep thinking about that while I work on adding limits to the other limitable methods.

(Related to #153, Optional limits on methods like readLine())

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -647,14 +647,21 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public String readUtf8LineStrict(long limit) throws EOFException {
     if (limit < 0) throw new IllegalArgumentException("limit < 0: " + limit);
-    long newline = indexOf((byte) '\n', 0, limit);
-    if (newline == -1) {
-      Buffer data = new Buffer();
-      copyTo(data, 0, Math.min(32, size));
-      throw new EOFException("\\n not found: scanLength=" + Math.min(size(), limit)
-          + " content=" + data.readByteString().hex() + "…");
+    if (limit == 0) throw new EOFException("limit == 0");
+    long scanLength = limit == Long.MAX_VALUE ? Long.MAX_VALUE : limit + 1;
+    long newline = indexOf((byte) '\n', 0, scanLength);
+    if (newline != -1) {
+      return readUtf8Line(newline);
+    } else if (
+        size() > scanLength
+            && getByte(scanLength - 1) == '\r'
+            && getByte(scanLength) == '\n') {
+      return readUtf8Line(scanLength);
     }
-    return readUtf8Line(newline);
+    Buffer data = new Buffer();
+    copyTo(data, 0, Math.min(32, size()));
+    throw new EOFException("\\n not found: limit=" + Math.min(size(), limit)
+        + " content=" + data.readByteString().hex() + '…');
   }
 
   String readUtf8Line(long newline) throws EOFException {
--- a/okio/src/main/java/okio/BufferedSource.java
+++ b/okio/src/main/java/okio/BufferedSource.java
@@ -392,10 +392,13 @@ public interface BufferedSource extends Source {
   String readUtf8LineStrict() throws IOException;
 
   /**
-   * Like {@link #readUtf8LineStrict()}, but this allows the caller to specify a maximum number of
-   * bytes to scan. If {@code limit} bytes are scanned without finding a line break, then an {@link
-   * java.io.EOFException} is thrown. A common use case is protecting against input that doesn't
-   * include {@code "\n"} or {@code "\r\n"}.
+   * Like {@link #readUtf8LineStrict()}, except this allows the caller to specify the longest
+   * allowed match. Use this to protect against streams that may not include
+   * {@code "\n"} or {@code "\r\n"}.
+   *
+   * <p>The returned string will have at most {@code limit} UTF-8 bytes, and the maximum number
+   * of bytes scanned is {@code limit + 2}. If {@code limit == 0} this will always throw
+   * an {@code EOFException} because no bytes will be scanned.
    *
    * <p>This method is safe. No bytes are discarded if the match fails, and the caller is free
    * to try another match: <pre>{@code
@@ -403,15 +406,12 @@ public interface BufferedSource extends Source {
    *   Buffer buffer = new Buffer();
    *   buffer.writeUtf8("12345\r\n");
    *
-   *   // This will throw! A newline character (\n) must be read within the limit.
-   *   buffer.readUtf8LineStrict(5);
+   *   // This will throw! There must be \r\n or \n at the limit or before it.
+   *   buffer.readUtf8LineStrict(4);
    *
    *   // No bytes have been consumed so the caller can retry.
-   *   assertEquals("12345", buffer.readUtf8LineStrict(100));
+   *   assertEquals("12345", buffer.readUtf8LineStrict(5));
    * }</pre>
-   *
-   * <p>The returned string be up to {@code limit - 1} UTF-8 bytes. If {@code limit == 0} this will
-   * always throw an {@code EOFException} because no bytes will be scanned.
    */
   String readUtf8LineStrict(long limit) throws IOException;
 
--- a/okio/src/test/java/okio/BufferedSourceTest.java
+++ b/okio/src/test/java/okio/BufferedSourceTest.java
@@ -475,18 +475,15 @@ public final class BufferedSourceTest {
   }
 
   @Test public void indexOfByteWithBothOffsets() throws IOException {
+    if (factory == Factory.ONE_BYTE_AT_A_TIME) {
+      // When run on Travis, ONE_BYTE_AT_A_TIME
+      // causes out-of-memory errors.
+      return;
+    }
     byte a = (byte) 'a';
     byte c = (byte) 'c';
 
-    int size;
-    if (factory == Factory.ONE_BYTE_AT_A_TIME) {
-      // Larger segment sizes for ONE_BYTE_AT_A_TIME
-      // cause Travis to kill the test.
-      size =  Segment.SIZE + 10;
-    } else {
-      size = Segment.SIZE * 5;
-    }
-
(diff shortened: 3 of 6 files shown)


swankjesse:
Unfortunately there’s a wrinkle in this. In an unlikely-but-possible scenario we have enough information to fail after reading N+1 bytes but we’ll block until we can read N+2 bytes. This could cause us to block forever!

As a contrived example, here’s an imaginary protocol succeeding:

1. Client sends `REQUEST ALPHABET 0..3\n`
2. Server responds `ABC\n`
3. Client calls `readLineUtf8(3)` and succeeds.
4. Client sends `REQUEST ALPHABET 3..7\n`
5. Server responds `DEFG\n`
6. Client calls `readLineUtf8(4)` and succeeds.

Now we introduce a regression where the server’s accidentally sends `n` instead of `\n`. It happens.

1. Client sends `REQUEST ALPHABET 0..3\n`
2. Server responds `ABCn`
3. Client calls `readLineUtf8(3)` and succeeds.

The client has enough information to detect the error but will wait indefinitely for a `\n` that isn’t coming. I think the awkward fix is to assume reading N+1 bytes will be sufficient, and to only read N+2 if we run into a `\r`.

swankjesse:
Unfortunately there’s a wrinkle in this. In an unlikely-but-possible scenario we have enough information to fail after reading N+1 bytes but we’ll block until we can read N+2 bytes. This could cause us to block forever!

As a contrived example, here’s an imaginary protocol succeeding:

1. Client sends `REQUEST ALPHABET 0..3\n`
2. Server responds `ABC\n`
3. Client calls `readLineUtf8(3)` and succeeds.
4. Client sends `REQUEST ALPHABET 3..7\n`
5. Server responds `DEFG\n`
6. Client calls `readLineUtf8(4)` and succeeds.

Now we introduce a regression where the server’s accidentally sends `n` instead of `\n`. It happens.

1. Client sends `REQUEST ALPHABET 0..3\n`
2. Server responds `ABCn`
3. Client calls `readLineUtf8(3)` and this call never returns.

The client has enough information to detect the error but will wait indefinitely for a `\n` that isn’t coming. I think the awkward fix is to assume reading N+1 bytes will be sufficient, and to only read N+2 if we run into a `\r`.

ericedens:
When planning this change, my thought was that a properly-set timeout would protect against a server that sent a buggy or malevolent response. In the case described above, where the server responds with `ABCn`, the client would eventually reach a timeout and throw. (Assuming a timeout is set!)

That said, I like your idea of detecting the malformed response as soon as possible. It ensures that we don't waste time waiting for the timeout, and it protects against the case where a client _doesn't_ set a timeout.

I'll make the update in the next day or two! Thanks for the feedback :)

ericedens:
Updated the PR! I took a shot at consolidating the similar code into an abstract class; if elongating the class hierarchy doesn't sound appealing, we could also do a static method.

okio/src/main/java/okio/Buffer.java
```diff
@@ -47,7 +47,8 @@
  * returning it to you. Even if you're going to write over that space anyway.
  * This class avoids zero-fill and GC churn by pooling byte arrays.
  */
-public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
+public final class Buffer extends AbstractBufferedSource
```
swankjesse: Though I do like the code sharing, I don’t like the asymmetry here.

okio/src/main/java/okio/AbstractBufferedSource.java
```diff
@@ -0,0 +1,45 @@
...
+    return readUtf8LineStrict(Long.MAX_VALUE);
+  }
+
+  @Override public String readUtf8LineStrict(long limit) throws IOException {
+    if (limit < 0) throw new IllegalArgumentException("limit < 0: " + limit);
+    if (limit == 0) throw new EOFException("limit == 0");
+    long scanLength = limit == Long.MAX_VALUE ? Long.MAX_VALUE : limit + 1;
+    long newline = indexOf((byte) '\n', 0, scanLength);
+    Buffer buffer = buffer();
+    if (newline > -1) {
```
swankjesse: nit: we typically write `!= -1` ’cause we’re comparing a true/false condition (was it found?) vs. a range question (was it at least 0?)

okio/src/main/java/okio/AbstractBufferedSource.java
```diff
@@ -0,0 +1,45 @@
...
+
+  @Override public String readUtf8LineStrict() throws IOException {
+    return readUtf8LineStrict(Long.MAX_VALUE);
+  }
+
+  @Override public String readUtf8LineStrict(long limit) throws IOException {
+    if (limit < 0) throw new IllegalArgumentException("limit < 0: " + limit);
+    if (limit == 0) throw new EOFException("limit == 0");
+    long scanLength = limit == Long.MAX_VALUE ? Long.MAX_VALUE : limit + 1;
+    long newline = indexOf((byte) '\n', 0, scanLength);
```
swankjesse: I think using this to cause the buffer to grow is really clever

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -392,26 +392,26 @@
...
    *
    * <p>This method is safe. No bytes are discarded if the match fails, and the caller is free
    * to try another match: <pre>{@code
    *
    *   Buffer buffer = new Buffer();
    *   buffer.writeUtf8("12345\r\n");
    *
-   *   // This will throw! A newline character (\n) must be read within the limit.
-   *   buffer.readUtf8LineStrict(5);
+   *   // This will throw! A newline character (\n) must be read within limit + 2 bytes.
```
swankjesse: I think the `+ 2 bytes` thing might be a distraction? Maybe something like

```
// This will throw! There must be \r\n or \n at the limit or before it.
```
?

okio/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -475,18 +475,15 @@
...
-    byte c = (byte) 'c';
-
-    int size;
     if (factory == Factory.ONE_BYTE_AT_A_TIME) {
-      // Larger segment sizes for ONE_BYTE_AT_A_TIME
-      // cause Travis to kill the test.
-      size =  Segment.SIZE + 10;
-    } else {
-      size = Segment.SIZE * 5;
+      // When run on Travis, ONE_BYTE_AT_A_TIME
```
swankjesse: Ick. We can probably do better

swankjesse: (but that’s off topic for this PR)

okio/src/test/java/okio/ReadUtf8LineTest.java
```diff
@@ -128,11 +146,38 @@
...
+    assertEquals("abc", source.readUtf8LineStrict(3));
+    assertTrue(source.exhausted());
+
+    data.writeUtf8("abc\r\n");
+    assertEquals("abc", source.readUtf8LineStrict(3));
+    assertTrue(source.exhausted());
+
+    data.writeUtf8("abc\r");
+    try {
+      source.readUtf8LineStrict(3);
```
swankjesse: nice test

ericedens:
Thanks for the review! I've made the changes that you suggested and posted the updated version. :)

swankjesse:
Neat. I was expecting to see a tricky case around needing to refill the buffer when it ends with `\r` right at the scan limit. But that case wasn’t there, and when I looked further I found a bug in the new `indexOf(byte, long, long)` API...

One of the trickiest problems with Okio is that we want to always read what you tell us to, but we also want to avoid reading when we don’t strictly need to. Otherwise our API blocks when it shouldn’t, and this can break interactive protocols.

In this case I think `BufferedSource.indexOf()` refills its buffer one more time than it needs to before giving up and returning -1. What we probably want is a test case on the slow reader to confirm that exactly the bytes needed to be read were read and nothing else.

Wanna fix it? Or I can.

ericedens:
I *think* the tricky case that you're describing is covered here when `Segment.SIZE - 1`. `\r` will be at the end of the first segment, and `realBufferedSource` will need to hit the source again to check for `\n`. That was the intent, anyway! (Maybe I'm not understanding your case, though?)

```
  @Test public void readUtf8LineStrictWithLimits() throws IOException {
    int[] lens = {1, Segment.SIZE - 1, Segment.SIZE, Segment.SIZE * 10};
    for (int len : lens) {
       ...
      data.writeUtf8(TestUtil.repeat('a', len)).writeUtf8("\r\n");
      assertEquals(len, source.readUtf8LineStrict(len).length());
      source.readUtf8();
```

Good catch on `indexOf(byte, long, long)`. I added a guard to avoid unnecessary reading, along with a test.

okio/src/test/java/okio/RealBufferedSourceTest.java
```diff
@@ -78,6 +78,19 @@
     }
   }
 
+  @Test public void indexOfStopsReadingAtLimit() throws Exception {
```
swankjesse: nice

swankjesse:
Thanks! Merged manually here:
https://github.com/square/okio/commit/11ad43ea84e04cbaeb6322d522ee3bd5fd733d8c

I made one slight behavior change to your PR before merging. We have this test:
```
    data.writeUtf8("abc\n");
    try {
      source.readUtf8LineStrict(2);
      fail();
    } catch (EOFException expected) {
    }
```

On a slow source we were eagerly reading 2 bytes beyond the limit so this was fetching both `c` and `\n`. I changed it to not read the `\n` after discovering the character at `scanLimit` was not `\r`. In practice this will likely not make a difference, but in theory it avoids a blocking read when one isn’t strictly necessary.

Other than that, great PR and thanks for your patience with me.

ericedens:
Good call on avoiding the extra read if the range doesn't end with `\r`! Thanks for all of the reviewing -- I really appreciate the feedback :)

Issue #280, opened 2017-01-19
Provide the ability to set the compression level in GzipSink


wdroste:
Currently GzipSink hard codes the underlying Deflater to DEFAULT_COMPRESSION it would be nice if that was an optional argument to the constructor to set the level so one could choose BEST_SPEED or BEST_COMPRESSION.

wdroste:
It looks like when I deleted my fork it removed all the comments. I'll closed this ticket. 

The issue was resolved by using:
`deflater().setLevel(..`

It was obvious to me that you could mutate the behavior of a sink after it was created.

PR #281, opened 2017-01-19
#280 -Added a compress level argument to the GzipSink constructor and…

wdroste:
… included the default.


swankjesse:
I’d prefer you use `deflater().setLevel()`. Overloaded constructors make your code less obvious and make Okio bigger.

wdroste:
I'll close the ticket. 

I never thought you could/should mutate the behavior of a sink/source after it was created so didn't even look, I'll pay more attention in the future now that I'm aware of the design decision.

PR #284, opened 2017-02-05
Add BufferedSource.readUtf8Line(long limit)

ericedens:
Here we go!

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -632,35 +632,42 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
   }
 
   @Override public String readUtf8Line() throws EOFException {
-    long newline = indexOf((byte) '\n');
-
-    if (newline == -1) {
-      return size != 0 ? readUtf8(size) : null;
-    }
-
-    return readUtf8Line(newline);
+    return readUtf8Line(Long.MAX_VALUE);
   }
 
   @Override public String readUtf8LineStrict() throws EOFException {
     return readUtf8LineStrict(Long.MAX_VALUE);
   }
 
+  @Override public String readUtf8Line(long limit) throws EOFException {
+    if (size == 0) return null;
+    long newline = findNewline(limit);
+    return newline != -1 ? readUtf8AndTrimNewline(newline) : readUtf8(Math.min(size, limit));
+  }
+
   @Override public String readUtf8LineStrict(long limit) throws EOFException {
-    if (limit < 0) throw new IllegalArgumentException("limit < 0: " + limit);
-    long scanLength = limit == Long.MAX_VALUE ? Long.MAX_VALUE : limit + 1;
-    long newline = indexOf((byte) '\n', 0, scanLength);
-    if (newline != -1) return readUtf8Line(newline);
-    if (scanLength < size()
-        && getByte(scanLength - 1) == '\r' && getByte(scanLength) == '\n') {
-      return readUtf8Line(scanLength); // The line was 'limit' UTF-8 bytes followed by \r\n.
-    }
+    long newline = findNewline(limit);
+    if (newline != -1) return readUtf8AndTrimNewline(newline);
     Buffer data = new Buffer();
     copyTo(data, 0, Math.min(32, size()));
     throw new EOFException("\\n not found: limit=" + Math.min(size(), limit)
         + " content=" + data.readByteString().hex() + '…');
   }
 
-  String readUtf8Line(long newline) throws EOFException {
+  /** Returns index of \n if it is within limit + 2 bytes, or -1 */
+  private long findNewline(long limit) throws EOFException {
+    if (limit < 0) throw new IllegalArgumentException("limit < 0: " + limit);
+    long scanLength = limit == Long.MAX_VALUE ? Long.MAX_VALUE : limit + 1;
+    long newline = indexOf((byte) '\n', 0, scanLength);
+    if (newline != -1) return newline;
+    if (scanLength < size()
+        && getByte(scanLength - 1) == '\r' && getByte(scanLength) == '\n') {
+      return scanLength; // The line was 'limit' UTF-8 bytes followed by \r\n.
+    }
+    return -1L;
+  }
+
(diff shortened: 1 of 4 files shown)


okio/src/main/java/okio/BufferedSource.java
```diff
@@ -381,6 +381,28 @@
   String readUtf8Line() throws IOException;
 
   /**
+   * Like {@link #readUtf8Line()}, but allows the caller to specify the longest
+   * allowed match.
+   *
+   * <p>The returned string will have at most {@code limit} UTF-8 bytes, and the maximum number
+   * of bytes scanned is {@code limit + 2}. If {@code limit == 0} this will return an empty string.
+   * If the source is exhausted, this returns null.{@code
```
swankjesse: `<pre>`

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -381,6 +381,28 @@
   String readUtf8Line() throws IOException;
 
   /**
+   * Like {@link #readUtf8Line()}, but allows the caller to specify the longest
+   * allowed match.
+   *
+   * <p>The returned string will have at most {@code limit} UTF-8 bytes, and the maximum number
+   * of bytes scanned is {@code limit + 2}. If {@code limit == 0} this will return an empty string.
```
swankjesse: I wonder if it’s simpler to be explicit about which bytes are behind the `+2` ? Like _“consumes up to limit bytes plus `\n` or `\r\n`”_

okio/src/main/java/okio/Buffer.java
```diff
@@ -632,35 +632,42 @@ int selectPrefix(Options options) {
...
+    long newline = findNewline(limit);
+    if (newline != -1) return readUtf8AndTrimNewline(newline);
+    Buffer data = new Buffer();
+    copyTo(data, 0, Math.min(32, size()));
+    throw new EOFException("\\n not found: limit=" + Math.min(size(), limit)
+        + " content=" + data.readByteString().hex() + '…');
+  }
+
+  /** Returns index of \n if it is within limit + 2 bytes, or -1 */
+  private long findNewline(long limit) throws EOFException {
```
swankjesse: nice

okio/src/main/java/okio/RealBufferedSource.java
```diff
@@ -197,33 +197,42 @@
...
+    long newline = findNewline(limit);
+    if (newline != -1) return buffer.readUtf8AndTrimNewline(newline);
+    Buffer data = new Buffer();
+    buffer.copyTo(data, 0, Math.min(32, buffer.size()));
+    throw new EOFException("\\n not found: limit=" + Math.min(buffer.size(), limit)
+        + " content=" + data.readByteString().hex() + '…');
+  }
+
+  /** Returns index of \n if it is within limit + 2 bytes, or -1 */
+  private long findNewline(long limit) throws IOException {
```
swankjesse: Maybe this should be namd `indexOfNewline` ?

okio/src/test/java/okio/ReadUtf8LineTest.java
```diff
@@ -194,6 +194,55 @@
...
+  @Test public void readUtf8LineWithLimits() throws IOException {
+    for (int limit : new int[] {
+        5, 6, 7, 10000
+    }) {
+      data.writeUtf8("panda\r\n");
+      assertEquals("panda", source.readUtf8Line(limit));
+      assertTrue(data.exhausted());
+    }
+
+    // newline not within limit
```
swankjesse: nit: http://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

okio/src/test/java/okio/ReadUtf8LineTest.java
```diff
@@ -194,6 +194,55 @@
...
+        "panda", "panda\n", "panda\r\n",
+    }) {
+      data.writeUtf8(input);
+      assertEquals("p", source.readUtf8Line(1));
+      assertEquals("an", source.readUtf8Line(2));
+      assertEquals("da", source.readUtf8Line(2));
+      assertTrue(data.exhausted());
+    }
+
+    // Trim trailing newlines
```
swankjesse: nice

swankjesse: generally I think it’s better to avoid loops in test cases, instead using helper methods. In this case you could imagine a method like so:

```
private void testTrimTrailingNewlines(String input) {
  Buffer data = new Buffer();
  data.writeUtf8(input);
  assertEquals("panda", source.readUtf8Line(5));
  assertTrue(data.exhausted());
}
```
Then your code reads like so:
```
trimTrailingNewlines("panda");
trimTrailingNewlines("panda\n");
trimTrailingNewlines("panda\r\n");
```

Two reasons why I like this structure better. First, if it fails you get a stacktrace that tells you exactly which input failed. With this a failure on `data.exhausted()` won’t tell you which of the inputs was bad. Second, control flow in test cases is risky because test cases don’t have test cases.

okio/src/test/java/okio/ReadUtf8LineTest.java
ericedens: Wow, that's a good article :)

okio/src/test/java/okio/ReadUtf8LineTest.java
ericedens: This is good feedback! As you mention it, if one of these cases starts failing, it wouldn't be clear which input caused the problem.

ericedens:
> Should it throw instead? I’m starting to question whether it makes sense to offer a line limit for this method.

I've been unsure whether this should be included as well, and to be honest I can't think of where I'd use it in preference of `readUtf8LineStrict(byte, long, long)` or `readUtf8(long)`. 

I'm more than happy to close this PR; we can always revive it if further use cases arise!

swankjesse:
Yeah I think that we close this for now, and revive it if/when somebody comes asking.

Does that mean we can close the issue now? Sweet.
https://github.com/square/okio/issues/153

ericedens:
Cool! Yeah, I'm good with closing #153.

PR #285, opened 2017-02-08
Close BufferedSource in README sample

NightlyNexus:
Is this too much ceremony for a README sample?


JakeWharton:
Maybe try-with-resources would be more appropriate?

PR #297, opened 2017-04-01
Write '?' for a partial surrogate rather than throwing.

swankjesse:
https://github.com/square/okhttp/pull/3189/files


cketti:
Why the question mark and not the replacement character U+FFFD?

swankjesse:
Consistency with java.io APIs. The `?` is worse, but I think consistency is worth it.

iglaweb:
Important fix, thank you!

Issue #283, opened 2017-02-03
API to measure how long a string will be when UTF-8 encoded


swankjesse:
We should promote this method from Wire.
https://github.com/square/wire/blob/master/wire-runtime/src/main/java/com/squareup/wire/ProtoWriter.java#L71

It’s probably worth the wrinkle to put it in. There’s actually _different_ ways to implement this and it’s important for the caller to use the exact strategy we use in ByteString.
 * If the string contains `\0` bytes then we have a UTF-8 / MUTF-8 disagreement. The strategy used in  `DataOutputStream` will return a result that’s off by one.
 * If the string contains invalid surrogate pairs we also count differently.

swankjesse:
FYI https://en.wikipedia.org/wiki/UTF-8#Modified_UTF-8

PR #300, opened 2017-04-10
New Utf8.size() method to get a string's UTF-8 length.

swankjesse:
Closes: https://github.com/square/okio/issues/283

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -843,7 +843,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public Buffer writeUtf8(String string, int beginIndex, int endIndex) {
     if (string == null) throw new IllegalArgumentException("string == null");
-    if (beginIndex < 0) throw new IllegalAccessError("beginIndex < 0: " + beginIndex);
+    if (beginIndex < 0) throw new IllegalArgumentException("beginIndex < 0: " + beginIndex);
     if (endIndex < beginIndex) {
       throw new IllegalArgumentException("endIndex < beginIndex: " + endIndex + " < " + beginIndex);
     }
--- a/okio/src/main/java/okio/Utf8.java
+++ b/okio/src/main/java/okio/Utf8.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright (C) 2017 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -843,7 +843,7 @@ public void clear() {
 
   @Override public Buffer writeUtf8(String string, int beginIndex, int endIndex) {
     if (string == null) throw new IllegalArgumentException("string == null");
-    if (beginIndex < 0) throw new IllegalAccessError("beginIndex < 0: " + beginIndex);
+    if (beginIndex < 0) throw new IllegalArgumentException("beginIndex < 0: " + beginIndex);
```
swankjesse: `IAE<tab>`

okio/src/main/java/okio/Utf8.java
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio;
+
+/** APIs for working with UTF-8 strings. */
+public final class Utf8 {
```
swankjesse: Would it be simpler as just `Okio.utf8Size()` ? I think this is pretty discoverable

okio/src/main/java/okio/Utf8.java
```diff
@@ -0,0 +1,82 @@
...
+  private Utf8() {
+  }
+
+  /**
+   * Returns the number of bytes used to encode {@code string} as UTF-8 when using {@link
+   * ByteString#encodeUtf8} or {@link Buffer#writeUtf8(String)}.
```
cketti: `ByteString#encodeUtf8()` currently uses `String#getBytes(Charset)`. Is the behavior of `getBytes()` with UTF-8 as charset sufficiently documented to guarantee this method exhibits the same behavior as Java runtime implementations?

swankjesse: Good question. We’ve got test cases for a bunch of interesting cases, and we strive to be compatible. Could probably use even more tests though – I know runs of mismatched surrogates have differed between Dalvik and Android in at least one case.

cketti: Sounds like the `ByteString#encodeUtf8()` part should be dropped from the JavaDoc. Alternatively, `ByteString#encodeUtf8()` could be changed to manually UTF-8-encode the input string.

swankjesse: What about just testing to confirm it’s consistent? The JDK’s implementation is faster, but only when it needs to allocate a `byte[]` anyway.

okio/src/main/java/okio/Utf8.java
JakeWharton: My only (very minor) complaint about having a `Utf8` class is that none of the other operations dealing with UTF are exposed here.

swankjesse: Lemme add Javadoc referencing all the other UTF-8 things

swankjesse:
screen shot 2017-04-11 at 10 23 29 pm

With apologies for bad default CSS.

PR #307, opened 2017-05-02
Import jsr305 and use it to mark @Nullable stuff.

swankjesse:
The result is that IntelliJ + Kotlin correctly infers the right nullability
for Okio's APIs. I used IntelliJ's "Specify types explicitly" quick fix on
the 'a' and 'b' variables below and it did the right thing.

    fun specifyTypeExplicitly(source: BufferedSource) {
      val a: String? = source.readUtf8Line()
      val b: String = source.readUtf8LineStrict()
    }

### DIFF
--- a/okio/pom.xml
+++ b/okio/pom.xml
@@ -14,11 +14,15 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.google.code.findbugs</groupId>
+      <artifactId>jsr305</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>org.codehaus.mojo</groupId>
       <artifactId>animal-sniffer-annotations</artifactId>
       <optional>true</optional>
     </dependency>
-
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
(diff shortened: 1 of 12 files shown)


swankjesse:
I had 3 choices for `@Nullable` (Android, JetBrains, JSR-305). I decided to go with JSR-305 for these reasons:

1. The package-level `@ParametersAreNonnullByDefault` means we don’t need to litter annotations everywhere.
2. Guava uses it, so many applications already import it

I’m not happy adding a dependency from Okio onto anything. I’m convincing myself that it’s okay.

adriancole:
Thanks for mentioning rationale. I have many times been tempted to add 305
but perhaps let my no deps dogma get the best of me and didnt. 305 is
something so common feels like standard JRE at this point.

On 2 May 2017 10:51 am, "Jesse Wilson" <notifications@github.com> wrote:

I had 3 choices for @Nullable (Android, JetBrains, JSR-305). I decided to
go with JSR-305 for these reasons:

   1. The package-level @ParametersAreNonnullByDefault means we don’t need
   to litter annotations everywhere.
   2. Guava uses it, so many applications already import it

I’m not happy adding a dependency from Okio onto anything. I’m convincing
myself that it’s okay.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/square/okio/pull/307#issuecomment-298484987>, or mute
the thread

swankjesse:
I’m going to try making this a provided dependency. That’d prevent it from showing up in end-users builds. It shouldn’t impact static analysis.

PR #321, opened 2017-09-28
Make Buffer conform to BufferedSource wrt/ EOF handling

ksperling:
Also throw IllegalArgument instead of ArrayIndexOutOfBounds for
negative byteCount.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -103,7 +103,10 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
   }
 
   @Override public void require(long byteCount) throws EOFException {
-    if (size < byteCount) throw new EOFException();
+    if (size < byteCount) {
+      clear(); // exhaust ourselves
+      throw new EOFException();
+    }
   }
 
   @Override public boolean request(long byteCount) {
@@ -112,7 +115,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   @Override public InputStream inputStream() {
     return new InputStream() {
-      @Override public int read() {
+      @Override public int read() throws IOException {
         if (size > 0) return readByte() & 0xff;
         return -1;
       }
@@ -273,8 +276,8 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return result;
   }
 
-  @Override public byte readByte() {
-    if (size == 0) throw new IllegalStateException("size == 0");
+  @Override public byte readByte() throws EOFException {
+    if (size == 0) throw new EOFException();
 
     Segment segment = head;
     int pos = segment.pos;
@@ -304,14 +307,14 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     }
   }
 
-  @Override public short readShort() {
-    if (size < 2) throw new IllegalStateException("size < 2: " + size);
+  @Override public short readShort() throws EOFException {
+    if (size == 0) throw new EOFException();
 
     Segment segment = head;
     int pos = segment.pos;
     int limit = segment.limit;
 
-    // If the short is split across multiple segments, delegate to readByte().
+    // If the first segment does not have a full short, delegate to readByte().
     if (limit - pos < 2) {
       int s = (readByte() & 0xff) << 8
           |   (readByte() & 0xff);
@@ -333,14 +336,14 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     return (short) s;
   }
 
-  @Override public int readInt() {
-    if (size < 4) throw new IllegalStateException("size < 4: " + size);
+  @Override public int readInt() throws EOFException {
+    if (size == 0) throw new EOFException();
 
     Segment segment = head;
     int pos = segment.pos;
     int limit = segment.limit;
 
-    // If the int is split across multiple segments, delegate to readByte().
+    // If the first segment does not have a full int, delegate to readByte().
     if (limit - pos < 4) {
       return (readByte() & 0xff) << 24
           |  (readByte() & 0xff) << 16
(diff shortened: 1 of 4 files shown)


ksperling:
I'm not entirely sure if `EOFException` is the right behaviour, given that e.g. `readShort` throws `IllegalStateException`, but `ArrayIndexOutOfBoundsException` seems wrong.

ksperling:
Looking further at what other `Buffer.read*` methods do, 

Throws `EofException`
* `readFully(Buffer, long)`
* `readFully(byte[])`

Throws `ArrayIndexOutOfBoundsException`
* `readByteArray(long)` (but declares EOFException)
* `readByteString(long)` (but declares EOFException)
* `readUtf8(long)` (but declares EOFException)
* `readString(long, Charset)` (but declares EOFException)

Throws `IllegalStateException`
* `read{Byte,Short,Int,Long}[Le]`
* `readDecimalLong` (for size == 0)
* `readHexadecimalUnsignedLong` (for size == 0)

So this seems somewhat consistent in that `readFully` variants take a user-supplied destination, perform a partial read, and throw `EOFException`, the other methods allocate and return a value, and throw a `RuntimeException` without consuming anything from the buffer.

If this difference in semantics between these two groups of methods is intentional, the `throws EOFException` should be removed from those methods (and maybe some class-level docs added explaining the general pattern of EOF handling). Also, `ArrayIndexOutOfBoundsException` still seems a pretty strange choice to me when there is no array involved. Maybe it should be `IllegalStateException`, so that e.g. `readLong` and `readByteArray(8)` behave consistently.

okio/src/test/java/okio/BufferTest.java
```diff
@@ -582,4 +583,14 @@
     Buffer buf = new Buffer().write(new byte[] { 0, 1, 2, 3 });
     assertEquals(1, buf.snapshot(1).size());
   }
+
+  @Test public void readByteArrayThrowsEofException() throws Exception {
```
JakeWharton: Can you try putting this in `BufferedSourceTest`? This will ensure behavior parity between `RealBufferedSource` and `Buffer` for this method.

okio/src/main/java/okio/Buffer.java
```diff
@@ -760,11 +760,17 @@ String readUtf8Line(long newline) throws EOFException {
   }
 
   @Override public byte[] readByteArray(long byteCount) throws EOFException {
-    checkOffsetAndCount(size, 0, byteCount);
-    if (byteCount > Integer.MAX_VALUE) {
+    if (byteCount < 0) {
+      throw new IllegalArgumentException("byteCount < 0: " + byteCount);
+    } else if (byteCount > Integer.MAX_VALUE) {
```
JakeWharton: super minor nit: we typically don't put `else` after `throw` or `return` and instead just do separate `if` statements.

JakeWharton:
There's definitely a fundamental disconnect happening here as a result of two competing desires. The first is for behavior parity between `RealBufferedSource` and `Buffer` when treated as a `BufferedSource`. The second is the removal of checked exceptions from `Buffer`'s methods because it operates entirely in-memory.

I think I'm supportive of prioritizing the first over the second. @swankjesse?

okio/src/main/java/okio/Buffer.java
post2shyam: Any particular reason for minor nit? Or just a coding convention ?

okio/src/test/java/okio/BufferTest.java
```diff
@@ -582,4 +583,14 @@
...
+    Buffer buf = new Buffer().write(new byte[] { 42 });
```
swankjesse: supernit: `buf` –> `buffer`

It makes the code a little more boring if things have fewer names. Trying to guess what short forms are in use is more work than I want to do!

ksperling:
@JakeWharton That definitely makes sense. 

The documentation of `BufferedSource` doesn't explicitly specify the exact behaviour with respect to EOF for every method, but the ones that do say they throw `EOFException`. Another aspect that's not fully specified is whether or not encountering an `EOFException` from a reading method means that all available input has been read/consumed; generally I would probably expect that to be the case, and that's what `RealBufferedSource` seems to do.

The big problem with making `Buffer` fully conform to the `BufferedSource` interface is that it would require the breaking API change of adding `throws EOFException` to all the read methods that don't currently have it.

The other option would be to go the other way and have `Buffer` never throw `EOFException` for  reads, i.e. remove the `throws` from `readByteArray`, `readByteString`, `readFully` etc and ensure they all have consistent behaviour wrt/ to a read encountering EOF not consuming any bytes and throwing an unchecked exception (`BufferUnderflowException` would IMO be a good candidate instead of `ArrayIndexOutOfBoundsException`). This would not break the API, but then `Buffer` wouldn't really conform to the semantics of the `BufferedSource` interface. This is already the status quo for some operations though.

This could be resolved by having `Buffer` not implement `BufferedSource` at all and instead provide a "`BufferBufferedSource`" that implements `BufferedSource` on top of a `Buffer` and adapts the EOF semantics. Of course that's also a breaking API change...

swankjesse:
Agreed that exceptions thrown here are inconsistent. In my experience it hasn’t been too much of a problem because it’s possible to know ahead-of-time whether a `Buffer` is going to throw an exception or not – just check its size. Obviously this isn’t the case for generic code that doesn’t know that the `BufferedSource` it’s operating on is actually a `Buffer`.

What action to take? I’m very reluctant to change the behavior of any method that currently throws a checked exception to taking an unchecked exception. This kind of change crashes apps and makes people grumpy.

Going the other direction – changing code that throws an unchecked exception to throwing a checked exception is generally safe. Reasonable programs never throw these exceptions unless they’re crashing, so they aren’t likely to happen in practice.

We can actually add checked exceptions without breaking binary compatibility. The source compatibility sucks, but I could be convinced that it would be reasonable.

I think the next step is homework – see what the damage is in libraries like OkHttp and Retrofit when adding these new exceptions. If the cost to fix isn’t too high, maybe we make a source-incompatible change.

JakeWharton:
Also worth noting we've made source incompatible changes like this before,
but thus far only removing checked exceptions, not adding them.

ksperling:
Wire tests pass. Okhttp tests pass with a small patch: https://gist.github.com/ksperling/97f8a6e39ff0cb6082f3eb017453a20d

ksperling:
Any further thoughts on this? Retrofit tests also pass with a similar small patch: https://gist.github.com/ksperling/2ff7d84530433c39e771b17b979495c0

We could add `Buffer.nextByte()` that returns -1 or 0..255 similar to `InputStream.read()`, or some variants of `read*` methods that throw an unchecked `BufferUnderflowException`, e.g. `take*`, but that seems unnecessary at this stage. The use of `Buffer` for the kind of string manipulation done by HttpUrl seems like something of a corner case anyway.

swankjesse:
@ksperling thanks for opening this issue and apologies that it’s taken us so long to address it. I’ve attempted to apply the spirit of this PR here:
https://github.com/square/okio/pull/486

I’m going to document on that PR the differences between this one and that one. We should confirm we’re all happy before releasing 2.x.

PR #322, opened 2017-09-28
Return SegmentedByteString from Buffer.readByteString for large reads

ksperling:
This improves performance by avoiding a copy each on both the input and output side when doing IO with large byte strings. One use case where this happens in practice is using Wire with protobufs that contain large amounts of binary data. (We're using Okio / and Wire both on the server and Android.)

#### Alternative Ideas

An alternative (or even complementary) approach I am considering is enhancing `RealBuffered{Source,Sink}` to bypass the buffer when reading or writing a large `byte[]` or `ByteString`, similar to what `Buffered{Input,Output}Stream` does when reading/writing a block larger than it's buffer size.

A fairly straightforward way of doing this would be to create extended `Source` and `Sink` interfaces that directly support `byte[]` reads / writes. `RealBuffered{Source,Sink}` could then have corresponding sub-classes that delegate to those unbuffered methods for large reads or writes. `Okio.buffer` would create the correct variant depending on whether the source/sink supports the extended interface. The `{Input,Output}Stream` source and sink implementations would be the primary ones that would need to support this extended interface.

(Another approach would be to try to avoid any changes to `Source` and `Sink` implementations and instead have the `RealBufferedSink` turn (for example) a large `write(byte[])` into a `sink.write()` with a special one-off `Buffer` that directly wraps the array to be written. This seems very complex though due to the large API of `Buffer`)

### DIFF
--- a/okio/src/test/java/okio/BufferTest.java
+++ b/okio/src/test/java/okio/BufferTest.java
@@ -19,6 +19,7 @@ import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Random;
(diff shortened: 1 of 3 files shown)


ksperling:
The second commit updates the threshold to 40kb; based on a hopefully somewhat realistic benchmark of reading and writing a proto file of a few MB in size consisting of a mixture of meta-data and a few hundred ByteStrings of the same size. At around 40k performance breaks even, at 80k the segmented version is about 12% faster (running on a laptop with Java 8).

okio/src/test/java/okio/BufferTest.java
```diff
@@ -163,6 +164,26 @@
...
+    for (int i=0, total=0; total <= Buffer.READBYTESTRING_SEGMENT_THRESHOLD; i++) {
+      int size = Segment.SIZE/2 + i;
+      src.writeAll(new Buffer().writeUtf8(repeat((char) ('a' + i), size)));
+      segmentsSizes.add(size);
+      total += size;
+    }
+    assertEquals(segmentsSizes, src.segmentSizes());
+
+    ByteString string = src.readByteString();
+    assertTrue("ByteString is segmented", string instanceof SegmentedByteString);
```
cketti: The message is displayed when the assertion fails. In that case the statement "ByteString is segmented" is wrong. I suggest to either use "should be" or "expected…to be".

Same issue with the assertion further down.

ksperling: Thanks! Too used to AssertJ...

swankjesse:
I’m cleaning up old PRs, and I’m disappointed to realize that we sat on this without doing anything. There’s a workaround to use `snapshot()` but I admit that that’s not as nice. Happy to revisit if you’d like.

bnorm:
We did this actually with https://github.com/square/okio/pull/666. Available since 2.4.3!

PR #326, opened 2017-11-13
Add CheckReturnValue to methods.

### LINKED ISSUE #324: Add @CheckReturnValue to appropriate methods.
I'm sure lots of my old code reads instead of skipping.

NightlyNexus:
Closes #324.
Unsure how much value this adds. All but a few of these methods are simple factories that are unambiguous already.

### DIFF
--- a/okio/src/main/java/okio/AsyncTimeout.java
+++ b/okio/src/main/java/okio/AsyncTimeout.java
@@ -18,6 +18,7 @@ package okio;
 import java.io.IOException;
 import java.io.InterruptedIOException;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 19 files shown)


NightlyNexus:
tricky that you can't `@CheckReturnValue` to a type without it getting applied to all subtypes, so you have to avoid it on extensible types.

swankjesse:
I think this is probably not as useful for Okio since I don’t think anyone is struggling with this stuff. It’s most useful in things like `InputStream.skip()` where it seems like ignoring the return value is okay.

NightlyNexus:
Yep, agree!

Issue #141, opened 2015-04-17
Unsafe access to segment byte arrays


swankjesse:
```
UnsafeBuffer b = new UnsafeBuffer();
b.acquireForRead(buffer, 1024);
Audio.read(b.array, b.offset, b.count);
b.release();
```

JakeWharton:
And just so we don't forget:

> if you want a 4096 byte[] we could just allocate it and create a segment that isn't recycled to the pool. and that's an implementation detail that might change if we get bucketed pools later on

PR #331, opened 2018-01-06
Unsafe access to segment byte arrays.

swankjesse:
This offers direct access to the byte arrays inside the buffer's
segments. It's very low-level and only appropriate for low-level
integration: compression frameworks, encryption libraries, etc.

Closes: https://github.com/square/okio/issues/141

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -15,6 +15,7 @@
  */
 package okio;
 
+import java.io.Closeable;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
@@ -182,7 +183,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
     // Copy one segment at a time.
     for (; byteCount > 0; s = s.next) {
-      Segment copy = new Segment(s);
+      Segment copy = s.sharedCopy();
       copy.pos += offset;
       copy.limit = Math.min(copy.pos + (int) byteCount, copy.limit);
       if (out.head == null) {
@@ -1672,10 +1673,10 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     Buffer result = new Buffer();
     if (size == 0) return result;
 
-    result.head = new Segment(head);
+    result.head = head.sharedCopy();
     result.head.next = result.head.prev = result.head;
     for (Segment s = head.next; s != head; s = s.next) {
-      result.head.prev.push(new Segment(s));
+      result.head.prev.push(s.sharedCopy());
     }
     result.size = size;
     return result;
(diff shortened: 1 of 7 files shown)


okio/src/main/java/okio/BufferCursor.java
```diff
@@ -0,0 +1,119 @@
...
+ * TODO(jwilson): support growing the buffer at the end
+ * TODO(jwilson): support acquiring for write (with seek() that gets writable segments)
+ */
+public final class BufferCursor {
+  private Buffer buffer;
+  public long size = -1L;
+
+  private Segment segment;
+  public long offset = -1L;
+  public byte[] data;
```
swankjesse: public!

okio/src/main/java/okio/BufferCursor.java
```diff
@@ -0,0 +1,119 @@
...
+ * Direct access to the underlying byte arrays in a buffer.
+ *
+ * Note that cursors are stateful and can be reused.
+ *
+ * It is an error to mutate a buffer that is currently being operated on with a cursor.
+ *
+ * Currently this cursor only supports reading. It is an error to mutate the fields of this cursor
+ * or the values of the data[] array.
+ *
+ * Access should be in acquire/release pairs. This guarded access may be used to detect bugs where
```
JakeWharton: How about `implement Closeable` and drop the checked exception? This lets you drop 'em in try-with-resources. We do this in SQL Brite to great effect.

JakeWharton: https://square.github.io/sqlbrite/3.x/sqlbrite/com/squareup/sqlbrite3/BriteDatabase.html#newTransaction--

swankjesse: Only drawback of that is that it makes it weird to re-use the same cursor. I’m imagining tools like BrotliSink having a single cursor and reusing it so there’s no allocation for the frames of a stream.

swankjesse: Though I think the try-with-resources is probably worth it.

JakeWharton: The trick SQL Brite uses to avoid that is there's only one instance of the returned object.

So
```java
Transaction t1, t2;
try (Transaction t = db.newTransaction()) {
  t1 = t;
}
try (Transaction t = db.newTransaction()) {
  t2 = t;
}
assertSame(t1, t2);
```
would pass.

It's purely an API scoping mechanism.

So here you're remove all public APIs from `BufferCursor` except `acquire` and put it on some nested type. `BufferCursor` instances would hold a single instance of that nested type, and `acquire` would mutate it and return.

```java
BufferCursor c = new BufferCursor();
try (Something s = c.acquire(buffer)) {
  s.data...
}
try (Something s = c.acquire(buffer)) {
  s.data...
}
```

You almost want to make the nested type called `Cursor` and give the outer type some other name. Or maybe just hang this directly on `Buffer`!

```java
try (Cursor c = buffer.acquireUnsafeCursor()) {
  c.data...
}
try (Cursor c = buffer.acquireUnsafeCursor()) {
  c.data...
}
```

JakeWharton: SQL Brite details:
  * Single internal instance: https://github.com/square/sqlbrite/blob/dae34762ca41f532094433c3ac1e515865e74754/sqlbrite/src/main/java/com/squareup/sqlbrite3/BriteDatabase.java#L73
 * Returned from `newTransaction()`: https://github.com/square/sqlbrite/blob/dae34762ca41f532094433c3ac1e515865e74754/sqlbrite/src/main/java/com/squareup/sqlbrite3/BriteDatabase.java#L235

okio/src/main/java/okio/BufferCursor.java
```diff
@@ -0,0 +1,119 @@
...
+  private Buffer buffer;
+  public long size = -1L;
+
+  private Segment segment;
+  public long offset = -1L;
+  public byte[] data;
+  public int pos = -1;
+  public int limit = -1;
+
+  public void acquireReadOnly(Buffer buffer) {
```
bnorm: What if we want to start consuming at a specific offset?  The best I've been able to come up with the following, which wasn't very intuitive.  Might be more intuitive to have an `acquireReadOnly(buffer, offset)` that primed the `next()` to start at offset instead of 0.

```java
cursor.acquireReadOnly(buffer);
try {
  for (int length = cursor.seek(offset); toConsume > 0; length = cursor.next()) {
    length = (int) Math.min(length, toConsume);
    consume(cursor.data, cursor.pos, length);
    toConsume -= length;
  }
} finally {
  cursor.release();
}
```

bnorm: Could a cursor be used to simplify the `findSegmentAndOffset` labeled code in Buffer?

okio/src/main/java/okio/BufferCursor.java
bnorm: Is it an error to acquire multiple cursors on the same buffer?  Should a buffer know there is an acquired cursor so it can error reads and writes until it is released?  Hanging cursor acquisition off of `Buffer` would allow for some of these checks and also use of the cursor internally for the `findSegmentAndOffset` blocks.

okio/src/main/java/okio/BufferCursor.java
```diff
@@ -0,0 +1,119 @@
...
+      this.data = null;
+      this.pos = -1;
+      this.limit = -1;
+      return -1;
+    }
+
+    // Navigate to the segment that contains `offset` starting from our current segment if possible.
+    Segment next;
+    long nextOffset;
+    if (segment == null || this.offset > offset) {
```
bnorm: Is seeking backwards going to be useful?  Or should the cursor just be released and re-acquired?  I'm not immediately seeing the usefulness but I'm guessing you have ideas?

okio/src/main/java/okio/BufferCursor.java
JakeWharton: Good questions. I think I'm a "yes" and "yes" for them.

swankjesse: I’m actually “no” for what I think is an interesting reason.

I like “effectively immutable” objects. The best example is `Collections.unmodifiableList()` to wrap a mutable type like `ArrayList`, but even mutable types can be documented “please do not mutate me” and provide a lot of value. For example, in `Base64.java` we use a `byte[]` as an effectively immutable value. In Java almost all of the good mutable types are safe for multiple concurrent reads as long as the reference to the instance itself is safely published.

It’s not unreasonable to use `Buffer` as an effectively-immutable object. Perhaps you populate it with a bunch of data (a bloom filter? some OpenGL geometry?) and expect many concurrent reads of that data.

If we really lock down cursors to be single-operator on Buffers, it breaks these hypothetical use cases.

JakeWharton: Isn't that what readByteString() and shared segments are for though?

okio/src/main/java/okio/BufferCursor.java
swankjesse: I’m imagining some uses of cursors that don’t necessarily traverse the entire dataset. Perhaps a buffer contains sorted lines of text and we’re binary searching for a value. For this reason I think it’s easiest to keep acquire and seek separate.

We could definitely use cursors in `findSegmentAndOffset()`. Though I don’t think it would be profitable to do so. The core problem with that is that we’d either need to allocate a cursor on each use (inefficient) or forbid concurrent reads (breaks my effectively-immutable goal above).

okio/src/main/java/okio/BufferCursor.java
swankjesse: Just for binary searching a buffer.

okio/src/main/java/okio/BufferCursor.java
```diff
@@ -0,0 +1,156 @@
...
+   * must be at least the current size. Newly added capacity may span multiple segments.
+   *
+   * <p>Warning: it its the caller’s responsibility to write new data to every byte of the
+   * newly-allocated capacity. Failure to do so may cause serious security problems as the data
+   * in the returned buffers is not zero filled. The buffers may contain dirty pooled segments that
+   * hold very sensitive data from other parts of the current process.
+   *
+   * TODO(jwilson): just zero-fill this to be safe? Or expose a parameter to make that easy?
+   */
+  public void enlarge(long newSize) {
```
swankjesse: new! And super dangerous!

okio/src/main/java/okio/BufferCursor.java
swankjesse: Yup. Though I’m still reluctant to have APIs where ”reading operations” do mutations.

bnorm: Is it worth having a `ReadCursor` and a `WriteCursor`?

okio/src/main/java/okio/BufferCursor.java
bnorm: In cases when you don't know how many bytes are going to be needed, there might need to be a way trim the buffer.  I'm thinking specifically of the `InflaterSource` and the `DeflaterSink` and how they might use a cursor.
```java
long offset = sink.size(); 
cursor.enlarge(...);
cursor.seek(offset);
int bytesInflated = inflater.inflate(cursor.data, cursor.pos, cursor.limit - cursor.pos);
cursor.trim(offset + bytesInflated);
```

Also, might be handy if enlarge returns the old size for things like `cursor.seek(cursor.enlarge(...))` which will get you immediately to the write point.

swankjesse: Yeah, I agree that this needs to be `resize()` and should either grow or truncate the buffer.

bnorm: I also found this method helpful.  The idea being that I want the next available set of bytes for writing but don't want to allocate another segment unless needed.
```java
public void grow() {
  Segment tail = buffer.writableSegment(1);
  resize(size + Segment.SIZE - tail.limit);
}
```

bnorm:
Last night I messed around with using a cursor for the `findSegmentAndOffset ` labeled code in `Buffer` and found it worked quite well.  The only problem was that some unit tests were taking a really long time.  Almost 10x as long.  I dug in and found the forward-only cursor seeking was the problem.  I re-worked the seek code so it could go both forwards and backwards from either the beginning, current offset, or the end which brought the performance inline with how `findSegmentAndOffset` is today.

Regardless of using a cursor internally in `Buffer` to replace the `findSegmentAndOffset` code, this is probably still a performance improvement we want.  Here's what I came up with, but it could probably be optimized even further: https://github.com/square/okio/compare/jwilson.0106.buffer_cursor...bnorm:bn/seek_performance

swankjesse:
Oh that’s amazing. Do you mind doing our contributor agreement thing so I can take your code?

swankjesse:

bnorm:
Sure thing!  I signed it.

swankjesse:
Here’s a sample usage of an API I’d like to propose.

```
    Buffer xoxo = new Buffer();

    try (Buffer.UnsafeCursor cursor = xoxo.readAndWriteUnsafe()) {
      cursor.resize(1_000_000L);

      while (cursor.next() != -1) {
        Arrays.fill(cursor.data, cursor.pos, cursor.limit, (byte) 'x');
      }

      cursor.seek(3L);
      cursor.data[cursor.pos] = 'o';

      cursor.seek(1L);
      cursor.data[cursor.pos] = 'o';

      cursor.resize(4L);
    }
```

I like the name `Buffer.UnsafeCursor` because it emphasizes that this is dangerous, and that it has a position within a Buffer.

It has these public members:

```
    public long size;
    public long offset;
    public byte[] data;
    public int pos;
    public int limit;
```

I’m not sure about `pos` or `position`. I don’t like abbreviations at all in Java.

I don’t know if it’s useful to have another copy of `size` here. Could probably drop that.

You get an instance with `readOnlyUnsafe()` or `readAndWriteUnsafe()`. You can also instantiate an `UnsafeCursor` yourself and pass it to one of these methods. In either case `close()` detaches the buffer from the cursor. Cursors can be reused, and I expect that to be handy in inner-loop things like compressors.

JGulbronson:
Few comments on the API

- `cursor.resize` isn't clear it's resizing the underlying Buffer, possibly `cursor.resizeBuffer` instead?
- `cursor.limit` isn't clear if it's the max number of items, or the end of the "slice". (similar comment for `pos`. Would rename to `start` and `end`.
- It's not immediately clear what the difference between `offset` and `start` are, need some way to delineate between a Buffer offset and a "slice" start.

swankjesse:
### Alright, I think this is ready for review.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1696,4 +1697,372 @@ public ByteString snapshot(int byteCount) {
...
+        for (long bytesToAdd = newSize - buffer.size; bytesToAdd > 0; ) {
+          Segment tail = buffer.writableSegment(1);
+          long segmentBytesToAdd = Math.min(bytesToAdd, Segment.SIZE - tail.limit);
+          tail.limit += segmentBytesToAdd;
+          bytesToAdd -= segmentBytesToAdd;
+        }
+      }
+
+      buffer.size = newSize;
+      buffer.size = newSize;
```
bnorm: Double assignment

okio/src/main/java/okio/Buffer.java
```diff
@@ -1696,4 +1697,372 @@ public ByteString snapshot(int byteCount) {
...
+   *   }
+   * }</pre>
+   *
+   * <p>Change the capacity of a buffer with {@link #resizeBuffer}. This is only permitted for
+   * read+write cursors. The buffer's size always changes from the end: shrinking it removes bytes
+   * from the end; growing it adds capacity to the end.
+   *
+   * <h3>Warnings</h3>
+   *
+   * <p>Most application developers should avoid this API. Those that do should respect these
```
bnorm: > Those that do should respect these warnings.

"Those that must use this API should respect these warnings"?

bnorm:
Looks great.  I really like the form of the final API.  Only a couple final questions.

1. Will this code be documented as experimental in anyway?  Subject to breaking changes in the next release?

2. Did you still want the performance improved seek code I provided a couple days ago?  I noticed the current code still only seeks forward.

3. Thoughts on resize returning the old size of the `Buffer`?  Could help users that are growing the `Buffer` and want to seek directly to the start of the new data.

4. Thoughts on a resize-like method to grow the `Buffer` to pad the last `Segment` or add another one if already full?  The analog of `next()` but for writing.  Use case would be for repeatedly writing unknown sized data into a `Buffer`, similar to InflaterSource and DeflaterSink.  Could help avoid churn of adding and removing `Segements`.  Something like this:
```java
public void growBuffer() {
  Segment tail = buffer.writableSegment(1);
  resizeBuffer(buffer.size + Segment.SIZE - tail.limit);
}
```


Otherwise, it looks awesome, and I can't wait to use it.  Fantastic work as usual.

swankjesse:
Thanks again @bnorm. 

1. I don’t think so. Our segment design is unlikely to change. (Changing it would have substantial performance consequences for users.)
2. I sure did. I’d like to add that in a follow-up PR.
3. Great idea. Done.
4. Yes, I think we need this. I was experimenting with using this new API to do streaming encryption, and it is definitely missing. We need something that grows the buffer, and that also does so with a minimum number of contiguous bytes.

okio/src/main/java/okio/Buffer.java
```diff
@@ -1696,4 +1697,376 @@ public ByteString snapshot(int byteCount) {
...
+     * <p>Warning: it its the caller’s responsibility to write new data to every byte of the
+     * newly-allocated capacity. Failure to do so may cause serious security problems as the data
+     * in the returned buffers is not zero filled. The buffers may contain dirty pooled segments
+     * that hold very sensitive data from other parts of the current process.
+     *
+     * @return the previous size of the buffer.
+     */
+    public long resizeBuffer(long newSize) {
+      if (buffer == null) {
+        throw new IllegalStateException("not acquired");
```
JakeWharton: s/acquired/attached to a buffer/

okio/src/main/java/okio/Buffer.java
```diff
@@ -1696,4 +1697,376 @@ public ByteString snapshot(int byteCount) {
...
+      // Buffer.writableSegment() invalidates our local segment state. Recompute it.
+      segment = null;
+      if (offset != -1L) seek(offset);
+
+      return oldSize;
+    }
+
+    @Override public void close() {
+      // TODO(jwilson): use edit counts or other information to track unexpected changes?
+      if (buffer == null) throw new IllegalStateException("not acquired");
```
JakeWharton: s/acquired/attached to a buffer/

okio/src/main/java/okio/Buffer.java
```diff
@@ -1696,4 +1697,376 @@ public ByteString snapshot(int byteCount) {
     if (byteCount == 0) return ByteString.EMPTY;
     return new SegmentedByteString(this, byteCount);
   }
+
+  public UnsafeCursor readUnsafe() {
+    return readUnsafe(new UnsafeCursor());
+  }
+
+  public UnsafeCursor readUnsafe(UnsafeCursor unsafeCursor) {
+    if (unsafeCursor.buffer != null) throw new IllegalStateException("already acquired");
```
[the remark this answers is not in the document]
JakeWharton: s/acquired/attached to a buffer/

swankjesse:
@bnorm’s proposed `growBuffer()` method, implemented here:
https://github.com/square/okio/pull/342

PR #333, opened 2018-01-12
Allow Buffer.getByte() to search backwards for the desired segment

bnorm:
Uses of Buffer.getByte() within the Okio project access it more often relative to the end of the Buffer instead of from the beginning.  Specifically the ReadBufferedSource.readHexadecimalUnsignedLong() and ReadBufferedSource.readDecimalLong() methods make heavy use of retrieving the latest byte from the Buffer and with this change, the unit tests for these methods with the OneByteAtATime source no longer take over a second to run.

A benchmark for this behavior has been created and verified to show significant improvement.

Before:
```
Benchmark                                    (bufferSize)    Mode  Samples  Score   Error  Units
c.s.o.b.GetByteBenchmark.getByteBeginning         2097152  sample   160460  0.037 ± 0.003  us/op
c.s.o.b.GetByteBenchmark.getByteEnd               2097152  sample   107573  1.481 ± 0.018  us/op
c.s.o.b.GetByteBenchmark.getByteMiddle            2097152  sample   112678  0.721 ± 0.013  us/op
```

After:
```
Benchmark                                    (bufferSize)    Mode  Samples  Score   Error  Units
c.s.o.b.GetByteBenchmark.getByteBeginning         2097152  sample   167078  0.035 ± 0.003  us/op
c.s.o.b.GetByteBenchmark.getByteEnd               2097152  sample   136939  0.036 ± 0.003  us/op
c.s.o.b.GetByteBenchmark.getByteMiddle            2097152  sample   120560  0.672 ± 0.015  us/op
```

In total, during all unit tests, the getByte() method is invoked 87,467 times.  Roughly 60% (52,152) of those are for bytes in the second half of the buffer.  Without the change, this results in an additional 67,196,832 passes through the segment search loop.  With the change, there are only an additional 6,213 passes through the forward loop and an additional 10,281 passes through the backwards loop.

### DIFF
--- a/benchmarks/src/main/java/com/squareup/okio/benchmarks/GetByteBenchmark.java
+++ b/benchmarks/src/main/java/com/squareup/okio/benchmarks/GetByteBenchmark.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright (C) 2018 Square, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okio.benchmarks;
+
+import java.io.IOException;
+import java.util.concurrent.TimeUnit;
+import okio.Buffer;
+import org.openjdk.jmh.Main;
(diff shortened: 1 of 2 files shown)


swankjesse:
Awesome.

swankjesse:
Nice PR.

For `select()`, I don’t think the unit tests are necessarily representative of real-world use. I spent a lot of time trying to tune `select()` to be faster for JSON parsing. I wrote lots of fancy code inside Options to precompute stuff to save work if options contained common prefixes.

In the end the naive brute force won. How frustrating.

If you’d like to take another run at it, please do. The thing I think we’re optimizing here is downstream users of select(). Most notably, JSON parsing in Moshi. In such cases the number of options is typically the number of fields of a bound object; typically between 2 and 10 and rarely greater than 20.

The API design is intended to permit better optimizations eventually, and perhaps even ones that are asymmetric. If you can find something that’s really fast but only works up to 4 options, that’s allowed!

PR #338, opened 2018-01-30
Buffer cursor seek performance

bnorm:
Allow the cursor to seek either forwards or backwards, starting from either the beginning, end, or current offset depending on which location is closest to the desired offset.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -1704,7 +1704,7 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
 
   public UnsafeCursor readUnsafe(UnsafeCursor unsafeCursor) {
     if (unsafeCursor.buffer != null) {
-      throw new IllegalStateException("already attached to a buffe");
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -1704,7 +1704,7 @@ public UnsafeCursor readUnsafe() {
 
   public UnsafeCursor readUnsafe(UnsafeCursor unsafeCursor) {
     if (unsafeCursor.buffer != null) {
-      throw new IllegalStateException("already attached to a buffe");
+      throw new IllegalStateException("already attached to a buffer");
```
swankjesse: thanks!

okio/src/main/java/okio/Buffer.java
```diff
@@ -2015,7 +2038,7 @@ public int seek(long offset) {
      */
     public long resizeBuffer(long newSize) {
       if (buffer == null) {
-        throw new IllegalStateException("not acquired");
+        throw new IllegalStateException("not attached to a buffer");
```
swankjesse: another thanks!

Issue #339, opened 2018-02-01
Include an automatic module name in the manifest


kashike:
related: https://github.com/square/okhttp/pull/3743

kashike:
This has been done in #340.

PR #340, opened 2018-02-02
Add 'Automatic-Module-Name' entry to manifest

kashike:
adds #339

### DIFF
--- a/okio/pom.xml
+++ b/okio/pom.xml
@@ -33,6 +33,17 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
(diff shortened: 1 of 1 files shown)


swankjesse:
I think we want `com.squareup.okio` ?

JakeWharton:
That advice is for those whose package names are already prefixed with
something like com.squareup which this library is not.

kashike:
@JakeWharton https://github.com/square/okhttp/pull/3743#issuecomment-361104656
Package name is `okio`: https://github.com/square/okio/tree/master/okio/src/main/java/okio.
> Your module name should be the same as the package name in that jar file, unless you have a good reason otherwise. ie. you as a human are deciding that one package is the base, and assigning that as the module name. Thus the okhttp3 pom.xml would declare a module name of okhttp3 (I dislike that this is not com.squareup.okhttp3 but you should still match the package name)

JakeWharton:
I don't agree with that, and it's not official policy anyway.

kashike:
Alright, there you go then.

PR #342, opened 2018-02-03
New API, UnsafeCursor.expandBuffer().

swankjesse:
This is intended to be useful when integrating APIs like Cipher that need
a contiguous space to write data into.

### DIFF
--- a/okio/src/main/java/okio/Buffer.java
+++ b/okio/src/main/java/okio/Buffer.java
@@ -2033,14 +2033,14 @@ public final class Buffer implements BufferedSource, BufferedSink, Cloneable {
     }
 
     /**
-     * Change the size of this buffer so that it equals {@code newSize} by either adding new
+     * Change the size of the buffer so that it equals {@code newSize} by either adding new
      * capacity at the end or truncating the buffer at the end. Newly added capacity may span
      * multiple segments.
      *
-     * <p>Warning: it its the caller’s responsibility to write new data to every byte of the
+     * <p>Warning: it is the caller’s responsibility to write new data to every byte of the
      * newly-allocated capacity. Failure to do so may cause serious security problems as the data
-     * in the returned buffers is not zero filled. The buffers may contain dirty pooled segments
-     * that hold very sensitive data from other parts of the current process.
+     * in the returned buffers is not zero filled. Buffers may contain dirty pooled segments that
+     * hold very sensitive data from other parts of the current process.
      *
      * @return the previous size of the buffer.
      */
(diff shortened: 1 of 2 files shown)


okio/src/main/java/okio/Buffer.java
```diff
@@ -2033,14 +2033,14 @@ public int seek(long offset) {
     }
 
     /**
-     * Change the size of this buffer so that it equals {@code newSize} by either adding new
+     * Change the size of the buffer so that it equals {@code newSize} by either adding new
      * capacity at the end or truncating the buffer at the end. Newly added capacity may span
      * multiple segments.
      *
      * <p>Warning: it its the caller’s responsibility to write new data to every byte of the
```
jshvarts: it is the caller's responsibility

okio/src/main/java/okio/Buffer.java
```diff
@@ -2092,6 +2092,48 @@ public long resizeBuffer(long newSize) {
...
+    /**
+     * Grow the buffer by adding a <strong>contiguous range</strong> of capacity in a single
+     * segment. This adds at least {@code minByteCount} bytes but may add up to a full segment of
+     * additional capacity.
+     *
+     * <p>If {@code minByteCount} bytes are available in the buffer's current tail segment that will
+     * be used; otherwise another segment will be allocated and appended. In either case this
+     * returns the number of bytes of capacity added to this buffer.
+     *
+     * <p>Warning: it its the caller’s responsibility to either write new data to every byte of the
```
jshvarts: it is the caller's responsibility

okio/src/main/java/okio/Buffer.java
```diff
@@ -2092,6 +2092,48 @@ public long resizeBuffer(long newSize) {
...
+      if (!readWrite) {
+        throw new IllegalStateException("expandBuffer() only permitted for read/write buffers");
+      }
+
+      Segment tail = buffer.writableSegment(minByteCount);
+      long result = Segment.SIZE - tail.limit;
+      tail.limit = Segment.SIZE;
+      buffer.size += result;
+
+      // Buffer.writableSegment() invalidates our local segment state. Recompute it.
```
bnorm: From what I can tell, the only thing that should need to be updated if the segment is non-null is the cursor end index which we should be able to grab from `segment.limit` again. Otherwise if the segment is null, and the offset is 0, we should just grab `buffer.head` and set values from that. Am I missing something? I see this in `resizeBuffer()` as well and I'm wondering if the call to `seek()` when expanding can be avoided.

okio/src/main/java/okio/Buffer.java
```diff
@@ -2092,6 +2092,48 @@ public long resizeBuffer(long newSize) {
...
+     * Failure to do so may cause serious security problems as the data in the returned buffers is
+     * not zero filled. Buffers may contain dirty pooled segments that hold very sensitive data from
+     * other parts of the current process.
+     *
+     * @param minByteCount the size of the contiguous capacity. Must be positive and less than the
+     *     capacity size of a single segment (8 KiB).
+     * @return the number of bytes expanded by. Not less than {@code minByteCount}.
+     */
+    public long expandBuffer(int minByteCount) {
+      if (minByteCount <= 0) {
```
cketti: It seems odd to check the lower bound here, but not the upper bound.
`Buffer.writableSegment()` checks both, but doesn't provide a helpful error message.

okio/src/main/java/okio/Buffer.java
swankjesse: fixed. thanks

okio/src/main/java/okio/Buffer.java
swankjesse: fixed!

okio/src/main/java/okio/Buffer.java
swankjesse: I think it can probably do less work here. I’m reluctant to change it ’cause if I do I’ll wanna add more test cases!

But even more, I think what we really want is for both resizeBuffer() and expandBuffer() to move the offset as a side-effect. After growing the buffer I think you’ll always want the offset to be at the first new byte. After shrinking? Probably at the end of the buffer.

okio/src/main/java/okio/Buffer.java
swankjesse: Good call. Fixed.

PR #343, opened 2018-02-04
Just use 'okio' for the automatic module name.

swankjesse:


JakeWharton:
What changed your mind? I prefer the old one.

kashike:
https://github.com/square/okhttp/pull/3743#issuecomment-362888517

JakeWharton:
I don't want to use 'retrofit2' for Retrofit. Matching package names is for people who have qualified package names.

swankjesse:
Discussed offline. Automatic-Module-Names shall be package names!

PR #344, opened 2018-02-05
InflaterSource needs to honor read byte count

bnorm:
Instead of reading the requested byte count, InflaterSource will read
to fill up the unused portion of the tail segment.  Correct this by
taking the minimum amount between the unused segment length and the byte
count.

Add two test cases to validate this behavior and catch any regressions.


swankjesse:
Nice find.

Issue #318, opened 2017-08-31
Buffer.read/write for NIO ByteBuffer


robeden:
I've seen issue #197 regarding NIO integration but would like to request a different type of integration. In my application I have `ByteBuffer`s which feed and are fed by `Buffer`s. I would like to request the addition of `read(ByteBuffer)` and `write(ByteBuffer)` methods.

I don't think the discussion in #197 applies since this doesn't involve a hypothetical custom implementation of ByteBuffer, but rather just uses them in the same way as byte arrays are currently used.

The currently process for integrating the two libraries requires the usage of extra `byte[]`s to copy between. This could be avoided by writing directly to or reading from the underlying `Segment` data arrays. Without such methods, there is no way I can see to avoid the extra copy.

ksperling:
Being able to read and write ByteBuffers was one of the things I was thinking of relating to #197, but also being able to e.g. expose a `MappedByteBuffer` as a `Source` directly, without having to copy into a `Buffer`.

bnorm:
With the addition of `Buffer.UnsafeCursor`, would something [like this](https://gist.github.com/bnorm/34fffd47169fd561eb522910a655a8a7) be of use? Allows for the conversion of a ByteBuffer to a Source or Sink.

robeden:
Thanks, @swankjesse! 👍

PR #346, opened 2018-02-11
Implement ReadableByteChannel and WritableByteChannel.

swankjesse:
I was looking at hooking up Okio to Brotli. Their JNI API wants us
to implement these JNI interfaces. The cost of implementing them is
pretty low, and the upside interop should be convenient.

These new APIs require bytes to be copied. This is just how the
NIO APIs work.

Closes: https://github.com/square/okio/issues/318


swankjesse:
InputStream and OutputStream are concrete classes with lots of methods, so the API complexity of implementing them directly is pretty high.

These Channel interfaces are comparatively slim, and the read/write methods fit with the existing overloads.

Issue #330, opened 2018-01-03
InterceptingSink and InterceptingSource


bnorm:
There seems to be a common pattern used in both GzipSink and HashingSink and another used in GzipSource and HashingSource that gives raw access to the segment bytes to checksum or hashing algorithms.  I needed to use a similar pattern to compute a 2's complement checksum for an IPMI sink but had to create this IpmiChecksumSink in the okio package because of package-private access to the segments.  I then created an abstract `InterceptingSink` which applied the common pattern and had a single abstract method which gave unsafe access to the segment data.

```java
public abstract class InterceptingSink extends ForwardingSink {

  protected InterceptingSink(Sink sink) {
    super(sink);
  }

  @Override public void write(Buffer source, long byteCount) throws IOException {
    checkOffsetAndCount(source.size, 0, byteCount);

    long remaining = byteCount;
    for (Segment head = source.head; remaining > 0; head = head.next) {
      int segmentLength = (int) Math.min(remaining, head.limit - head.pos);
      intercept(head.data, head.pos, segmentLength);
      remaining -= segmentLength;
    }

    // Write those bytes to the sink.
    super.write(source, byteCount);
  }

  protected abstract void intercept(byte[] data, int pos, int len) throws IOException;
}
```

Is this idea of an InterceptingSink or InterceptingSource something you think would be worth adding to Okio?

swankjesse:
Really good idea. We're already tracking this here, though we haven't yet settled on an API.
https://github.com/square/okio/issues/141

JakeWharton:
Not that we support web socket streaming anymore in OkHttp, but something like this would be useful for their dumb data xor-ing.

bnorm:
I took a stab at implementing this and you can see my results here: https://github.com/square/okio/compare/master...bnorm:bn/intercept

I'm not super happy with how the GzipSource/GzipSink constructors worked out but I couldn't think of a better way.  If you like what you see, let me know and I'll document, write some unit tests, and open a pull-request.  Otherwise I'm more than happy taking comments about improvements.

swankjesse:
Thanks @bnorm. The most difficult challenge we face is figuring out the API to expose here. I think your InterceptingSink / InterceptingSource is one such API, but there’s a few other options to explore.

I’m partial to something that works on arbitrary buffers, so you can use it independent of stream processing. That’s what #141 is trying to get at.

bnorm:
[...]

PR #349, opened 2018-02-12
Sample to show intercepting of source/sink Buffer via UnsafeCursor

bnorm:
Create a sample which demonstrates a basic use for Buffer.UnsafeCursor.
This shows an InterceptingSource and InterceptingSink which both allow
raw access to the bytes being read and written before they are
propagated. These are used to demo a circular cipher.

Fixes: #330

### DIFF
--- a/samples/src/main/java/okio/samples/Interceptors.java
+++ b/samples/src/main/java/okio/samples/Interceptors.java
@@ -1,0 +1,161 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio.samples;
+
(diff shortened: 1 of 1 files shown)


samples/src/main/java/okio/samples/Interceptors.java
```diff
@@ -0,0 +1,161 @@
...
+  public void run() throws Exception {
+    final byte cipher = (byte) (new Random().nextInt(256) - 128);
+    System.out.println("Cipher   : " + cipher);
+
+    Buffer wire = new Buffer();
+
+    // Create a Sink which will intercept and negatively rotate each byte by `cipher`
+    Sink sink = new InterceptingSink(wire) {
+      @Override
+      protected void intercept(byte[] data, int offset, int length) {
```
swankjesse: Dig

swankjesse:
Great example.

bnorm:
What other samples do you want to include for UnsafeCursor or in general? I've got a couple others I could submit - like a CipherSource/Sink pair and ReplicatingSource/Sink pair - but I'm not sure what kinds of samples you are looking for.

cketti:
I'd be grateful for a `CipherSource` implementation. I built one before `UnsafeCursor` was available and don't know when I'll have time to rewrite it myself.

swankjesse:
Here’s a start on that. It turns out to be slightly awkward because of the way a Cipher may produce more output than input, but doesn’t have a mechanism to drain the output without adding input.

https://gist.github.com/swankjesse/d7427c1ff2891ff00c4661bedcb1f424

bnorm:
Here's an example of what I've been working on for another project. It is definitely awkward working with a Cipher but I think I've handled the corner cases correctly.

https://gist.github.com/bnorm/1295b68b7ebe498ba829136b5b70fa1b

PR #355, opened 2018-02-26
Unit tests for parsing hexadecimal numbers.

pola:
I realised there seemed to be no unit tests for the `readHexadecimalUnsignedLong` method in the `Buffer` class. These two tests should therefore vastly improve the test coverage for that method.


swankjesse:
Those tests are here:
https://github.com/square/okio/blob/a7a492335323c92e0db8f9a7c053bdb6ca13698e/okio/src/test/java/okio/BufferedSourceTest.java#L724

pola:
Thank you for pointing that out! I must have missed those...

PR #337, opened 2018-01-30
Null checking and throwing IOException

JohnSolstice:
### Description
- This PR is coming from an okhttp PR that attempted to fix the issue.  I was asked to make this fix here instead.  https://github.com/square/okhttp/pull/3796
- Checking if input or output streams are null and throwing IOException so that okhttp properly catches the error.  This is to avoid an uncommon crash where a Socket's in/output stream is null

```
Fatal Exception: java.lang.IllegalArgumentException: out == null
       at okio.Okio.sink(SourceFile:69)
       at okio.Okio.sink(SourceFile:118)
       at okhttp3.internal.connection.RealConnection.connectSocket(SourceFile:251)
       at okhttp3.internal.connection.RealConnection.connect(SourceFile:158)
       at okhttp3.internal.connection.StreamAllocation.findConnection(SourceFile:256)
       at okhttp3.internal.connection.StreamAllocation.findHealthyConnection(SourceFile:134)
       at okhttp3.internal.connection.StreamAllocation.newStream(SourceFile:113)
       at okhttp3.internal.connection.ConnectInterceptor.intercept(SourceFile:42)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.internal.cache.CacheInterceptor.intercept(SourceFile:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.internal.http.BridgeInterceptor.intercept(SourceFile:93)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RetryAndFollowUpInterceptor.intercept(SourceFile:125)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:147)
       at okhttp3.internal.http.RealInterceptorChain.proceed(SourceFile:121)
       at okhttp3.RealCall.getResponseWithInterceptorChain(SourceFile:200)
       at okhttp3.RealCall$AsyncCall.execute(SourceFile:147)
       at okhttp3.internal.NamedRunnable.run(SourceFile:32)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1113)
[...]

PR #357, opened 2018-02-26
catching exception instead of null checking

JohnSolstice:
I got this PR merged about a month ago https://github.com/square/okio/pull/337
For some crazy reason I'm still getting these crashes when null checking.  I don't understand why at all.  It makes 0 sense.  To resolve this issue once and for all, I'm remove the null checks I added and instead catching the `IllegalArgumentException`s and throwing `IOException`s.

### DIFF
--- a/okio/src/main/java/okio/Okio.java
+++ b/okio/src/main/java/okio/Okio.java
@@ -114,9 +114,13 @@ public final class Okio {
    */
   public static Sink sink(Socket socket) throws IOException {
     if (socket == null) throw new IllegalArgumentException("socket == null");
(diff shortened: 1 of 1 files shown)


JohnSolstice:
Okay, so I feel dumb right now... Turns out our app is using Splunk which, unless disabled, uses MonitoringSocketImpl.  I think that's what is causing the issue here.  I'm not 100% sure, but I'm going to assume it is for now.  Closing this PR, and hopefully you won't hear from me again ;p

swankjesse:
Splunk! Good to know!

PR #358, opened 2018-02-27
Improved test coverage for `GzipSource::updateCrc`.

pola:
This unit test enters the for loop on line 190 in `GzipSource`. According to OpenClover, the for loop is not covered by any test case, as seen in the screen shot below. This contribution tests that edge case.

screenshot_20180227_141142

### DIFF
--- a/okio/src/test/java/okio/GzipSourceTest.java
+++ b/okio/src/test/java/okio/GzipSourceTest.java
@@ -78,6 +78,22 @@ public final class GzipSourceTest {
     assertGzipped(gzipped);
   }
 
+  @Test public void gunzip_skipCRCCalculationForIrrelevantSegments() throws Exception {
+    Buffer gzipped = new Buffer();
+    gzipped.write(gzipHeader);
+    gzipped.write(deflated);
+    gzipped.write(gzipTrailer);
+
+    GzipSource source = new GzipSource(gzipped);
+    Buffer result = new Buffer();
+    source.read(result, 30);
+    result.head.pos = result.head.limit;
+    source.read(result, 100);
+    result.head.pos = 0;
+
+    assertEquals("It's a UNIX system! I know this!", result.readUtf8());
+  }
+
   /**
    * For portability, it is a good idea to export the gzipped bytes and try running gzip.  Ex.
    * {@code echo gzipped | base64 --decode | gzip -l -v}


okio/src/test/java/okio/GzipSourceTest.java
```diff
@@ -78,6 +78,22 @@
...
+  @Test public void gunzip_skipCRCCalculationForIrrelevantSegments() throws Exception {
+    Buffer gzipped = new Buffer();
+    gzipped.write(gzipHeader);
+    gzipped.write(deflated);
+    gzipped.write(gzipTrailer);
+
+    GzipSource source = new GzipSource(gzipped);
+    Buffer result = new Buffer();
+    source.read(result, 30);
+    result.head.pos = result.head.limit;
```
swankjesse: Probably easier to just do nothing here? I think it'll still cover the right thing

pola: I am not sure if I am with you. When I ran all tests in OpenClover, the entire `GzipSource::updateCrc` method was covered except for the part that was inside the `for` loop. If the four lines you referenced are excluded, I think the test will be a duplicate of `GzipSourceTest::gunzip()`.

swankjesse: What about just removing the two lines that update the buffer’s pos?

pola: I think they are both needed. The first `result.head.pos = result.head.limit;` forces the `for` loop to be entered, which covers a previously untested construct. The second `result.head.pos = 0;` rewinds the buffer so that the assertion a few lines further down can read from the `result` buffer object.

swankjesse: Ahh, gotcha. If we wanna make sure the buffer has some segments in it before we start, I’d prefer to do that through the API rather than by manipulating Buffer’s implementation details.

This gets pretty close to your test without accessing non-public members of Buffer:
```
    Buffer gzipped = new Buffer();
    gzipped.write(gzipHeader);
    gzipped.write(deflated);
    gzipped.write(gzipTrailer);

    GzipSource source = new GzipSource(gzipped);
    Buffer result = new Buffer();
    result.writeUtf8(TestUtil.repeat('a', Segment.SIZE));
    source.read(result, 30);
    source.read(result, 100);
    result.skip(Segment.SIZE);

    assertEquals("It's a UNIX system! I know this!", result.readUtf8());
```

swankjesse:
Obsolete.

PR #365, opened 2018-03-28
Add Kotlin extensions for ByteString, Okio, and Utf8.

JakeWharton:


okio-kotlin/src/main/java/okio/ByteString.kt
```diff
@@ -0,0 +1,69 @@
...
+ */
+inline fun ByteBuffer.toByteString(): ByteString = ByteString.of(this)
+
+/**
+ * Returns a [ByteString] containing the [charset]-encoded bytes of this [String].
+ *
+ * @see ByteString.encodeString
+ */
+inline fun String.encodeByteString(charset: Charset = UTF_8): ByteString =
+    ByteString.encodeString(this, charset)
```
NightlyNexus: If the charset is UTF_8, is it worth doing ByteString.encodeUtf8? (or have a separate function for utf8?)

JakeWharton: Both have the same implementation, calling `string.getBytes(charset)`.

JakeWharton: That optimization should live inside of encodeString if we want to make it.

NightlyNexus: Yeah, I guess I meant to ask if it's worth having a String.encodeUtf8ByteString() extension here.

JakeWharton: I don't think so. That actively harms the Kotlin API. I would rather see the optimization encapsulated behind the Java API.

okio-kotlin/src/main/java/okio/ByteString.kt
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Returns a [ByteString] decoded from this base 64 [String].
```
swankjesse: nit: base 64 -> base64

okio-kotlin/src/main/java/okio/Okio.kt
```diff
@@ -0,0 +1,106 @@
...
+
+import java.io.File
+import java.io.InputStream
+import java.io.OutputStream
+import java.net.Socket
+import java.nio.file.OpenOption
+import java.nio.file.Path
+
+/**
+ * Returns a [BufferedSource] from this [Source].
```
swankjesse: Worth a warning that repeated calls are very bad?

swankjesse: I’m mostly anxious about this:

```
val a = source.buffer().readInt()
val b = source.buffer().readInt()
```

okio-kotlin/src/main/java/okio/Okio.kt
```diff
@@ -0,0 +1,106 @@
...
+ * @see Okio.sink
+ */
+inline fun File.sink(): Sink = Okio.sink(this)
+
+/**
+ * Returns a [Sink] that appends to this [File].
+ *
+ * @see Okio.appendingSink
+ */
+inline fun File.appendingSink(): Sink = Okio.appendingSink(this)
```
swankjesse: could also be `sink(append = true)` ?

swankjesse: it’s less desirable in Java because there’s no named parameters

okio-kotlin/src/test/java/okio/OkioTest.kt
```diff
@@ -0,0 +1,150 @@
...
+  @Test fun pathSource() {
+    val file = temp.newFile()
+    file.writeText("a")
+    val source = file.toPath().source()
+    val buffer = Buffer()
+    source.read(buffer, 1)
+    assertThat(buffer.readUtf8()).isEqualTo("a")
+  }
+
+  @Ignore("Not sure how to test this")
```
swankjesse: interesting

okio-kotlin/src/main/java/okio/Okio.kt
JakeWharton: Yaaaaaas

JakeWharton: Done

okio-kotlin/src/main/java/okio/Okio.kt
JakeWharton: Done

okio-kotlin/src/main/java/okio/ByteString.kt
JakeWharton: DOne

okio-kotlin/src/main/java/okio/Okio.kt
```diff
@@ -0,0 +1,112 @@
...
+ * holds buffered data that will be lost if unconsumed.
+ *
+ * @see Okio.buffer
+ */
+inline fun Source.buffer(): BufferedSource = Okio.buffer(this)
+
+/**
+ * Returns a [BufferedSink] from this [Sink].
+ *
+ * Note: Do not invoke this method multiple times without first calling [BufferedSink.emit] or
```
swankjesse: these comments are perfect

PR #366, opened 2018-03-29
Add gzip and deflate extensions.

JakeWharton:


okio-kotlin/src/test/java/okio/DeflaterSinkTest.kt
```diff
@@ -0,0 +1,37 @@
...
+
+class DeflaterSinkTest {
+  @Test fun deflate() {
+    val data = Buffer()
+    val deflater = (data as Sink).deflate()
+    Okio.buffer(deflater).writeUtf8("Hi!").close()
+    assertThat(data.readByteString().hex()).isEqualTo("789cf3c854040001ce00d3")
+  }
+
+  @Test fun deflateWithDefalter() {
```
swankjesse: Nit: Defalter

okio-kotlin/src/test/java/okio/GzipSourceTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+package okio
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+
+class GzipSourceTest {
+  @Test fun gzip() {
+    val buffer = Buffer().write(ByteString.decodeHex("1f8b0800000000000000f3c8540400dac59e7903000000"))
+    val gzip = (buffer as Source).gzip()
```
swankjesse: The cast is interesting here.

JakeWharton: Yep. Same problem as `Okio.buffer(new Buffer())` would give you in Java.

PR #367, opened 2018-03-29
512-bit hashing source & Kotlin extensions for hashing

JakeWharton:


okio-kotlin/src/main/java/okio/HashingSink.kt
```diff
@@ -0,0 +1,71 @@
...
+ */
+inline fun Sink.hashSha512(): HashingSink = HashingSink.sha512(this)
+
+/**
+ * Returns a [HashingSink] that uses the obsolete SHA-1 HMAC algorithm with [key] to produce 160-bit
+ * hashes.
+ *
+ * @see HashingSink.hmacSha1
+ */
+inline fun Sink.hashHmacSha1(key: ByteString): HashingSink = HashingSink.hmacSha1(this, key)
```
JakeWharton: Should `key` become a nullable `hmacKey` parameter on `hashSha1()` with a default of `null`? Then it's one function that switches between two impls rather than two functions.

swankjesse: I think the use cases are pretty different.

swankjesse: Also: omit the word hash here? It's like ATM machine or ABS brakes.

okio/src/main/java/okio/HashingSource.java
```diff
@@ -57,6 +57,11 @@ public static HashingSource sha256(Source source) {
     return new HashingSource(source, "SHA-256");
   }
 
+  /** Returns a sink that uses the SHA-512 hash algorithm to produce 512-bit hashes. */
+  public static HashingSource sha512(Source source) {
```
swankjesse: Thanks!

okio-kotlin/src/main/java/okio/HashingSink.kt
JakeWharton: Just for the hmac ones or for all of them?

The names are interesting for these overloads because the return type is significant here. You have to retain the `HashinkSink`/`HashingSource` in order to grab the hash. Should these names be `hashingSha1()`? What does that do to the hmac names then?

okio/src/main/java/okio/HashingSource.java
JakeWharton: Merged this commit to master ahead of the extensions.

okio-kotlin/src/main/java/okio/HashingSink.kt
swankjesse: sha1Sink? Tough call. I don’t think these things get much use.

swankjesse: I think we want:
sha1() –> returns ByteString
hmacSha1(ByteString) -> returns ByteString
sha1Sink() –> returns HashingSink

JakeWharton: The only thing I don't like about that is that it creates a disparity for things like `.gzip()`.

swankjesse:
Obsolete!

PR #369, opened 2018-04-16
Build Okio with Gradle

swankjesse:


benchmarks/build.gradle
```diff
@@ -0,0 +1,15 @@
+apply plugin: 'org.jetbrains.kotlin.platform.jvm'
+
+apply plugin: "me.champeau.gradle.jmh"
+
+sourceCompatibility = JavaVersion.VERSION_1_7
+targetCompatibility = JavaVersion.VERSION_1_7
```
JakeWharton: Could be 8 if we want...

build.gradle
```diff
@@ -0,0 +1,49 @@
...
+
+  dependencies {
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
+    classpath "me.champeau.gradle:jmh-gradle-plugin:0.4.5"
+  }
+
+  repositories {
+    mavenCentral()
+    jcenter()
+    maven { url "https://plugins.gradle.org/m2/" }
```
JakeWharton: there's `gradlePluginPortal()` now!

okio-kotlin/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'org.jetbrains.kotlin.platform.jvm'
+
+sourceCompatibility = JavaVersion.VERSION_1_6
+targetCompatibility = JavaVersion.VERSION_1_6
+
+dependencies {
+  implementation project(':okio')
```
JakeWharton: needs to be `api`

okio-kotlin/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'org.jetbrains.kotlin.platform.jvm'
+
+sourceCompatibility = JavaVersion.VERSION_1_6
```
digitalbuddha: some modules are 1.6, while others are 1.7 is that intentional?

okio/build.gradle
```diff
@@ -0,0 +1,12 @@
+apply plugin: 'org.jetbrains.kotlin.platform.jvm'
+
+sourceCompatibility = JavaVersion.VERSION_1_7
+targetCompatibility = JavaVersion.VERSION_1_7
+
+dependencies {
+  implementation deps.kotlin.stdLib.jre7
```
NightlyNexus: ditto Mike's comment: jre6?

Egorand: We should probably move everything to jre7 (which is actually jdk7 now)

JakeWharton: do we use anything from 7? i wouldn't bother unless we do

Egorand: Diamond operator, underscores in numeric literals

JakeWharton: That is not tied to this dependency

naturalwarren:
Out of curiosity, what's the motivation behind this change?

swankjesse:
@naturalwarren https://github.com/square/okio/issues/370

build.gradle
```diff
@@ -0,0 +1,49 @@
...
+      'jsr305': '3.0.2',
+      'animalSniffer': '1.10',
+      'junit': '4.12',
+      'assertj': '1.7.0'
+  ]
+
+  ext.deps = [
+      'kotlin': [
+          'stdLib': [
+              'common': "org.jetbrains.kotlin:kotlin-stdlib-common:${versions.kotlin}",
```
tbroyer: You don't need to specify the version, it's inherited from the plugin: https://kotlinlang.org/docs/reference/using-gradle.html#configuring-dependencies
> Starting with Kotlin 1.1.2, the dependencies with group `org.jetbrains.kotlin` are by default resolved with the version taken from the applied plugin.

benchmarks/build.gradle
```diff
@@ -0,0 +1,15 @@
+apply plugin: 'org.jetbrains.kotlin.platform.jvm'
```
tbroyer: Shouldn't this just be `org.jetbrains.kotlin.jvm`? (without the `platform` part)

Egorand: You're right, it should be for this module

build.gradle
```diff
@@ -0,0 +1,49 @@
...
+          ],
+          'test': [
+              'common': "org.jetbrains.kotlin:kotlin-test-common:${versions.kotlin}",
+              'annotations': "org.jetbrains.kotlin:kotlin-test-annotations-common:${versions.kotlin}",
+              'jre': "org.jetbrains.kotlin:kotlin-test-junit:${versions.kotlin}",
+          ],
+      ],
+      'jmh': "org.openjdk.jmh:jmh-core:${versions.jmh}",
+      'jsr305': "com.google.code.findbugs:jsr305:${versions.jsr305}",
+      'animalSniffer': "org.codehaus.mojo:animal-sniffer-annotations:${versions.animalSniffer}",
```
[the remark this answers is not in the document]
JakeWharton: I think this would be a great PR for a follow-up

build.gradle
```diff
@@ -0,0 +1,49 @@
...
+      'test': [
+          'junit': "junit:junit:${versions.junit}",
+          'assertj': "org.assertj:assertj-core:${versions.assertj}"
+      ]
+  ]
+
+  dependencies {
+    classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:${versions.kotlin}"
```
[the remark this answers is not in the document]
JakeWharton: Is there any advantage to this? It's weirder than Gradle normal level of weird.

build.gradle
```diff
@@ -0,0 +1,49 @@
...
+      'animalSniffer': '1.10',
+      'junit': '4.12',
+      'assertj': '1.7.0'
+  ]
+
+  ext.deps = [
+      'kotlin': [
+          'stdLib': [
+              'common': "org.jetbrains.kotlin:kotlin-stdlib-common:${versions.kotlin}",
+              'jre6': "org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlin}",
```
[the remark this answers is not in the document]
JakeWharton: There will eventually be 7 and 8 versions to support stdlib-jdk7 and stdlib-jdk8

.travis.yml
```diff
@@ -8,6 +8,9 @@ addons:
     packages:
       - oracle-java8-installer # Updates JDK 8 to the latest available.
 
+script:
+  - ./gradlew clean build jmh
```
JakeWharton: Does this run jmh or just build it? We shouldn't need to run it.

build.gradle
```diff
@@ -1,44 +1,59 @@
+group = 'com.squareup.okio'
+version = '1.15.0-SNAPSHOT'
```
JakeWharton: I think both of these need to be in `allprojects { }` block.

okio/build.gradle
```diff
@@ -1,12 +1,19 @@
+group = 'com.squareup.okio'
```
JakeWharton: With the `allprojects { }` block you can get rid of this from the subproject.

Egorand:
Alright, that should be it except for publishing setup, I'll do this as a follow-up, then we can get rid of Maven files

okio/build.gradle
```diff
@@ -0,0 +1,29 @@
+description = 'Okio'
+
+apply plugin: 'java-library'
+apply plugin: 'osgi'
```
JakeWharton: Oh dear...

Egorand: Yeah, that's used for `Automatic-Module-Name`, I'll if there's a better way

tbroyer: You don't actually need it to configure the jar manifest: https://docs.gradle.org/current/userguide/java_plugin.html#sub:manifest

PR #372, opened 2018-04-21
Configure 'gradle-mvn-push'

### LINKED ISSUE #371: Migrate Maven publishing configuration to Gradle
Follow up to #369, we should be able to publish artifacts with Gradle

Egorand:
Closes #371


build.gradle
```diff
@@ -55,12 +55,14 @@ buildscript {
...
+  group = GROUP
+  version = VERSION_NAME
 }
 
 subprojects {
   repositories {
     mavenCentral()
   }
+
+  apply from: "$rootDir/gradle/gradle-mvn-push.gradle"
```
JakeWharton: I'm not sure we should do this as it causes the bechmarks, samples, and root project to be pushed. Maven sort of forced that on us, but there's no reason to push those things with Gradle.

Egorand:
PTAL @JakeWharton

PR #375, opened 2018-04-21
Port Okio type's public API to Kotlin.

JakeWharton:


okio/src/main/java/okio/OldOkio.java
```diff
@@ -0,0 +1,177 @@
...
+import java.io.OutputStream;
+import java.net.Socket;
+import java.net.SocketTimeoutException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import javax.annotation.Nullable;
+
+import static okio.Util.checkOffsetAndCount;
+
+final class OldOkio {
```
JakeWharton: This isn't a new file. It's the dying echoes of `Okio.java`. Since it's all implementation detail, it can be ported to Kotlin separately.

okio/src/main/java/okio/Okio.kt
```diff
@@ -0,0 +1,109 @@
...
+fun Socket.source(): Source {
+  val timeout = SocketAsyncTimeout(this)
+  val source = InputStreamSource(getInputStream(), timeout)
+  return timeout.source(source)
+}
+
+/** Returns a sink that writes to `file`. */
+@JvmOverloads
+@Throws(FileNotFoundException::class)
+fun File.sink(append: Boolean = false): Sink = FileOutputStream(this, append).sink()
```
JakeWharton: There's a new API being exposed to Java callers hidden here: `Okio.sink(File, boolean)`. I didn't deprecate `Okio.appendingSink` so it currently serves as an overload of functionality.

We could deprecate `appendingSink` but it reduces the API quality for Java consumers. We could remove the `append` parameter and overloads from this function but it reduces the API quality for Kotlin consumers.

Egorand: I wish there was a way to hide API from Java callers. `@JvmName("_sink")` is super hacky, but it will at least hide the method from autocomplete.

JakeWharton: And it's even more complicated here, because we only want to hide one of the overloads. No good solution that I can see.

PR #376, opened 2018-04-21
Port Utf8 type's public API to Kotlin.

JakeWharton:


okio/src/main/java/okio/Utf8.kt
```diff
@@ -0,0 +1,115 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Okio assumes most applications use UTF-8 exclusively, and offers optimized implementations of
+ * common operations on UTF-8 strings.
+ *
+ * <table border="1" cellspacing="0" cellpadding="3" summary="">
```
Egorand: We should be able to do this in Markdown

JakeWharton: Yeah. Neither Markdown nor HTML tables work currently: https://youtrack.jetbrains.com/issue/KT-16228

Egorand: :doh:

PR #379, opened 2018-04-21
Port ForwardingSource/Sink to Kotlin.

JakeWharton:


okio/src/main/java/okio/ForwardingSink.kt
```diff
@@ -0,0 +1,40 @@
...
+
+import java.io.IOException
+
+/** A [Sink] which forwards calls to another. Useful for subclassing. */
+abstract class ForwardingSink(
+  /** [Sink] to which this instance is delegating. */
+  @get:JvmName("delegate")
+  val delegate: Sink
+) : Sink {
+  // TODO 'Sink by delegate' once https://youtrack.jetbrains.com/issue/KT-23935 is fixed.
```
Egorand: 👍

swankjesse: API checks already paying for themselves!

PR #382, opened 2018-04-21
Move benchmarks under the main module

Egorand:


JakeWharton:
Are these still compiled automatically?

Egorand:
Yes, `:okio:jmh` should build and run the tests

JakeWharton:
I mean as part of a task that's run on CI. I would want these hooked up to `assemble` to ensure they're being compiled as part of `build`.

PR #383, opened 2018-04-21
Port InflateSource/DeflateSink to Kotlin.

JakeWharton:


okio/src/main/java/okio/InflaterSource.kt
```diff
@@ -0,0 +1,132 @@
...
+import java.io.IOException
+import java.util.zip.DataFormatException
+import java.util.zip.Inflater
+
+/**
+ * A source that uses [DEFLATE](http://tools.ietf.org/html/rfc1951)
+ * to decompress data read from another source.
+ */
+class InflaterSource
+/**
```
swankjesse: I’m curious if KDoc is gonna get this right

PR #385, opened 2018-04-21
Add Okio recipes to the readme

swankjesse:


samples/src/main/java/okio/samples/ReadFileLineByLine.java
```diff
@@ -0,0 +1,48 @@
...
+import okio.Source;
+
+public final class ReadFileLineByLine {
+  public void run() throws Exception {
+    readLines(new File("README.md"));
+  }
+
+  public void readLines(File file) throws IOException {
+    try (Source fileSource = Okio.source(file);
+         BufferedSource bufferedFileSource = Okio.buffer(fileSource)) {
```
JakeWharton: Any reason to keep both here instead of just doing `Okio.buffer(Okio.source(file))`?

swankjesse: Just wanted to expose both types for the first example.

PR #389, opened 2018-04-23
Convert BufferedSink and BufferedSource interfaces to Kotlin.

JakeWharton:


okio/src/main/java/okio/BufferedSink.kt
```diff
@@ -0,0 +1,358 @@
...
+   * assertEquals((byte) 0x00, buffer.readByte());
+   * assertEquals((byte) 0x00, buffer.readByte());
+   * assertEquals((byte) 0x00, buffer.readByte());
+   * assertEquals((byte) 0x00, buffer.readByte());
+   * assertEquals((byte) 0x00, buffer.readByte());
+   * assertEquals(0, buffer.size());
+   * ```
+   */
+  @Throws(IOException::class)
+  fun writeLongLe(v: Long): BufferedSink
```
Egorand: nit: `v` for `value`? why not `l`?

JakeWharton: I guess.

I think we'll want to make a pass over all the parameter names before releasing since they're technically public API now.

okio/src/main/java/okio/BufferedSink.kt
```diff
@@ -0,0 +1,358 @@
...
+   */
+  @Throws(IOException::class)
+  fun write(source: ByteArray): BufferedSink
+
+  /**
+   * Like [OutputStream.write], this writes `byteCount`
+   * bytes of `source`, starting at `offset`.
+   */
+  @Throws(IOException::class)
+  fun write(source: ByteArray, offset: Int, byteCount: Int): BufferedSink
```
swankjesse: for follow-up, I’d love to give `offset` and `byteCount` defaults and combine this with the preceding method

swankjesse: (and we’d default `byteCount` to `source.length - offset` !)

okio/src/main/java/okio/BufferedSink.kt
```diff
@@ -0,0 +1,358 @@
...
+   * buffer.writeByte(' ');
+   * buffer.writeUtf8("That's what I said: you're a nerd.\n", 29, 33);
+   * buffer.writeByte(' ');
+   * buffer.writeUtf8("I prefer to be called a hacker!\n", 24, 31);
+   *
+   * assertEquals("hacker nerd hacker!", buffer.readUtf8());
+   * ```
+   */
+  @Throws(IOException::class)
+  fun writeUtf8(string: String, beginIndex: Int, endIndex: Int): BufferedSink
```
swankjesse: same

okio/src/main/java/okio/BufferedSink.kt
```diff
@@ -0,0 +1,358 @@
...
+  /** Encodes `string` in `charset` and writes it to this sink. */
+  @Throws(IOException::class)
+  fun writeString(string: String, charset: Charset): BufferedSink
+
+  /**
+   * Encodes the characters at `beginIndex` up to `endIndex` from `string` in
+   * `charset` and writes it to this sink.
+   */
+  @Throws(IOException::class)
+  fun writeString(string: String, beginIndex: Int, endIndex: Int, charset: Charset): BufferedSink
```
swankjesse: same

okio/src/main/java/okio/BufferedSink.kt
swankjesse: Thumbs down on `l`, which looks like `1` and `I`

okio/src/main/java/okio/BufferedSource.kt
```diff
@@ -0,0 +1,574 @@
...
+   * Buffer buffer = new Buffer();
+   * buffer.writeUtf8("Don't move! He can't see us if we don't move.");
+   *
+   * byte m = 'm';
+   * assertEquals(6,  buffer.indexOf(m));
+   * assertEquals(40, buffer.indexOf(m, 12));
+   * ```
+   */
+  @Throws(IOException::class)
+  fun indexOf(b: Byte, fromIndex: Long): Long
```
swankjesse: fromIndex should be defaulted?

okio/src/main/java/okio/BufferedSource.kt
```diff
@@ -0,0 +1,574 @@
...
+  /**
+   * Returns the index of `b` if it is found in the range of `fromIndex` inclusive
+   * to `toIndex` exclusive. If `b` isn't found, or if `fromIndex == toIndex`,
+   * then -1 is returned.
+   *
+   * The scan terminates at either `toIndex` or the end of the buffer, whichever comes
+   * first. The maximum number of bytes scanned is `toIndex-fromIndex`.
+   */
+  @Throws(IOException::class)
+  fun indexOf(b: Byte, fromIndex: Long, toIndex: Long): Long
```
swankjesse: indexes should be defaulted?

okio/src/main/java/okio/BufferedSource.kt
```diff
@@ -0,0 +1,574 @@
...
+   *
+   * Buffer buffer = new Buffer();
+   * buffer.writeUtf8("Don't move! He can't see us if we don't move.");
+   *
+   * assertEquals(6,  buffer.indexOf(MOVE));
+   * assertEquals(40, buffer.indexOf(MOVE, 12));
+   * ```
+   */
+  @Throws(IOException::class)
+  fun indexOf(bytes: ByteString, fromIndex: Long): Long
```
swankjesse: indexes should be defaulted?

okio/src/main/java/okio/BufferedSource.kt
```diff
@@ -0,0 +1,574 @@
...
+   *
+   * Buffer buffer = new Buffer();
+   * buffer.writeUtf8("Dr. Alan Grant");
+   *
+   * assertEquals(4,  buffer.indexOfElement(ANY_VOWEL));    // 'A' in 'Alan'.
+   * assertEquals(11, buffer.indexOfElement(ANY_VOWEL, 9)); // 'a' in 'Grant'.
+   * ```
+   */
+  @Throws(IOException::class)
+  fun indexOfElement(targetBytes: ByteString, fromIndex: Long): Long
```
swankjesse: indexes should be defaulted?

okio/src/main/java/okio/BufferedSource.kt
```diff
@@ -0,0 +1,574 @@
...
+  @Throws(IOException::class)
+  fun rangeEquals(offset: Long, bytes: ByteString): Boolean
+
+  /**
+   * Returns true if `byteCount` bytes at `offset` in this source equal `bytes`
+   * at `bytesOffset`. This expands the buffer as necessary until a byte does not match, all
+   * bytes are matched, or if the stream is exhausted before enough bytes could determine a match.
+   */
+  @Throws(IOException::class)
+  fun rangeEquals(offset: Long, bytes: ByteString, bytesOffset: Int, byteCount: Int): Boolean
```
swankjesse: more defaults?

JakeWharton:
Unfortunately I'm pretty sure we can't do any defaulting because this is an interface where `@JvmOverloads` doesn't work.

PR #397, opened 2018-04-25
Print JApicmp output on build failure

### LINKED ISSUE #392: JApicmp should print to console when it fails
Makes it easier to see failure on CI

Egorand:
Closes #392 

This will `cat` even if the failure wasn't caused by Japicmp, maybe there's a cleaner solution

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ script:
 after_success:
   - .buildscript/deploy_snapshot.sh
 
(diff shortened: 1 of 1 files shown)


JakeWharton:
The plugin can't do this?

Egorand:
Haven't found config options for that, I'll open an issue to see if they can add it

PR #399, opened 2018-04-25
Convert Base64 to Kotlin

swankjesse:


okio/src/test/java/okio/DeflateKotlinTest.kt
```diff
@@ -16,6 +16,7 @@
 
 package okio
 
+import okio.ByteString.Companion.decodeHex
```
swankjesse: This doesn’t seem right. Any advice? It yields the right Java API.

okio-kotlin/src/main/java/okio/ByteString.kt
```diff
@@ -46,20 +46,6 @@ inline fun ByteBuffer.toByteString(): ByteString = ByteString.of(this)
...
- * @see ByteString.decodeBase64
- */
-inline fun String.decodeBase64(): ByteString? = ByteString.decodeBase64(this)
-
-/**
- * Returns a [ByteString] decoded from this hex [String].
- *
- * @see ByteString.decodeHex
- */
-inline fun String.decodeHex(): ByteString = ByteString.decodeHex(this)
```
JakeWharton: Delete/move the tests specific to this?

okio/src/test/java/okio/DeflateKotlinTest.kt
swankjesse: screen shot 2018-04-25 at 3 27 13 pm

swankjesse: not yet

JakeWharton: Ah, yeah. We went through this on KotlinPoet. Sigh.

okio-kotlin/src/main/java/okio/ByteString.kt
swankjesse: Done.

PR #400, opened 2018-04-26
Networking recipe.

swankjesse:
I chose a real network protocol here, which makes the example more complex
than I'd prefer, but it's also more representative of what real code would
need to do.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,22 +363,22 @@ around these traps:
    `writeByte()` and `writeShort()`. You can pass an “unsigned” byte like 255
    and Okio will do the right thing.
 
-|           Value | Method       | Width | Endianness | Encoded Bytes             |
-| --------------: | :----------- | ----: | :--------- | :------------------------ |
-|               3 | writeByte    |     1 |            | `03`                      |
-|               3 | writeShort   |     2 | big        | `00 03`                   |
-|               3 | writeInt     |     4 | big        | `00 00 00 03`             |
-|               3 | writeLong    |     8 | big        | `00 00 00 00 00 00 00 03` |
-|               3 | writeShortLe |     2 | little     | `03 00`                   |
-|               3 | writeIntLe   |     4 | little     | `03 00 00 00`             |
-|               3 | writeLongLe  |     8 | little     | `03 00 00 00 00 00 00 00` |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -363,22 +363,22 @@ around these traps:
...
-|               3 | writeIntLe   |     4 | little     | `03 00 00 00`             |
-|               3 | writeLongLe  |     8 | little     | `03 00 00 00 00 00 00 00` |
-|  Byte.MAX_VALUE | writeByte    |     1 |            | `7f`                      |
-| Short.MAX_VALUE | writeShort   |     2 | big        | `7f ff`                   |
-|   Int.MAX_VALUE | writeInt     |     4 | big        | `7f ff ff ff`             |
-|  Long.MAX_VALUE | writeLong    |     8 | big        | `7f ff ff ff ff ff ff ff` |
-| Short.MAX_VALUE | writeShortLe |     2 | little     | `ff 7f`                   |
-|   Int.MAX_VALUE | writeIntLe   |     4 | little     | `ff ff ff 7f`             |
-|  Long.MAX_VALUE | writeLongLe  |     8 | little     | `ff ff ff ff ff ff ff 7f` |
+| Method       | Width | Endianness |           Value | Encoded Value             |
```
swankjesse: Just moved the value column over

README.md
```diff
@@ -444,6 +444,87 @@ Encoding other binary formats is usually quite similar. Some tips:
...
+
+Sinks buffer outbound data to minimize I/O operations. This is efficient but it
+means you must manually call `flush()` to transmit data. Typically
+message-oriented protocols flush after each message. Note that Okio will
+automatically flush when the buffered data exceeds some threshold. This is
+intended to save memory and you shouldn’t rely on it for interactive protocols.
+
+Okio builds on `java.io.Socket` for connectivity. Create your socket as a server
+or as a client, then use `Okio.source(Socket)` to read and `Okio.sink(Socket)`
+to write. These APIs also work with `SSLSocket`. You should use SSL unless you
```
cketti: Are you using "SSL" to avoid having to explain why it's `SSLSocket` but the protocol is called TLS?

swankjesse: Yuppers. They're synonyms.

PR #401, opened 2018-04-26
Multiplatform setup POC

Egorand:


common/src/main/kotlin/okio/System.kt
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package okio
+
+expect fun arraycopy(src: ByteArray, srcPos: Int, dest: ByteArray, destPos: Int, length: Int)
```
swankjesse: internal ?

Egorand: Yeah, I wonder if it's gonna work as expected, since technically those are different modules

JakeWharton: Not a bug. You can't cross modules with `internal`.

Egorand: Makes sense, hence we should leave it public and ensure it's not exposed as a transitive dependency

JakeWharton: Why are we adding a transitive dependency instead of making the okio module multiplatform?

Egorand: Good point, I guess a better setup would be `okio-common` and `okio` itself being the JVM module

JakeWharton: I've been using nested projects for multiplatform if that interests you:
```
okio/
  js/
  jvm/
  src/
```

PR #407, opened 2018-04-27
Prepare tests to move Buffer to Kotlin

swankjesse:
Our tests access package-private members of Buffer, Segment, and SegmentPool. This
changes the tests to use members of TestUtil instead. This migrates TestUtil to
Kotlin so that it can access those members.


.gitignore
```diff
@@ -11,7 +11,6 @@ out
 lib
 reports
 
-target
 pom.xml.*
 release.properties
```
JakeWharton: All three can be nuked

okio/jvm/src/test/java/okio/BufferCursorKotlinTest.kt
```diff
@@ -0,0 +1,113 @@
...
+  @Parameter lateinit var bufferFactory: BufferCursorTest.BufferFactory
+
+  @Test fun acquireReadOnlyDoesNotCopySharedDataArray() {
+    val buffer = deepCopy(bufferFactory.newBuffer())
+    assumeTrue(buffer.size() > 0)
+
+    val shared = buffer.clone()
+    assertTrue(buffer.head!!.shared)
+
+    buffer.readUnsafe().use({ cursor ->
```
Egorand: supernit: `use {}`, without `()`

okio/jvm/src/test/java/okio/BufferCursorKotlinTest.kt
```diff
@@ -0,0 +1,113 @@
+package okio
```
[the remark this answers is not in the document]
swankjesse: Fixed!

.gitignore
swankjesse: Fixed.

okio/jvm/src/test/java/okio/BufferCursorKotlinTest.kt
swankjesse: Fixed all.

PR #409, opened 2018-04-28
Port SegmentSharingTest to Kotlin.

JakeWharton:


JakeWharton:
Sorry @swankjesse this conflicts with your PR. I'm a bit hesitant to change both a runtime class and a test class at the same time (when it's not a trivial change) for worry of behavior skew so I did each individually here (so no squashing!).

swankjesse:
Nice.

PR #411, opened 2018-04-28
Move Util into the common module

Egorand:


swankjesse:
Weird:
```
   > Detected binary changes between jvm-1.15.0-SNAPSHOT.jar and okio-1.14.0.jar. See failure report at file:///home/travis/build/square/okio/okio/jvm/build/reports/japi.txt
```

okio/src/main/kotlin/okio/Util.kt
```diff
@@ -0,0 +1,65 @@
...
+
+package okio
+
+internal fun checkOffsetAndCount(size: Long, offset: Long, byteCount: Long) {
+  if (offset or byteCount < 0 || offset > size || size - offset < byteCount) {
+    throw ArrayIndexOutOfBoundsException("size=$size offset=$offset byteCount=$byteCount")
+  }
+}
+
+@JvmName("reverseBytesShort")
```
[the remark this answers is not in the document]
Egorand: But yeah, Japicmp actually marks them

Egorand:
I suggest we stall this until all other non-API breaking changes are made, and then include it in the API-breaking pass prior to 2.0 release.

okio/src/main/kotlin/okio/Util.kt
```diff
@@ -0,0 +1,63 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:JvmName("Util")
+
+package okio
+
+internal fun checkOffsetAndCount(size: Long, offset: Long, byteCount: Long) {
```
swankjesse: Does defining these in a `Util` object avoid the japicmp problem? I’d love to get rid of the last `.java` files here.

Egorand: It doesn't, you'll still have a `public class Util`

swankjesse:
@Egorand wanna make these `internal` in `Okio.kt` ? That might solve our visibility problems.

Egorand:
Yeah, I'll try that. Japicmp will still fail since I've renamed the `reverseBytes` methods

JakeWharton:
There's surely a way to add things to ignore. Maybe we should start using
that.

Egorand:
Totally

Egorand:
Actually excluding `okio.Util` solves all issues, and this feels like the right thing to do - this is mergeable now

okio/src/main/kotlin/okio/Util.kt
```diff
@@ -0,0 +1,63 @@
...
+
+internal fun Short.reverseBytes(): Short {
+  val i = toInt() and 0xffff
+  val reversed = (i and 0xff00 ushr 8) or
+                 (i and 0x00ff  shl 8)
+  return reversed.toShort()
+}
+
+internal fun Int.reverseBytes(): Int {
+  return (this and -0x1000000 ushr 24) or
```
swankjesse: the -0x10000000 is sad. Not much we can do though

swankjesse: (negative hex values are bad)

swankjesse: https://youtrack.jetbrains.com/issue/KT-4749

PR #412, opened 2018-04-28
Migrate Buffer to Kotlin

swankjesse:
This first pass is intended to get things converted over and correct. We need to take
another pass in follow-up to replace overloaded methods with JvmOverloads and default
parameters. Also to consider using vals where we currently use funs, and other
Kotlin niceties.


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -0,0 +1,2176 @@
...
+
+      val data = segment!!.data
+      var pos = segment.pos
+      val limit = segment.limit
+
+      while (pos < limit) {
+        val digit: Int
+
+        val b = data[pos]
+        if (b >= '0'.toByte() && b <= '9'.toByte()) {
```
swankjesse: for follow-up, these should be ranges

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -0,0 +1,2176 @@
...
+  }
+
+  override fun writeHexadecimalUnsignedLong(v: Long): Buffer {
+    var v = v
+    if (v == 0L) {
+      // Both a shortcut and required since the following code can't handle zero.
+      return writeByte('0'.toInt())
+    }
+
+    val width = java.lang.Long.numberOfTrailingZeros(java.lang.Long.highestOneBit(v)) / 4 + 1
```
swankjesse: Obvious candidate for platform.kt

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -0,0 +1,2176 @@
...
+  /** Returns the 128-bit MD5 hash of this buffer.  */
+  @Suppress("NON_FINAL_MEMBER_IN_FINAL_CLASS") // Required to keep JApicmp happy.
+  open fun md5() = digest("MD5")
+
+  /** Returns the 160-bit SHA-1 hash of this buffer.  */
+  @Suppress("NON_FINAL_MEMBER_IN_FINAL_CLASS") // Required to keep JApicmp happy.
+  open fun sha1() = digest("SHA-1")
+
+  /** Returns the 256-bit SHA-256 hash of this buffer.  */
+  @Suppress("NON_FINAL_MEMBER_IN_FINAL_CLASS") // Required to keep JApicmp happy.
```
swankjesse: I’m tempted to cut an intermediate 1.x release that makes all of these final, just so we have that as our baseline

JakeWharton: Alternatively, once we're ready for 2.0 we disable japicmp, fix all these, release, and make 2.0 the baseline.

brettwooldridge: Just a fly on the wall here, but @JakeWharton 's suggestion sounds reasonable to this observer.

JakeWharton:
Had a bunch of nits, but we'll need to do a pass over everything anyway so whevs. Forward progress is more important.

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: I ran into a wall trying to get the `size()` method on Options to work. I think we quietly release Okio 1.14.1 which just adds a bunch of redundant final modifiers.

https://github.com/square/okio/pull/417/files

PR #413, opened 2018-04-29
Move Segment and SegmentPool to common.

JakeWharton:


okio/src/main/kotlin/okio/SegmentPool.kt
```diff
@@ -34,7 +34,7 @@ internal object SegmentPool {
 
   @JvmStatic
   fun take(): Segment {
-    synchronized(SegmentPool::class.java) {
+    synchronized(this) {
```
swankjesse: The bytecode here uses the singleton instance, right? I think that's fine. Alternative would be ummm, `SingletonPool::class`, but I think that also does some extra work.

okio/jvm/src/main/java/okio/platform.kt
```diff
@@ -17,6 +17,8 @@
 package okio
 
 actual typealias JvmOverloads = kotlin.jvm.JvmOverloads
+actual typealias JvmField = kotlin.jvm.JvmField
```
swankjesse: doesn’t work as I was hoping:
https://youtrack.jetbrains.com/issue/KT-18882#comment=27-2847603

JakeWharton: Damn. Revert?

swankjesse: Something. I don’t wanna release 2.0 until it’s fixed or reverted.

Egorand: It does work, it's the IDE plugin that is broken. I decompiled `Segment` with JD-GUI and I see that it has `public final` for fields that are marked with `@okio.JvmField` and no getters. If I create the following Java file:
```
public class Test {
  public static void main(String[] args) {
    Segment segment = new Segment();
    segment.getData();
  }
}
```
The IDE won't complain, but compiling it from command line will fail with `cannot find symbol: method getData()`. Changing this to `segment.data` works fine with both IDE and compiler.

swankjesse: Whoa! Thats great news. Weird that the IDE plugin can disagree here.

I'll try to figure out what bug to report.

PR #414, opened 2018-04-29
Move null checks to be more eager on their source.

JakeWharton:


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -2058,19 +2058,19 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
      */
     @Suppress("NON_FINAL_MEMBER_IN_FINAL_CLASS") // Required to keep JApicmp happy.
     open fun resizeBuffer(newSize: Long): Long {
-      check(buffer != null) { "not attached to a buffer" }
+      val buffer = checkNotNull(buffer) { "not attached to a buffer" }
```
swankjesse: Neat

PR #417, opened 2018-04-30
Explicitly declare some members as final in Okio 1.x.

swankjesse:
This will make it easier for 2.x to be compatible according to japicmp.


swankjesse:
That’s my plan!

JakeWharton:
We should file a japicmp bug about this

swankjesse:
https://github.com/siom79/japicmp/issues/206

PR #418, opened 2018-04-30
Migrate Options to Kotlin

swankjesse:
Also switch japicmp to compare against Okio 1.14.1, which adds a bunch of redundant
final modifiers for its benefit


okio/jvm/src/main/java/okio/Options.kt
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import java.util.AbstractList
```
Egorand: Both of these are available in Kotlin as typealiases

okio/jvm/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -882,7 +882,7 @@ private void assertLongDecimalString(String s, long expected) throws IOException
   }
 
   @Test public void select() throws IOException {
-    Options options = Options.of(
+    Options options = Options.Companion.of(
```
Egorand: `Companion` is redundant (this was probably auto-generated, but still)

Egorand:
Javadoc task fails - it can't find Java files! 🎆

okio/jvm/src/main/java/okio/Options.kt
swankjesse: Will fix in follow-up

okio/jvm/src/test/java/okio/BufferedSourceTest.java
swankjesse: Will fix in follow-up.

PR #419, opened 2018-04-30
Add syntactic sugar for bitwise operations

swankjesse:


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -1034,7 +1034,7 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
     val data = tail.data
     var limit = tail.limit
     data[limit++] = (s.ushr(8) and 0xff).toByte()
-    data[limit++] = (s and 0xff).toByte()
+    data[limit++] = (s                  and 0xff).toByte()
```
Egorand: This looks weird

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -1061,19 +1061,20 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
+    data[limit++] = (v ushr 24 and 0xffL).toByte()
+    data[limit++] = (v ushr 16 and 0xffL).toByte()
+    data[limit++] = (v ushr  8 and 0xffL).toByte()
+    data[limit++] = (v         and 0xffL).toByte()
     tail.limit = limit
     size += 8L
     return this
   }
 
+
```
Egorand: nit: empty line

okio/jvm/src/main/java/okio/Okio.kt
```diff
@@ -213,3 +213,23 @@ fun Path.source(vararg options: OpenOption): Source =
 internal val AssertionError.isAndroidGetsocknameError: Boolean get() {
   return cause != null && message?.contains("getsockname failed") ?: false
 }
+
+// TODO(jwilson): move these to Util.kt once that’s Kotlin?
```
Egorand: 👍

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: Umm it was an honest mistake?
screen shot 2018-04-30 at 8 50 41 am

swankjesse: Fixed. `ushr` wanted to be infix!

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: Fixed.

okio/jvm/src/main/java/okio/Buffer.kt
Egorand: Ah, I see. I don't like parameter name hints precisely because they screw up formatting.

PR #421, opened 2018-04-30
Share code when seeking to a specific segment

swankjesse:
I was hoping I could use an inline function + destructuring return to do this.
Unfortunately that approach always allocates


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -1291,69 +1289,75 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
-    // Pick the first segment to scan. This is the first segment with offset <= fromIndex.
-    var s: Segment?
-    var offset: Long
-    s = head
     when {
-      s == null -> {
-        // No segments to scan!
-        return -1L
-      }
       size - fromIndex < fromIndex -> {
```
JakeWharton: this should probably just be an if/else now that it's only two branches

PR #423, opened 2018-05-01
Improve usability from Kotlin

swankjesse:
Test named parameters. Add operator functions to Buffer and ByteString.


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -46,10 +46,9 @@ import javax.crypto.spec.SecretKeySpec
  */
 class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
   @JvmField internal var head: Segment? = null
-  @JvmField internal var size: Long = 0L
 
-  /** Returns the number of bytes currently in this buffer.  */
-  fun size() = size
+  @get:JvmName("size")
+  var size: Long = 0L
```
Egorand: This makes it public?

swankjesse: I don’t suppose I can name the getter `size()` and also access it as a field from within Kotlin?

Closest I got was having a `@JvmField private var size_` separate from the `var size` that had a getter. Felt unsatisfying, especially referring to `size_` everywhere inside of `Buffer.kt`.

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -270,25 +276,11 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
   }
 
   /** Returns the byte at `pos`.  */
-  fun getByte(pos: Long): Byte {
+  @JvmName("getByte")
+  operator fun get(pos: Long): Byte {
```
Egorand: Nice!

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -1299,7 +1280,7 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
       var offset = 0L
       while (true) {
         val nextOffset = offset + (s.limit - s.pos)
-        if (nextOffset >= fromIndex) break
```
swankjesse: this might have been a bug in `indexOf()`. I’m looking.

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -1299,7 +1280,7 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
+        if (nextOffset > fromIndex) break
```
Egorand: Is this intentional?

okio/jvm/src/test/java/okio/BufferKotlinTest.kt
```diff
@@ -0,0 +1,139 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Assert
```
Egorand: Tests should be using `kotlin.test`, can be a followup

okio/jvm/src/main/java/okio/Buffer.kt
JakeWharton: If it was `private set` you could. Not sure about how `internal set` (which you're missing, as Egor points out) behaves.

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: It was a performance bug, not a correctness bug! We were returning one segment too soon in some off-by-one situations.

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: Yep. It was a bug!

swankjesse: But only a performance bug thankfully.

okio/jvm/src/test/java/okio/BufferKotlinTest.kt
swankjesse: Fixed

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: Added `private set` to suppress the public setter!

PR #425, opened 2018-05-05
Move helpers from Okio to Util

Egorand:


okio/src/main/kotlin/okio/Util.kt
```diff
@@ -49,6 +49,24 @@ internal fun Long.reverseBytes(): Long {
...
+internal inline infix fun Byte.and(other: Long): Long = toLong() and other
+
+@Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
+internal inline infix fun Int.and(other: Long): Long = toLong() and other
+
+@Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
+internal inline fun minOf(a: Long, b: Int): Long = minOf(a, b.toLong())
+
+@Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
+internal inline fun minOf(a: Int, b: Long): Long = minOf(a.toLong(), b)
```
JakeWharton: Unrelated: I'm pretty sure we always use these with positive numbers. Perhaps we should switch to overloading `coerceAtLeast` so that we can return an `Int` and avoid the downcast at the consumer site.

PR #426, opened 2018-05-05
Remove JDK dependencies from Options

Egorand:


JakeWharton:
The error doesn't really indicate the change. Any idea what it is?

Egorand:
I'm pretty sure it's because `kotlin.AbstractList` is not a typealias but actually a different class. The error points to `SegmentedByteString` which is `internal` and should be added to excludes.

Egorand:
Like introducing our own `okio.AbstractList` and typealiasing it to `java.util.AbstractList`? Sure, but is it worth it? I think it's a perfectly fine breaking change for 2.0

swankjesse:
Will somebody in 1.x be able to downcast Options to List? If that works I'm happy.

Egorand:
I can actually do `java.util.List<ByteString> options = Options.of()` inside a Java file, although I don't completely understand how this works, as there are no typealiases in the hierarchy.. @JakeWharton is this a compiler trick? Would someone who has a pure-Java project with okio 2 and all necessary Kotlin dependencies on the classpath be able to compile this code?

Egorand:
I just tried putting an okio JAR along with a Kotlin stdlib JAR into a Java-only project and casting `Options` to `java.util.List`, and it won't compile, hence you'll need kotlinc to be able to build the project. I'll decline the PR for now, but we might add our own `AbstractList` typealias going forward.

PR #427, opened 2018-05-08
SelectBenchmark measures select(Options)

swankjesse:


okio/jvm/build.gradle
```diff
@@ -46,6 +51,7 @@ dependencies {
 
   jmh deps.jmh.core
   jmh deps.jmh.generator
+  compile "org.openjdk.jmh:jmh-core:1.19"
```
Egorand: `compile deps.jmh.core`

okio/jvm/build.gradle
```diff
@@ -23,7 +27,21 @@ configurations {
   baseline
 }
 
+jmhJar {
```
swankjesse: Struggled for a long time and eventually added this. I’m not sure if this is what makes it work, but it appears to make it work.

PR #433, opened 2018-05-09
Remove test-only method `segmentSizes` from Buffer

bnorm:
With the new Buffer.UnsafeCursor API, this method can be replaced by
navigating the Buffer with a cursor. This removes a method that is only
used in the test package and was marked as for testing only.

### DIFF
--- a/okio/jvm/src/test/java/okio/TestUtil.kt
+++ b/okio/jvm/src/test/java/okio/TestUtil.kt
@@ -32,7 +32,17 @@ object TestUtil {
   const val SEGMENT_SIZE = Segment.SIZE
   const val REPLACEMENT_CHARACTER: Int = Buffer.REPLACEMENT_CHARACTER
   @JvmStatic fun segmentPoolByteCount() = SegmentPool.byteCount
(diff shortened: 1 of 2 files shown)


okio/jvm/src/test/java/okio/TestUtil.kt
```diff
@@ -32,7 +32,17 @@ object TestUtil {
   const val SEGMENT_SIZE = Segment.SIZE
   const val REPLACEMENT_CHARACTER: Int = Buffer.REPLACEMENT_CHARACTER
   @JvmStatic fun segmentPoolByteCount() = SegmentPool.byteCount
-  @JvmStatic fun segmentSizes(buffer: Buffer) = buffer.segmentSizes()
+
+  @JvmStatic
+  fun segmentSizes(buffer: Buffer): List<Int> {
+    val result = mutableListOf<Int>()
+    buffer.readUnsafe().use { cursor ->
+        while (cursor.next() > 0) {
```
Egorand: nit: I believe the indentation should be 2 spaces here

PR #436, opened 2018-05-10
Avoid O(N*N) loops constructing Options

swankjesse:
We had two N*N loops here. We can do better.

### DIFF
--- a/okio/jvm/src/main/java/okio/Options.kt
+++ b/okio/jvm/src/main/java/okio/Options.kt
@@ -36,36 +36,38 @@ class Options private constructor(
 
       // Sort the byte strings which is required when recursively building the trie. Map the sorted
       // indexes to the caller's indexes.
-      val sortedByteStrings = byteStrings.sortedArray()
-      // TODO(jwilson): this is a O(N*N) loop to recover the original indexes. Ick.
-      val sortedIndexes = IntArray(sortedByteStrings.size) {
-        byteStrings.indexOf(sortedByteStrings[it])
+      val list = byteStrings.toMutableList()
+      list.sort()
+      val indexes = mutableListOf(*byteStrings.map { -1 }.toTypedArray())
+      byteStrings.forEachIndexed { callerIndex, byteString ->
+        val sortedIndex = list.binarySearch(byteString)
+        indexes[sortedIndex] = callerIndex
       }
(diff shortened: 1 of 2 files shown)


okio/jvm/src/main/java/okio/Options.kt
```diff
@@ -37,28 +37,34 @@ class Options private constructor(
...
-      collectEligible@
-      for (i in 0 until sortedByteStrings.size) {
-        val byteString = sortedByteStrings[i]
-        val index = sortedIndexes[i]
-        for (j in trieIndexes.size - 1 downTo 0) {
-          if (byteString.startsWith(trieByteStrings[j]) && index >= trieIndexes[j]) {
-            continue@collectEligible // This value can never be returned.
+      val trieByteStrings = sortedByteStrings.toMutableList()
+      val trieIndexes = sortedIndexes.toMutableList()
+      var a = 0
```
swankjesse: kotlin’s lack of a 3-part `for` loop hurts us here. I need the index ’cause I’m looping over parallel arrays.

okio/jvm/src/main/java/okio/Options.kt
```diff
@@ -37,28 +37,34 @@ class Options private constructor(
...
+      while (a < trieByteStrings.size) {
```
swankjesse: neither of these can be range loops ’cause we’re shortening trieByteStrings.size as we go

okio/jvm/src/main/java/okio/Options.kt
```diff
@@ -37,28 +37,34 @@ class Options private constructor(
...
-      val trieIndexes = mutableListOf<Int>()
-      collectEligible@
-      for (i in 0 until sortedByteStrings.size) {
-        val byteString = sortedByteStrings[i]
-        val index = sortedIndexes[i]
-        for (j in trieIndexes.size - 1 downTo 0) {
-          if (byteString.startsWith(trieByteStrings[j]) && index >= trieIndexes[j]) {
-            continue@collectEligible // This value can never be returned.
+      val trieByteStrings = sortedByteStrings.toMutableList()
+      val trieIndexes = sortedIndexes.toMutableList()
```
Egorand: `sortedIndexes` can be declared as `MutableList` from the beginning to avoid creating a copy. Same for `sortedByteStrings`, we can sort `byteStrings` in place and copy to a `MutableList` - currently it's being copied twice.

PR #437, opened 2018-05-10
Change ByteString.size from fun to val

### LINKED ISSUE #428: ByteString.size
I’d like to change this from a fun to a val for Kotlin.

Egorand:
Closes #428

### DIFF
--- a/okio/jvm/src/main/java/okio/ByteString.kt
+++ b/okio/jvm/src/main/java/okio/ByteString.kt
@@ -227,7 +227,11 @@ internal constructor(
   operator fun get(index: Int): Byte = getByte(index)
 
   /** Returns the number of bytes in this ByteString. */
-  open fun size() = data.size
+  val size
+    @JvmName("size") get() = getSize()
+
+  // Hack to work around Kotlin's limitation for using JvmName on open/override vals/funs
+  internal open fun getSize() = data.size
 
   /** Returns a byte array containing a copy of the bytes in this `ByteString`. */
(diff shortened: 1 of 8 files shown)


okio/jvm/src/main/java/okio/ByteString.kt
```diff
@@ -227,7 +227,11 @@ internal constructor(
   operator fun get(index: Int): Byte = getByte(index)
 
   /** Returns the number of bytes in this ByteString. */
-  open fun size() = data.size
+  val size
+    @JvmName("size") get() = getSize()
+
+  // Hack to work around Kotlin's limitation for using JvmName on open/override vals/funs
+  protected open fun getSize() = data.size
```
Egorand: On a second thought, `internal` might be a better option? This way it shouldn't appear in KDoc

Egorand:
Japicmp: `methodExcludes = ['okio.ByteString#size()', 'okio.SegmentedByteString#size()']`

swankjesse:
Is the exclusion for final/non-final?

Egorand:
@swankjesse in `SegmentedByteString` it's for "REMOVED METHOD: PUBLIC(-) int size()", which makes sense, since this method is not overridden in that class anymore - it shouldn't manifest in the bytecode if it's inherited from the parent, right?

swankjesse:
Gotcha. SegmentedByteString is not a public type, so I think this is a mistake in japicmp.

Issue #431, opened 2018-05-09
JavaScript ByteString


swankjesse:
I’d like to confirm that ByteString will work with Kotlin for JavaScript. This can be pretty basic: the model type and a test case.

PR #443, opened 2018-05-12
Crossplatform Base64

Egorand:
Prerequisite for #431


okio/src/main/kotlin/okio/Base64.kt
```diff
@@ -143,5 +143,5 @@ internal fun ByteArray.encodeBase64(map: ByteArray = BASE64): String {
       out[index] = '='.toByte()
     }
   }
-  return String(out, Charsets.US_ASCII)
-}
\ No newline at end of file
+  return out.toAsciiString()
```
swankjesse: You can use UTF-8 here. The behavior is the same!

Egorand: Yep, true

PR #446, opened 2018-05-13
Backport trie-based select to Okio 1.x

swankjesse:


okio/src/main/java/okio/Options.java
```diff
@@ -16,18 +16,233 @@
...
+    for (int a = 0; a < list.size(); a++) {
+      ByteString prefix = list.get(a);
+      for (int b = a + 1; b < list.size(); ) {
+        ByteString byteString = list.get(b);
+        if (!byteString.startsWith(prefix)) break;
+        if (byteString.size() == prefix.size()) {
+          throw new IllegalArgumentException("duplicate option: " + byteString);
+        }
+        if (indexes.get(b) > indexes.get(a)) {
+          list.remove(b);
```
Egorand: We're allocating a `List` and copying the input every time, however this branch may not be reached at all. Would it be better to stick to arrays, mark an element if it needs to be removed, and copy elements into a smaller array if anything has been marked?

Egorand:
Would be great to benchmark both versions and compare the results

okio/src/main/java/okio/Options.java
swankjesse: The logic gets awkward if there are multiple such elements to remove. I’m more annoyed that we allocate a bunch of boxed integers!

PR #463, opened 2018-06-04
Update README.md

GianfrancoMS:
In the cheat sheet for bytes, shorts and ints, where it says 'an unsigned value into a signed value' it should say 'a signed value into an unsigned value'.


swankjesse:
Thanks!

PR #465, opened 2018-06-18
Remove impossible ByteBuffer IOExceptions.

NightlyNexus:
This is a source-incompatible change. Unsure if that's acceptable.


swankjesse:
Nope. We want to start with the same exact exceptions as 1.x.

NightlyNexus:
👍

PR #471, opened 2018-07-11
Avoid excess ByteArray copies when creating ByteStrings

bnorm:
ByteString.of() was copying the data 2 extra times before passing it to
the commonOf() method. Avoid these extra copies by converting the
commonOf() method to take a ByteArray directly and only copy the array
in that method.

MessageDigests and Macs create unique ByteArrays which can safely be
wrapped directly with a ByteString without having to copy the array.
Avoid the spread operator and using the ByteString.of() create method.

### DIFF
--- a/okio/jvm/src/test/java/okio/ByteStringTest.java
+++ b/okio/jvm/src/test/java/okio/ByteStringTest.java
@@ -93,6 +93,14 @@ public final class ByteStringTest {
   @Parameter(0) public Factory factory;
   @Parameter(1) public String name;
 
+  @Test public void ofCopy() {
+    byte[] bytes = "Hello, World!".getBytes(Charsets.UTF_8);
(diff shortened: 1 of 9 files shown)


okio/jvm/src/test/java/okio/ByteStringTest.java
```diff
@@ -93,6 +93,14 @@
   @Parameter(0) public Factory factory;
   @Parameter(1) public String name;
 
+  @Test public void ofCopy() {
+    byte[] bytes = "Hello, World!".getBytes(Charsets.UTF_8);
+    ByteString byteString = ByteString.of(bytes);
+    // Verify that the bytes were copied out.
+    bytes[4] = (byte) 'a';
```
adriancole: nice

PR #482, opened 2018-07-19
Remove obsolete assertion message.

NightlyNexus:
The empty byte string is supported now.

The assertion is still right, but the error message seems confusing.


NightlyNexus:
no, whoops, i'm catching up. nm!

PR #486, opened 2018-07-24
Make Buffer exceptions consistent with RealBufferedSource

swankjesse:
This is inspired by ksperling's work on improving exception-consistency
between Buffer and RealBufferedSource. I wasn't willing to make an
exception-signature change for a 1.x release, but the 2.x release is a
good opportunity to fix this longstanding problem.

### DIFF
--- a/okio/jvm/src/main/java/okio/Buffer.kt
+++ b/okio/jvm/src/main/java/okio/Buffer.kt
@@ -252,8 +252,9 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
     return result
   }
 
+  @Throws(EOFException::class)
   override fun readByte(): Byte {
-    check(size != 0L) { "size == 0" }
+    if (size == 0L) throw EOFException()
 
     val segment = head!!
     var pos = segment.pos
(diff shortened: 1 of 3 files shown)


swankjesse:
Differences vs. https://github.com/square/okio/pull/321 that we should consider:

 * [ ] When we throw EOFException we don’t consistently exhaust the throwing Buffer/BufferedSource. Is it worth it? If so we should use @ksperling’s nice docs.
 * [ ] We should add a one-at-a-time Buffer to BufferedSourceTest.

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -252,8 +252,9 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
     return result
   }
 
+  @Throws(EOFException::class)
   override fun readByte(): Byte {
-    check(size != 0L) { "size == 0" }
+    if (size == 0L) throw EOFException()
```
Egorand: Would it make sense to keep the messages?

swankjesse: I think IllegalStateException needs the message ’cause it could be anything. For EOFException the class name says it all.

PR #469, opened 2018-07-09
Implement ByteString for Kotlin/Native

bnorm:
Fixes #453 

Not sure if this issue was really up for grabs but I've been dipping my toe in Kotlin/Native a little and thought this would be a good first exercise.

Issue #490, opened 2018-08-01
Throttler okio source implement?


Gohan:
Hi, I can't find the implement of the throttler implement of okio. Don't know where I can find it.

![default](https://user-images.githubusercontent.com/212567/43537461-f19bc2d4-95f1-11e8-8842-a198103719f3.png)

https://youtu.be/WvyScM_S88c?t=29m37s

JakeWharton:
It was never finished and the branch seems to have been deleted. I might have it on a backup somewhere. What's your use case?

Gohan:
Just for curious, learn some tricks around okio

Tolriq:
@JakeWharton Did not know there was already something as some users did asked me that.

They would like that I add an option to not download at full speed as okio is so fast that it takes all their 4G BW and prevent browsing at the same time.

Or maybe there's already a recipe somewhere else?

PR #491, opened 2018-08-03
Implement UTF-8 encode and decode for JS

### LINKED ISSUE #452: ByteArray to/from String implementations for JavaScript
Blocks #431 

We should either:

1) Implement this ourselves in Kotlin
2) Implement this ourselves in JS and call from Kotlin
3) Call existing JS code from Kotlin if it's available

bnorm:
Kotlin/JS does not currently have a implementation for transforming a
UTF-8 encoded ByteArray into a String and vice versa. Create an
implementation using some of the logic which already exists within Okio.
While this implementation could be used for Kotlin/JVM and
Kotlin/Native, both of these platforms have their own implementation
which is faster.

Fixes #452

Since I was already working in this area because of #469, I made an attempt to implement UTF-8 encoding and decoding for Kotlin/JS. I decided on implementing this in Kotlin because there might be other places throughout the project we might want to use these inline functions.

For contrast, [kotlinx-io](https://github.com/Kotlin/kotlinx-io/) uses the `text-encoding` JS library. I would submit that implementing native in Okio keeps the number of dependencies of the project to just the Kotlin stdlib and also with Kotlin inline functions, a lot of common UTF-8 behavior could be centralized.

### DIFF
--- a/okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
+++ b/okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
@@ -1,0 +1,119 @@
+/*
+ * Copyright (C) 2018 Square, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.squareup.okio.benchmarks;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.TimeUnit;
+
+import kotlin.text.StringsKt;
+import okio.Buffer;
+import okio.ByteString;
+
+import org.openjdk.jmh.Main;
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.runner.RunnerException;
+
+@Fork(1)
+@Warmup(iterations = 5, time = 2)
+@Measurement(iterations = 5, time = 2)
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class BufferUtf8Benchmark {
+
+  private static final Map<String, String> strings = new HashMap<>();
+  static {
+    strings.put(
+        "ascii",
+        StringsKt.repeat(
+            "Um, I'll tell you the problem with the scientific power that you're using here, "
+                + "it didn't require any discipline to attain it. You read what others had done and you "
+                + "took the next step. You didn't earn the knowledge for yourselves, so you don't take any "
+                + "responsibility for it. You stood on the shoulders of geniuses to accomplish something "
+                + "as fast as you could, and before you even knew what you had, you patented it, and "
+                + "packaged it, and slapped it on a plastic lunchbox, and now you're selling it, you wanna "
+                + "sell it.",
+            500));
+
+    strings.put(
+        "sparse",
+        StringsKt.repeat(
+            "Um, I'll tell you the problem with the scientific power that you're using here, "
+                + "it didn't require any discipline to attain it. 𝒀ο𝗎 read what others had done and you "
+                + "took the next step. 𝘠ⲟ𝖚 didn't earn the knowledge for yourselves, so you don't take any "
+                + "responsibility for it. 𝛶𝛔𝔲 stood on the shoulders of geniuses to accomplish something "
+                + "as fast as you could, and before you even knew what you had, you patented it, and "
+                + "packaged it, and slapped it on a plastic lunchbox, and now you're selling it, you wanna "
+                + "sell it.",
+            500));
+
+    strings.put(
+        "utf8",
+        StringsKt.repeat(
+            "Սｍ, I'll 𝓽𝖾ll ᶌօ𝘂 ᴛℎ℮ 𝜚𝕣०ｂl𝖾ｍ ｗі𝕥𝒽 𝘵𝘩𝐞 𝓼𝙘𝐢𝔢𝓷𝗍𝜄𝚏𝑖ｃ 𝛠𝝾ｗ𝚎𝑟 𝕥ｈ⍺𝞃 𝛄𝓸𝘂'𝒓𝗲 υ𝖘𝓲𝗇ɡ 𝕙𝚎𝑟ｅ, "
+                + "𝛊𝓽 ⅆ𝕚𝐝𝝿'𝗍 𝔯𝙚𝙦ᴜ𝜾𝒓𝘦 𝔞𝘯𝐲 ԁ𝜄𝑠𝚌ι𝘱lι𝒏ｅ 𝑡𝜎 𝕒𝚝𝖙𝓪і𝞹 𝔦𝚝. 𝒀ο𝗎 𝔯𝑒⍺𝖉 ｗ𝐡𝝰𝔱 𝞂𝞽һ𝓮𝓇ƽ հ𝖺𝖉 ⅾ𝛐𝝅ⅇ 𝝰πԁ 𝔂ᴑᴜ 𝓉ﮨ၀𝚔 "
+                + "т𝒽𝑒 𝗇𝕖ⅹ𝚝 𝔰𝒕е𝓅. 𝘠ⲟ𝖚 𝖉ⅰԁ𝝕'τ 𝙚𝚊ｒ𝞹 𝘵Ꮒ𝖾 𝝒𝐧هｗl𝑒𝖉ƍ𝙚 𝓯૦ｒ 𝔂𝞼𝒖𝕣𝑠𝕖l𝙫𝖊𝓼, 𐑈о ｙ𝘰𝒖 ⅆە𝗇'ｔ 𝜏α𝒌𝕖 𝛂𝟉ℽ "
+                + "𝐫ⅇ𝗌ⲣ๐ϖ𝖘ꙇᖯ𝓲l𝓲𝒕𝘆 𝐟𝞼𝘳 𝚤𝑡. 𝛶𝛔𝔲 ｓ𝕥σσ𝐝 ﮩ𝕟 𝒕𝗁𝔢 𝘴𝐡𝜎ᴜlⅾ𝓮𝔯𝚜 𝛐𝙛 ᶃ𝚎ᴨᎥս𝚜𝘦𝓈 𝓽𝞸 ａ𝒄𝚌𝞸ｍρl𝛊ꜱ𝐡 𝓈𝚘ｍ𝚎𝞃𝔥⍳𝞹𝔤 𝐚𝗌 𝖋ａ𝐬𝒕 "
+                + "αｓ γ𝛐𝕦 𝔠ﻫ𝛖lԁ, 𝚊π𝑑 Ь𝑒𝙛૦𝓇𝘦 𝓎٥𝖚 ⅇｖℯ𝝅 𝜅ո𝒆ｗ ｗ𝗵𝒂𝘁 ᶌ੦𝗎 ｈ𝐚𝗱, 𝜸ﮨ𝒖 𝓹𝝰𝔱𝖾𝗇𝓽𝔢ⅆ і𝕥, 𝚊𝜛𝓭 𝓹𝖺ⅽϰ𝘢ℊеᏧ 𝑖𝞃, "
+                + "𝐚𝛑ꓒ 𝙨l𝔞р𝘱𝔢𝓭 ɩ𝗍 ہ𝛑 𝕒 ｐl𝛂ѕᴛ𝗂𝐜 l𝞄ℼ𝔠𝒽𝑏ﮪ⨯, 𝔞ϖ𝒹 ｎ𝛔ｗ 𝛾𝐨𝞄'𝗿𝔢 ꜱ℮ll𝙞ｎɡ ɩ𝘁, 𝙮𝕠𝛖 ｗ𝑎ℼ𝚗𝛂 𝕤𝓮ll 𝙞𝓉.",
+            500));
+  }
+
+  @Param({"ascii", "sparse", "utf8"})
+  String encoding;
+
+  Buffer buffer;
+  ByteString decode;
+  String encode;
+
+  @Setup
+  public void setup() {
+    buffer = new Buffer();
+    encode = strings.get(encoding);
+
+    Buffer temp = new Buffer();
+    temp.writeUtf8(encode);
+    decode = temp.snapshot();
+  }
+
+  @Benchmark
+  public void writeUtf8() {
+    buffer.writeUtf8(encode);
+    buffer.clear();
+  }
+
+  @Benchmark
+  public String readUtf8() {
+    buffer.write(decode);
+    return buffer.readUtf8();
+  }
+
+  public static void main(String[] args) throws IOException, RunnerException {
+    Main.main(new String[] { BufferUtf8Benchmark.class.getName()});
+  }
+}
--- a/okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
+++ b/okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
@@ -20,10 +20,6 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
-import kotlin.text.StringsKt;
-import okio.Buffer;
-import okio.ByteString;
-
 import org.openjdk.jmh.Main;
 import org.openjdk.jmh.annotations.Benchmark;
 import org.openjdk.jmh.annotations.BenchmarkMode;
@@ -38,6 +34,9 @@ import org.openjdk.jmh.annotations.State;
 import org.openjdk.jmh.annotations.Warmup;
 import org.openjdk.jmh.runner.RunnerException;
 
+import okio.Buffer;
+import okio.ByteString;
+
 @Fork(1)
 @Warmup(iterations = 5, time = 2)
 @Measurement(iterations = 5, time = 2)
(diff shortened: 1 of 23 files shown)


okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -898,71 +898,13 @@ class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
-        c < 0x80 -> {
-          val tail = writableSegment(1)
-          val data = tail.data
-          val segmentOffset = tail.limit - i
-          val runLimit = minOf(endIndex, Segment.SIZE - segmentOffset)
-
-          // Emit a 7-bit character with 1 byte.
-          data[segmentOffset + i++] = c.toByte() // 0xxxxxxx
-
-          // Fast-path contiguous runs of ASCII characters. This is ugly, but yields a ~4x performance
```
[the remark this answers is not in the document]
bnorm: I'll create a benchmark to verify.

My hope is that by now always writing directly to the segment and skipping `writeByte`, we keep the speed improvement and also introduce gains for non-ASCII paths.

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -114,89 +114,169 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
+  val chars = CharArray(size)
+  var length = 0
+  processUtf8CodePoints(0, size) { c ->
+    if (c < 0x10000) {
+      chars[length++] = c.toChar()
+    } else {
+      // Unicode code point:    00010000000000000000 + xxxxxxxxxxyyyyyyyyyy (21 bits)
+      // UTF-16 high surrogate: 110110xxxxxxxxxx (10 bits)
+      // UTF-16 low surrogate:  110111yyyyyyyyyy (10 bits)
+      chars[length++] = ((c - 0x010000) shr 10 and 0x03ff or 0xd800).toChar()
```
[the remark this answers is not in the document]
bnorm: I believe it is necessary. But the unit tests currently do not fail if it is removed so I will will look into either creating a new unit test, adding better documentation, or removing it if I find that it is not required.

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -114,89 +114,169 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
   (codePoint in 0x00..0x1F) || (codePoint in 0x7F..0x9F)
 
-// TODO: Combine with Buffer.readUtf8CodePoint if possible
-internal fun ByteArray.codePointAt(index: Int): Int {
-  val b0 = this[index]
+// TODO combine with Buffer.readUtf8CodePoint?
+internal inline fun ByteArray.processUtf8CodePoints(
+  beginIndex: Int,
+  endIndex: Int,
+  yield: (Int) -> Unit
```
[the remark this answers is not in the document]
bnorm: I think so? I know that the [SequenceBuilder](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental/-sequence-builder/index.html) receiver class used by [`buildIterator`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental/build-iterator.html) and [`buildSequence`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental/build-sequence.html) uses yield as a function name. I was kind of following that convention but kept it as a function reference so it was fully inline'd.

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -114,89 +114,169 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
-      return REPLACEMENT_CHARACTER
+    // Read the continuation bytes. If we encounter a non-continuation byte, the sequence thus far is
+    // decoded as the replacement character.
+    for (i in 1..continuationCount) {
+      val b = this[index++]
+      if (b and 0xc0 == 0x80) {
+        // 0x10xxxxxx
+        codePoint = (codePoint shl 6) or (b and 0x3f)
+      } else {
+        yield(REPLACEMENT_CHARACTER)
```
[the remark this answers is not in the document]
bnorm: I believe you are correct. Good catch. I'll create some more malformed UTF-8 unit tests to catch cases like this.

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -114,89 +114,169 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
-    val b = this[index + i]
-    if (b and 0xc0 == 0x80) {
-      // 0x10xxxxxx
-      codePoint = codePoint shl 6
-      codePoint = codePoint or (b and 0x3f)
-    } else {
-      return REPLACEMENT_CHARACTER
+    // Read the continuation bytes. If we encounter a non-continuation byte, the sequence thus far is
+    // decoded as the replacement character.
+    for (i in 1..continuationCount) {
```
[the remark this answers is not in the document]
bnorm: I think I went down a few roads of very ugly - but ultimately not very fruitful - optimization attempts where having this be an inclusive representation was more important. I'll convert it back to exclusive.

bnorm:
I completely agree about speed. Especially for Moshi, this is probably one of the most important areas of Okio. That's why I opted for only using this new code where I thought it would give us a performance improvement (the `Buffer.writeUtf8` function) or where it was required (Kotlin/JS ByteString encoding/decoding).

I did benchmark it against the native Java UTF-8 implementation and wasn't able to get it to perform better and went down a few different paths of very confusing code which didn't yield any significant gains. It was within a close margin, but I could never quite get it where I was happy with the performance to replace the current implementation. But with your comment of favoring performant code over easy to understand code, I will give it another swing.

Question: Java's decoding of UTF-8 is very strict and will throw exceptions instead of introduce replacement characters. This was another reason for not wanting to move to the more lenient version that Okio uses. If the performance is able to beat or match Java's, should we switch to Okio's more lenient encoding/decoding?

Let's do this! Current tasks:
[] Create unit test(s) to verify math in calculating high surrogate from a code point.
[] Create benchmark to verify no performance degradation of `Buffer.writeUtf8`.
[] Create unit tests to verify correct handling of all cases where a replacement character is used when decoding UTF-8.
[] Covert back to [inclusive..exclusive) for-loop iteration.
[] Investigate even the most minor of performance gains for encoding and decoding UTF-8.

Let me know if I missed anything.

swankjesse:
@bnorm:
> _Java's decoding of UTF-8 is very strict and will throw exceptions instead of introduce replacement characters._

I think it depends on which API you use. The constructor and getBytes of java.lang.String use replacement characters.

In any case, your plan is perfect. You may have found an opportunity to be faster than what we have today.

Please optimize & benchmark for the ASCII (ie. 7-bit) subset of UTF-8. Most strings that aren’t user-facing are ASCII-only (JSON field names, HTTP headers, etc.). And also for actual user data with many languages and occasional emojis!

The only two charsets we really care about are UTF-8 and UTF-16. Code points are usually just a bridge between them. We want fast transcoding and don't really care about UTF-8 to codepoints or codepoints to UTF-16.

Last thing: in my experience code that reads and writes fields more than necessary is slower than code that doesn't. Incrementing `limit` can be slow!

bnorm:
> I think it depends on which API you use. The constructor and getBytes of java.lang.String use replacement characters.

You are correct. I was too deep in the encoder/decoder and was only seeing the exceptions being thrown.

So I've moved a lot of `Utf8Test` and the JVM `ByteStringTest` over to common to improve cross platform UTF-8 testing and I've discovered a couple things that I've found interesting:
1. ByteString.EMPTY is `null` for both Kotlin/JS and Kotlin/Native. The same applies for HEX_DIGITS but I was able to work around that by moving it exclusively to `common/ByteString`. I'm guessing a bug will have to be reported though.
2. Kotlin/Native string to byte array conversion is slightly different in that it will encode unknown codepoints as the replacement character (`\ufffd`) instead of `?` like the JVM. This isn't terribly surprising to me but it does causes unit tests to fail for only Kotlin/Native. Not sure what to do here as part of me wants the library to be consistent with itself and the other part wants it to be consistent with the platform. Any guidance?

okio/src/main/kotlin/okio/Utf8.kt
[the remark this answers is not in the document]
bnorm: Yes, this is necessary. The Kotlin/JS unit tests will now fail without it.

okio/jvm/src/main/java/okio/Buffer.kt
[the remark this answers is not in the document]
bnorm: Yeah, we lost it. I've reverted it with some updates to the other branches that resulted in about a 30% increase in performance for full UTF-8 encoding. I added a benchmark to confirm.

Previous
```
Benchmark                                             (encoding)  (multiples)    Mode    Cnt      Score     Error  Units
BufferWriteUtf8Benchmark.writeUtf8                         ascii         1000  sample  33926    294.632 ±   1.734  us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.00         ascii         1000  sample           239.872            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.50         ascii         1000  sample           256.512            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.90         ascii         1000  sample           442.880            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.95         ascii         1000  sample           486.400            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.99         ascii         1000  sample           588.800            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.999        ascii         1000  sample          1149.227            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.9999       ascii         1000  sample          1497.247            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p1.00         ascii         1000  sample          5005.312            us/op
BufferWriteUtf8Benchmark.writeUtf8                          utf8         1000  sample   1810   5525.509 ± 109.329  us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.00          utf8         1000  sample          4440.064            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.50          utf8         1000  sample          4841.472            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.90          utf8         1000  sample          8165.786            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.95          utf8         1000  sample          8683.520            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.99          utf8         1000  sample          9386.230            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.999         utf8         1000  sample         10717.037            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.9999        utf8         1000  sample         13959.168            us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p1.00          utf8         1000  sample         13959.168            us/op
```

With changes
```
Benchmark                                             (encoding)  (multiples)    Mode    Cnt      Score    Error  Units
BufferWriteUtf8Benchmark.writeUtf8                         ascii         1000  sample  36298    275.342 ±  2.382  us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.00         ascii         1000  sample           221.952           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.50         ascii         1000  sample           238.080           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.90         ascii         1000  sample           395.264           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.95         ascii         1000  sample           439.296           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.99         ascii         1000  sample           608.266           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.999        ascii         1000  sample          1148.316           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.9999       ascii         1000  sample          1780.978           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p1.00         ascii         1000  sample         14319.616           us/op
BufferWriteUtf8Benchmark.writeUtf8                          utf8         1000  sample   2871   3486.085 ± 50.592  us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.00          utf8         1000  sample          2945.024           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.50          utf8         1000  sample          3125.248           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.90          utf8         1000  sample          4672.717           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.95          utf8         1000  sample          5786.829           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.99          utf8         1000  sample          6443.500           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.999         utf8         1000  sample          7521.042           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p0.9999        utf8         1000  sample          8036.352           us/op
BufferWriteUtf8Benchmark.writeUtf8:writeUtf8·p1.00          utf8         1000  sample          8036.352           us/op
```

bnorm:
I've been digging into UTF-8 decoding lately and been getting some encouraging results. Here's the latest benchmarks.

```
Benchmark                                 (encoding)   Mode  Cnt     Score      Error  Units
Utf8Benchmark.bytesToStringJava                ascii  thrpt    5  6639.318 ± 1508.839  ops/s
Utf8Benchmark.bytesToStringOkio                ascii  thrpt    5  4476.771 ±  624.647  ops/s
Utf8Benchmark.bytesToStringAsciiBaseline       ascii  thrpt    5  4426.589 ±  242.728  ops/s
Utf8Benchmark.bytesToStringJava               sparse  thrpt    5  2728.167 ±  558.590  ops/s
Utf8Benchmark.bytesToStringOkio               sparse  thrpt    5  1707.473 ±   33.954  ops/s
Utf8Benchmark.bytesToStringJava                 utf8  thrpt    5   651.735 ±   15.232  ops/s
Utf8Benchmark.bytesToStringOkio                 utf8  thrpt    5   760.409 ±    9.082  ops/s
```

Java gets to cheat with the ASCII dencoding because it doesn't have to copy the char array and we do. That's why I created the baseline benchmark just so I had a realistic comparison. But getting better performance than Java for heavy UTF-8 dencoding is pretty awesome.

Now to figure out what's up with the sparse case (mostly ASCII and a little UTF-8) because we should be able to achieve the same performance as Java for that case as well.

bnorm:
Improved the performance of the sparse decoding. Used the same trick found in Buffer and fast-path'd contiguous ASCII.

Decoding
```
Benchmark                        (encoding)  (length)   Mode  Cnt     Score     Error  Units
Utf8Benchmark.bytesToStringJava       ascii      long  thrpt    5  7268.469 ± 604.575  ops/s
Utf8Benchmark.bytesToStringJava      sparse      long  thrpt    5  3144.136 ± 125.198  ops/s
Utf8Benchmark.bytesToStringJava        utf8      long  thrpt    5   983.142 ± 262.183  ops/s
Utf8Benchmark.bytesToStringOkio       ascii      long  thrpt    5  4630.527 ± 375.846  ops/s
Utf8Benchmark.bytesToStringOkio      sparse      long  thrpt    5  3799.555 ± 139.058  ops/s
Utf8Benchmark.bytesToStringOkio        utf8      long  thrpt    5  1077.456 ±  57.538  ops/s
```

Again, used the same trick for encoding and am crushing Java for sparse UTF-8 data.

Encoding
```
Benchmark                        (encoding)  (length)   Mode  Cnt     Score     Error  Units
Utf8Benchmark.stringToBytesJava       ascii      long  thrpt    5  5634.818 ± 353.374  ops/s
Utf8Benchmark.stringToBytesJava      sparse      long  thrpt    5  2529.262 ± 123.744  ops/s
Utf8Benchmark.stringToBytesJava        utf8      long  thrpt    5  1274.056 ±  87.039  ops/s
Utf8Benchmark.stringToBytesOkio       ascii      long  thrpt    5  5502.655 ± 110.648  ops/s
Utf8Benchmark.stringToBytesOkio      sparse      long  thrpt    5  4566.115 ± 770.189  ops/s
Utf8Benchmark.stringToBytesOkio        utf8      long  thrpt    5  1065.153 ±  31.273  ops/s
```

It hurts the heavy UTF-8 de/encoding a little because every space encountered tries to fast-path now. That might make for a good special case but not sure how real world heavy UTF-8 encoding is though.

Anyways, sorry for all the benchmark data, hopefully it was useful, but I think I'm finally happy with the performance I'm seeing. If there are areas you still want me to investigate, let me know.

swankjesse:
For `\ufffd` vs. `?`, I think either is acceptable. They both render as something looking like a `?`!

swankjesse:
Just studied these benchmarks. The numbers you’re hitting are really good!

I won’t worry too much about space characters harming the mostly non-ASCII case. There isn’t _that_ much extra code to set up the fast path, and it’d be similarly complex to avoid it.

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -110,89 +110,354 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
+      }
+      b0 shr 3 == -2 -> {
+        // 0b11110xxx
+        index += process4Utf8Bytes(index, endIndex) { codePoint ->
+          if (codePoint != REPLACEMENT_CODE_POINT) {
+            // Unicode code point:    00010000000000000000 + xxxxxxxxxxyyyyyyyyyy (21 bits)
+            // UTF-16 high surrogate: 110110xxxxxxxxxx (10 bits)
+            // UTF-16 low surrogate:  110111yyyyyyyyyy (10 bits)
+            /* ktlint-disable no-multi-spaces */
+            yield(((codePoint ushr 10   ) + (0xd800 - (0x010000 ushr 10))).toChar())
```
swankjesse: I wonder how much of this gets evaluated at build time

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -110,89 +110,354 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
+        // 0b111111xxx - Unknown encoding
+        yield(REPLACEMENT_CHARACTER)
+        index++
+      }
+    }
+  }
+}
+
+/* ktlint-disable indent */
+internal val mask2 =
```
swankjesse: const?

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -110,89 +110,354 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
-      codePoint = b0 and 0x1f
-      byteCount = 2 // 11 bits (5 + 6).
-      min = 0x80
+  }
+  return 2
+}
+
+/* ktlint-disable indent */
+internal val mask3 =
+    ((0xe0.toByte() shl 12) xor
```
swankjesse: what’s the toByte() for here? I don’t quite follow these masks

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -110,89 +110,354 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
+    yield(REPLACEMENT_CODE_POINT)
+    return 2
+  }
+  val b3 = this[beginIndex + 3]
+  if (!isUtf8Continuation(b3)) {
+    yield(REPLACEMENT_CODE_POINT)
+    return 3
+  }
+
+  val codePoint =
```
swankjesse: I’m really lost on how this works . . . the codepoint shouldn’t have any masks?

okio/src/main/kotlin/okio/internal/-Utf8.kt
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+package okio.internal
+
+import okio.processUtf8Bytes
+import okio.processUtf8Chars
+
+// TODO For benchmarking, these methods need to be available but preferably invisible
+// to everything else. Putting them in this file, `-Utf8.kt`, makes them invisible to
+// Java but still visible to Kotlin.
```
swankjesse: That’s a trick we employ a lot in Okio!

okio/src/main/kotlin/okio/internal/-Utf8.kt
```diff
@@ -0,0 +1,55 @@
...
+    chars[length++] = c
+  }
+
+  return String(chars, 0, length)
+}
+
+fun String.commonAsUtf8ToByteArray(): ByteArray {
+  val bytes = ByteArray(4 * length)
+
+  // Assume ASCII until a UTF-8 code point is observed. This is ugly but yields
```
swankjesse: Nice

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,67 @@
...
+import okio.internal.commonAsUtf8ToByteArray
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
+import kotlin.test.assertNotEquals
+import kotlin.test.assertSame
 import kotlin.test.assertTrue
+import kotlin.test.fail
+
+class CommonByteStringTest {
```
swankjesse: neat!

swankjesse: ... I assume JUnit’s parameterized tests don’t work for us here because they’re not cross-platform? Too bad; they’re more capable than subclassing

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,67 @@
...
+  class Platform : AbstractByteStringTest(ByteStringFactory.PLATFORM)
+  class Okio : AbstractByteStringTest(ByteStringFactory.OKIO)
+}
+
+interface ByteStringFactory {
+  fun decodeHex(hex: String): ByteString
```
swankjesse: why is this in the factory? It’s the same for both implementations?

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,67 @@
...
+  fun encodeUtf8(s: String): ByteString
+
+  companion object {
+    val PLATFORM: ByteStringFactory = object : ByteStringFactory {
+      override fun decodeHex(hex: String) = hex.decodeHex()
+      override fun encodeUtf8(s: String) = s.encodeUtf8()
+    }
+    val OKIO: ByteStringFactory = object : ByteStringFactory {
+      override fun decodeHex(hex: String) = hex.decodeHex()
+      override fun encodeUtf8(s: String) = ByteString.of(*s.commonAsUtf8ToByteArray())
```
swankjesse: Cool!

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,67 @@
...
+interface ByteStringFactory {
+  fun decodeHex(hex: String): ByteString
+  fun encodeUtf8(s: String): ByteString
+
+  companion object {
+    val PLATFORM: ByteStringFactory = object : ByteStringFactory {
+      override fun decodeHex(hex: String) = hex.decodeHex()
+      override fun encodeUtf8(s: String) = s.encodeUtf8()
+    }
+    val OKIO: ByteStringFactory = object : ByteStringFactory {
```
swankjesse: worth a small bit of doc to explain the consequences of two implementations?

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,67 @@
...
+abstract class AbstractByteStringTest(
+  private val factory: ByteStringFactory
+) {
+  @Test fun get() {
+    val actual = factory.encodeUtf8("abc")
+    assertEquals(actual[0], 'a'.toByte())
+    assertEquals(actual[1], 'b'.toByte())
+    assertEquals(actual[2], 'c'.toByte())
+    try {
+      actual[-1]
```
swankjesse: assert the returned value’s size also?

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -111,10 +157,246 @@ class ByteStringTest {
...
   @Test fun indexOfByteArray() {
-    val byteString = "112233".decodeHex()
+    val byteString = factory.decodeHex("112233")
     assertEquals(0, byteString.indexOf("112233".decodeHex().toByteArray()).toLong())
     assertEquals(1, byteString.indexOf("2233".decodeHex().toByteArray()).toLong())
     assertEquals(2, byteString.indexOf("33".decodeHex().toByteArray()).toLong())
     assertEquals(-1, byteString.indexOf("112244".decodeHex().toByteArray()).toLong())
   }
+
+  // TODO Kotlin/Native and Kotlin/JS have a null ByteString.EMPTY
```
swankjesse: I think I’d prefer an `@Ignore` test instead of commented out code?

swankjesse: ... does kotlin.test have that?

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -111,10 +157,246 @@ class ByteStringTest {
...
+  @Test fun testHashCode() {
+    val byteString = factory.decodeHex("0102")
+    assertEquals(byteString.hashCode().toLong(), byteString.hashCode().toLong())
+    assertEquals(byteString.hashCode().toLong(), "0102".decodeHex().hashCode().toLong())
+  }
+
+  @Test fun toAsciiLowerCaseNoUppercase() {
+    val s = factory.encodeUtf8("a1_+")
+    assertEquals(s, s.toAsciiLowercase())
+    if (factory === ByteStringFactory.PLATFORM) {
```
swankjesse: neat

okio/src/test/kotlin/okio/ByteStringTest.kt
JakeWharton: Yes

okio/src/main/kotlin/okio/Utf8.kt
```diff
@@ -110,89 +110,354 @@ fun String.utf8Size(beginIndex: Int = 0, endIndex: Int = length): Long {
...
+        // 0b111111xxx - Unknown encoding
+        yield(REPLACEMENT_CODE_POINT)
+        index++
+      }
+    }
+  }
+}
+
+// TODO combine with Buffer.readUtf8?
+internal inline fun ByteArray.processUtf8Chars(
```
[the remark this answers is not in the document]
bnorm: Good catch. That would be a correct name for this function.

okio/src/main/kotlin/okio/Utf8.kt
bnorm: I'm not sure. I bet the Java JIT compiler quickly does, but I'm not sure how it would perform on JS or Native. I could pull it out as a constant to make sure.

okio/src/main/kotlin/okio/Utf8.kt
bnorm: Unfortunately xor stops me from marking it as a const. I could hand calculate it and leave the formula as a comment.

okio/src/main/kotlin/okio/Utf8.kt
bnorm: I'll try and come up with a good way of documenting these calculations. It's not as straight forward as it was before.

okio/src/main/kotlin/okio/internal/-Utf8.kt
bnorm: I know; I saw it and stole it! 😄

okio/src/test/kotlin/okio/ByteStringTest.kt
bnorm: I couldn't find any information on parameterized unit tests for multiplatform. There is potential for a framework like [Spek](https://spekframework.org/), but from what I've seen it's not quite ready yet for multiplatform use.

okio/src/test/kotlin/okio/ByteStringTest.kt
bnorm: This file is an attempt to recreate the ByteStringTest class from the JVM project. When/if SegmentedByteString moves to common there will be cases where we will need this method. (It also just made it easier when I copy and pasted the unit tests... 😄)

okio/src/test/kotlin/okio/ByteStringTest.kt
bnorm: Yes, I'll document both and explain what each is trying to test.

okio/src/test/kotlin/okio/ByteStringTest.kt
bnorm: I'm hesitant to disable the whole test because there are valuable assertions which should be tested. I can make a specific comment for each assertion I commented out with a reference to the Kotlin ticket I opened. Would that be acceptable?

okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
```diff
@@ -0,0 +1,119 @@
...
+@Warmup(iterations = 5, time = 2)
+@Measurement(iterations = 5, time = 2)
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class BufferUtf8Benchmark {
+
+  private static final Map<String, String> strings = new HashMap<>();
+  static {
+    strings.put(
```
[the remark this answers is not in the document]
bnorm: I added this to `Utf8Benchmark` so I will add it here as well.

bnorm:
Work before merge:
- [ ] Resolve conflicts with recent ktlint changes and fix local problems
- [ ] Document UTF-8 byte decoding logic
- [ ] Add short string Buffer UTF-8 encoding/decoding performance tests
- [ ] Document why we are not using the native UTF-8 decoding function in Kotlin/Native project
- [ ] Other general cleanup (documentation, constants, unit test assertions, etc.)

I'll make all of these as a separate commit so it is easier to review. Than I can rebase after you've had a chance to review to clean it all up. Than I think this will be ready for merge.

okio/src/test/kotlin/okio/ByteStringTest.kt
bnorm: Or I can pull the 2 assertions out into a separate test and `@Ignore` it with a comment about the Kotlin ticket. That's probably what you wanted in the first place. I got there eventually.

okio/src/main/kotlin/okio/Utf8.kt
swankjesse: Oh, that’s too bad. I think `const` could yield faster bytecode: it might be used directly rather than retrieved.

okio/src/test/kotlin/okio/ByteStringTest.kt
swankjesse: Yep!

okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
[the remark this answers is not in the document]
bnorm: Okay, here's something interesting. I was testing running with small Strings (length 20) and here are the results of the benchmark:

```
Benchmark                        (encoding)  (length)   Mode  Cnt         Score         Error  Units
Utf8Benchmark.bytesToStringJava       ascii        20  thrpt    5  20873546.434 ±  438347.891  ops/s
Utf8Benchmark.bytesToStringJava        utf8        20  thrpt    5  10444191.861 ±  165604.155  ops/s
Utf8Benchmark.bytesToStringJava      sparse        20  thrpt    5  14540137.548 ±  497195.598  ops/s
Utf8Benchmark.bytesToStringOkio       ascii        20  thrpt    5  23529437.325 ±  917328.408  ops/s
Utf8Benchmark.bytesToStringOkio        utf8        20  thrpt    5  12866318.977 ±  594794.057  ops/s
Utf8Benchmark.bytesToStringOkio      sparse        20  thrpt    5  18402101.386 ±  271219.689  ops/s
Utf8Benchmark.stringToBytesJava       ascii        20  thrpt    5  17729707.485 ±  437001.196  ops/s
Utf8Benchmark.stringToBytesJava        utf8        20  thrpt    5  10255343.590 ±  765155.780  ops/s
Utf8Benchmark.stringToBytesJava      sparse        20  thrpt    5  14185915.304 ±  168507.019  ops/s
Utf8Benchmark.stringToBytesOkio       ascii        20  thrpt    5  24924806.444 ± 2024211.052  ops/s
Utf8Benchmark.stringToBytesOkio        utf8        20  thrpt    5  12437156.508 ±  299953.331  ops/s
Utf8Benchmark.stringToBytesOkio      sparse        20  thrpt    5  19959764.792 ±  315293.795  ops/s
```

I'm just going to leave this here and hopefully someone else can confirm what I'm seeing after this gets merged.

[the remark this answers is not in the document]
swankjesse: Faster than Java across the board? I'll take it!

okio/src/main/kotlin/okio/Utf8.kt
bnorm: I've added a lengthy - probably too much so - section on decoding of UTF-8 bytes into a code point. Hopefully it clears up the mask + xor calculations used in these process functions.

bnorm:
Pending further review and a rebase, I believe this to be ready for merge!

okio/native/src/main/kotlin/okio/-Platform.kt
```diff
@@ -46,25 +47,13 @@ internal actual fun arraycopy(
...
-
-// TODO remove if https://youtrack.jetbrains.com/issue/KT-20641 provides a better solution
-actual open class IndexOutOfBoundsException actual constructor(
-  message: String
-) : RuntimeException()
+// We are NOT using the Kotlin/Native function `toUtf8()` because it encodes
+// invalide UTF-16 characters as '\ufffd' instead of '?' like Okio does. In an
+// effort to keep the library consistent with itself, we use the Okio encoder
+// implementation instead. This will hopefully help avoid weird gotcha's as this
+// library starts to be used more across platforms.
```
swankjesse: Agree with the decision & appreciate the documentation.

swankjesse: ... important consequence: Utf8.size() is consistent!

okio/src/test/kotlin/okio/ByteStringTest.kt
```diff
@@ -16,21 +16,77 @@
...
+class CommonByteStringTest {
+  class ByteString : AbstractByteStringTest(ByteStringFactory.BYTE_STRING)
+  class OkioEncoder : AbstractByteStringTest(ByteStringFactory.OKIO_ENCODER)
+}
+
+interface ByteStringFactory {
+  // Unused, but useful if/when SegmentedByteString is moved to common
+  // This also keeps the unit tests the same between Kotlin/JVM for easy
+  // copying as functionality and unit tests are moved
+  fun decodeHex(hex: String): ByteString
```
swankjesse: +1

okio/jvm/src/jmh/java/com/squareup/okio/benchmarks/BufferUtf8Benchmark.java
```diff
@@ -0,0 +1,134 @@
...
+import okio.ByteString;
+
+@Fork(1)
+@Warmup(iterations = 5, time = 2)
+@Measurement(iterations = 5, time = 2)
+@State(Scope.Benchmark)
+@BenchmarkMode(Mode.Throughput)
+@OutputTimeUnit(TimeUnit.SECONDS)
+public class BufferUtf8Benchmark {
+  private static final Map<String, String> strings = new HashMap<>();
```
swankjesse: not important here, but a rule I follow religiously:
https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

okio/jvm/src/main/java/okio/-Platform.kt
```diff
@@ -34,15 +32,9 @@ internal actual fun arraycopy(
...
 
-internal actual fun hashCode(a: ByteArray): Int = Arrays.hashCode(a)
-
 internal actual fun ByteArray.toUtf8String(): String = String(this, Charsets.UTF_8)
 
-internal actual fun CharArray.createString(): String = String(this)
-
 internal actual fun String.asUtf8ToByteArray(): ByteArray = toByteArray(Charsets.UTF_8)
 
 // TODO remove if https://youtrack.jetbrains.com/issue/KT-20641 provides a better solution
```
Egorand: Should this comment be removed?

bnorm: Yes, it should be removed. Looks like I removed it from JS and Native; must have missed this one.

swankjesse:
Oooh, here’s a few that are up for grabs:
https://github.com/square/okio/issues/490
https://github.com/square/okio/issues/506

PR #496, opened 2018-08-25
Add deprecated methods to ease migration for Kotlin callers

swankjesse:
We'll remove these in Okio 2.1 or 2.2.


okio/js/src/main/kotlin/okio/ByteString.kt
```diff
@@ -96,10 +96,10 @@ internal actual constructor(
       commonSubstring(beginIndex, endIndex)
 
   /** Returns the byte at `pos`.  */
-  internal actual open fun getByte(pos: Int) = commonGetByte(pos)
+  internal actual open fun internalGet(pos: Int) = commonGetByte(pos)
```
swankjesse: the other name was colliding with a deprecated bridge method

okio/jvm/src/main/java/okio/OkioUpgrade.kt
```diff
@@ -0,0 +1,123 @@
...
+package okio
+
+import java.io.File
+import java.io.InputStream
+import java.io.OutputStream
+import java.net.Socket
+import java.nio.file.OpenOption
+import java.nio.file.Path
+
+val Okio = OkioUpgrade
```
swankjesse: This is my sneakiest hack . . . callsites say `Okio.source()` but they’re dereferencing a package-level val called Okio, not a class. I’d prefer the class, but anything like that collides with the JVM name of `Okio.kt`, and there’s no `@JvmName` overrides for class names.

swankjesse:
I’m using deprecation *errors* here. Users will need to fix as soon as they switch. I think that’s the best time to pay the cost.

swankjesse:
screen shot 2018-08-25 at 3 20 57 pm

okio/jvm/src/main/java/okio/OkioUpgrade.kt
JakeWharton: We need to prevent Java callers from seeing the val and the object. More `@JvmName`? `@JvmSynthetic`?

JakeWharton: Might have to do it on the file to hide `OkioUpgradeKt` from Java

PR #498, opened 2018-08-26
Hide Base64, ByteStringKt, Platform, and Util code from Java

swankjesse:
These are 'internal' in Kotlin but 'public' to Java callers. Changing
class names to prefix with '-' prevents Java from calling through.


swankjesse:
It works!
```
Task :okio:jvm:compileTestJava warning: [options] bootstrap class path not set in conjunction with -source 1.7 /home/travis/build/square/okio/okio/jvm/src/test/java/okio/GzipSourceTest.java:45: error: cannot find symbol gzipped.writeShort(Util.reverseBytes((short) hcrc.getValue())); // little endian 
```

PR #499, opened 2018-08-29
What asserts shouldBe

jmfayard:
I had a look at okio while learning it was being converted to kotlin, great news!

One little thing that hits me when I read the source file is the junit-style asserts used in the tests. Of course it makes sense because the library has just been converted from java. But I can't help thinking that those got the reading and writing style wrong.

`it shouldBe likeThis`. 

Which is what I've been using for years now with the library [kotlintest](https://github.com/kotlintest/kotlintest). I cherry picked the most useful function from that library and  converted a class so that you can see for yourself

PS: the contributing guide still says to run `mvn clean verify` while the project is using gradle


jmfayard:
Makes sense!

PR #502, opened 2018-08-31
Set up ktlint plugin

Egorand:
[ktlint](https://github.com/shyiko/ktlint) is a linter for Kotlin, similar to checkstyle for Java


build.gradle
```diff
@@ -12,7 +12,8 @@ buildscript {
       'assertj': '1.7.0',
       'shadowPlugin': '2.0.2',
       'nodePlugin': '1.2.0',
-      'mavenPublishPlugin': '0.4.0'
+      'mavenPublishPlugin': '0.4.0',
+      'ktlintPlugin': '5.0.0'
```
JakeWharton: If you add a trailing comma it will future diffs

PR #504, opened 2018-08-31
Fix ktlint issues in platform modules

Egorand:


Egorand:
ktlint Gradle plugin doesn't yet support Kotlin/Native modules in multiplatform projects: https://github.com/JLLeitschuh/ktlint-gradle/issues/119

Egorand:
Agree, they're a little messy, but I wouldn't want to disable the checks globally - there's only a small number of cases where we justify violations.

PR #508, opened 2018-09-07
Embed ProGuard rules in the jar

gabrielittner:
I think the blanket `-dontwarn okio.**` isn't necessary anymore because the `java.nio.file` APIs are available on Android since API 26. The only remaining warnings I'm seeing are for annotations.


okio/jvm/src/main/resources/META-INF/proguard/okio.pro
```diff
@@ -0,0 +1,5 @@
+# JSR 305 annotations are for embedding nullability information.
+-dontwarn javax.annotation.**
```
hzsweers: These aren't used in Okio anymore right?

gabrielittner: I just checked the dependencies and saw it there https://github.com/square/okio/blob/master/okio/jvm/build.gradle#L55

gabrielittner: GitHub search doesn't find any usages though so I'll remove it.

ellet0:
**Not related to this PR**: 
I couldn't find this section in `README.md` or somewhere else (the Proguard rules for the library), does that mean it's no longer required to use any rules on recent versions of Okio? I use OkHttp, and once I include it, the Proguard Gradle task fails with `It appears you are missing some classes resulting in an incomplete class hierarchy`, I included the OkHttp rules but it says you must also include Okio Proguard rules but I couldn't find them in the recent versions.

So I'm wondering if I missed something.

PR #509, opened 2018-09-08
Change BufferedSource.buffer() to be a val in Kotlin.

swankjesse:
Same for BufferedSink.buffer().

This is because for Kotlin callers these methods shadow the
Source.buffer() and Sink.buffer() methods that have different
behavior. Methods with different behavior shouldn't have the
same names!

One consequence of this is that it will nudge Java users to
change from BufferedSource.buffer() to BufferedSource.getBuffer()
which is slightly worse. I couldn't find a good workaround due
to the lack of support for @JvmName in Kotlin interfaces.

Most users will never need to use either method.

### DIFF
--- a/okio/jvm/src/main/java/okio/RealBufferedSink.kt
+++ b/okio/jvm/src/main/java/okio/RealBufferedSink.kt
@@ -24,10 +24,14 @@ import java.nio.charset.Charset
 internal class RealBufferedSink(
   @JvmField val sink: Sink
 ) : BufferedSink {
-  @JvmField val buffer = Buffer()
+  @JvmField var bufferField = Buffer()
   @JvmField var closed: Boolean = false
 
-  override fun buffer() = buffer
+  @Suppress("OVERRIDE_BY_INLINE") // Prevent internal code from calling the getter.
+  override val buffer: Buffer
+    inline get() = bufferField
+
+  override fun buffer() = bufferField
 
   override fun write(source: Buffer, byteCount: Long) {
     check(!closed) { "closed" }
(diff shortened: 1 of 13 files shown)


okio/jvm/src/main/java/okio/RealBufferedSink.kt
```diff
@@ -24,10 +24,14 @@ import java.nio.charset.Charset
 internal class RealBufferedSink(
   @JvmField val sink: Sink
 ) : BufferedSink {
-  @JvmField val buffer = Buffer()
+  @JvmField var bufferField = Buffer()
```
JakeWharton: Why `var`?

okio/jvm/src/main/java/okio/RealBufferedSource.kt
```diff
@@ -24,10 +24,14 @@ import java.nio.charset.Charset
 internal class RealBufferedSource(
   @JvmField val source: Source
 ) : BufferedSource {
+  @JvmField var bufferField = Buffer()
```
JakeWharton: Same question

okio/jvm/src/main/java/okio/RealBufferedSink.kt
swankjesse: Just a mistake. Fixed to val.

okio/jvm/src/main/java/okio/RealBufferedSource.kt
swankjesse: Same answer!

PR #511, opened 2018-09-14
Add a method to BufferedSource for peeking at buffered data

### LINKED ISSUE #506: BufferedSource.peek()
We’ve discussed an idea for a fancy peek API. Here’s an API sketch:

```
interface BufferedSource {
  /**
   * Returns a new `BufferedSource` that can read data from this `BufferedSource`
   * without consuming it. The returned source becomes invalid once this source
   * is next read or closed.
   *
   * For example, we can use `peek()` to lookahead and read the same data multiple
   * times.
   * 
   * ```
   * val buffer = Buffer()
   * buffer.writeUtf8("abcdefghij")
   *
   * buffer.readUtf8(3) // returns "abc", buffer contains "defghi"
   *
   * val peek = buffer.peek()
   * peek.readUtf8(3) // returns "def", buffer contains "defghi"
   * peek.readUtf8(3) // returns "ghi", buffer contains "defghi"
   *
   * buffer.readUtf8() // returns "def", buffer contains "ghi"
   * ```
   */
  BufferedSource peek()
}
```

Behind the scenes the peeking source would copy data from the upstream buffer into its own private buffer.

bnorm:
Closes #506 

Create a new method on BufferedSource, peek(), that creates another
BufferedSource which copies and expands the buffered data of the
upstream source as needed but does not consume it. This allows for look
ahead reading of the data without consumption.


okio/jvm/src/main/java/okio/BufferedSource.kt
```diff
@@ -572,6 +572,27 @@ interface BufferedSource : Source, ReadableByteChannel {
...
+   * buffer.readUtf8(3) // returns "abc", buffer contains "defghi"
+   *
+   * val peek = buffer.peek()
+   * peek.readUtf8(3) // returns "def", buffer contains "defghi"
+   * peek.readUtf8(3) // returns "ghi", buffer contains "defghi"
+   *
+   * buffer.readUtf8(3) // returns "def", buffer contains "ghi"
+   * ```
+   */
+  fun peek(): BufferedSource
```
bnorm: Since buffering is so easy, maybe this should return `Source` for a better understanding of the invalid-state behavior?

okio/jvm/src/main/java/okio/PeekSource.kt
```diff
@@ -0,0 +1,70 @@
...
+      // Source becomes invalid if there is an expected Segment and it and the expected position
+      // do not match the current head and head position of the upstream buffer
+      check(expectedSegment == buffer.head && expectedPos == (buffer.head?.pos ?: -1)) { "invalid" }
+    }
+
+    upstream.request(pos + byteCount)
+    if (expectedSegment == null && buffer.head != null) {
+      // Only once the buffer actually holds data should an expected Segment and position be
+      // recorded. This allows reads from the peek source to repeatedly return -1 and for data to be
+      // added later. Unit tests depend on this behavior.
```
bnorm: I'm not sure I like this behavior, but it did allow using a PeekSource as a parameter value for the BufferedSourceTest. Which was really nice for testing.

okio/jvm/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -140,6 +140,42 @@
       }
     };
 
+    Factory PEEK_BUFFER = new Factory() {
```
bnorm: One cool side-effect here is now there are test cases for peeks-of-peeks. Peek-ception!

okio/jvm/src/main/java/okio/BufferedSource.kt
JakeWharton: I have a hard time seeing anyone doing anything but buffering the source as soon as they get it.

okio/jvm/src/main/java/okio/PeekSource.kt
```diff
@@ -0,0 +1,70 @@
...
+  private var expectedSegment = buffer.head
+  private var expectedPos = buffer.head?.pos ?: -1
+
+  private var pos = 0L
+
+  override fun read(sink: Buffer, byteCount: Long): Long {
+    if (expectedSegment != null) {
+      // Source becomes invalid if there is an expected Segment and it and the expected position
+      // do not match the current head and head position of the upstream buffer
+      check(expectedSegment == buffer.head && expectedPos == (buffer.head?.pos ?: -1)) { "invalid" }
```
JakeWharton: nit: `===` for reference equality on segment comparison (not that it changes behavior, you're going to do this anyway but this gets there faster)

JakeWharton: Can we put a better message here? Maybe "Peeked source is invalid because original source was used"

okio/jvm/src/main/java/okio/PeekSource.kt
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    upstream.request(pos + byteCount)
+    if (expectedSegment == null && buffer.head != null) {
+      // Only once the buffer actually holds data should an expected Segment and position be
+      // recorded. This allows reads from the peek source to repeatedly return -1 and for data to be
+      // added later. Unit tests depend on this behavior.
+      expectedSegment = buffer.head
+      expectedPos = buffer.head?.pos ?: -1
```
JakeWharton: Seems like it should be `buffer.head!!.pos`

okio/jvm/src/main/java/okio/PeekSource.kt
```diff
@@ -0,0 +1,70 @@
...
+    pos += toCopy
+    return toCopy
+  }
+
+  override fun timeout(): Timeout {
+    return upstream.timeout()
+  }
+
+  override fun close() {
+    // Closed when upstream is closed
```
JakeWharton: Hmm we still might want to track this state and throw in `read()`

okio/jvm/src/test/java/okio/BufferedSourceTest.java
```diff
@@ -1231,6 +1269,62 @@ private void assertLongDecimalString(String s, long expected) throws IOException
...
+    assertEquals("ghi", peek2.readUtf8(3));
+    assertFalse(peek2.request(1));
+
+    assertEquals("ghi", peek1.readUtf8(3));
+    assertFalse(peek1.request(1));
+
+    assertEquals("def", source.readUtf8(3));
+  }
+
+  @Test public void peekInvalid() throws IOException {
```
JakeWharton: I think we probably want a test that consumes at least a `Segment.SIZE` worth of bytes as well, just to be sure.

okio/jvm/src/main/java/okio/PeekSource.kt
bnorm: I thought about that after I submitted the pull-request. I will add.

JakeWharton:
This makes me wonder if a `peek(limit: Long)` would be a useful addition. I don't think we need to add it right away, but I _think_ I can see potential uses for it.

PR #515, opened 2018-09-26
Create a Throttler sample

### LINKED ISSUE #490: Throttler okio source implement?

bnorm:
Closes #490 

Created a sample Throttler which can throttle Sources and Sinks to a
desired throughput. Multiple Sources and Sinks can be attached to the
same Throttler and their combined throughput will not exceed the desired
throughput. Multiple Throttlers can also be used on the same Source or
Sink and they will all be honored.

Added some basic unit tests to verify desired behavior.

### DIFF
--- a/okio/jvm/src/main/java/okio/Throttler.kt
+++ b/okio/jvm/src/main/java/okio/Throttler.kt
@@ -1,0 +1,151 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import java.io.IOException
+import java.io.InterruptedIOException
+
+/**
+ * Enables limiting of Source and Sink throughput. Attach to this throttler via [.source] and
+ * [.sink] and set the desired throughput via [.bytesPerSecond]. Multiple Sources and Sinks can be
+ * attached to a single Throttler and they will be throttled as a group, where their combined
+ * throughput will not exceed the desired throughput. The same Source or Sink can be attached to
+ * multiple Throttlers and its throughput will not exceed the desired throughput of any of the
+ * Throttlers.
+ *
+ * This class has these tuning parameters:
+ *
+ *  * `bytesPerSecond`: Maximum sustained throughput. Use 0 for no limit.
+ *  * `waitByteCount`: When the requested byte count is greater than this many bytes and isn't
+ *    immediately available, only wait until we can allocate at least this many bytes. Use this to
+ *    set the ideal byte count during sustained throughput.
+ *  * `maxByteCount`: Maximum number of bytes to allocate on any call. This is also the number of
+ *    bytes that will be returned before any waiting.
+ */
+class Throttler {
+  private var bytesPerSecond: Long = 0L
+  private var waitByteCount: Long = 8 * 1024 // 8 KiB.
+  private var maxByteCount: Long = 256 * 1024 // 256 KiB.
+  private var nanosForMaxByteCount: Long = -1L
+
+  /**
+   * The nanoTime that we've consumed all bytes through. This is never greater than the current
+   * nanoTime plus nanosForMaxByteCount.
+   */
+  private var allocatedUntil: Long = System.nanoTime()
+
+  /** Sets the rate at which bytes will be allocated. Use 0 for no limit. */
+  @JvmOverloads
+  fun bytesPerSecond(
+    bytesPerSecond: Long,
+    waitByteCount: Long = this.waitByteCount,
+    maxByteCount: Long = this.maxByteCount
+  ) {
+    synchronized(this) {
+      require(bytesPerSecond >= 0)
+      require(waitByteCount > 0)
+      require(maxByteCount >= waitByteCount)
+
+      this.bytesPerSecond = bytesPerSecond
+      this.waitByteCount = waitByteCount
+      this.maxByteCount = maxByteCount
+      this.nanosForMaxByteCount = when {
+        bytesPerSecond != 0L -> maxByteCount * 1_000_000_000L / bytesPerSecond
+        else -> -1L
+      }
+
+      (this as Object).notifyAll()
+    }
+  }
+
+  /**
+   * Take up to `byteCount` bytes, waiting if necessary. Returns the number of bytes that were
+   * taken.
+   */
+  internal fun take(byteCount: Long): Long {
+    require(byteCount > 0)
+
+    synchronized(this) {
+      while (true) {
+        if (bytesPerSecond == 0L) return byteCount // No limits.
+
+        val now = System.nanoTime()
+        val idleInNanos = maxOf(allocatedUntil - now, 0L)
+        val usableNanos = nanosForMaxByteCount - idleInNanos
+        val immediateBytes = bytesPerSecond * usableNanos / 1_000_000_000L
+
+        // Fulfill the entire request without waiting.
+        if (immediateBytes >= byteCount) {
+          val byteCountNanos = byteCount * 1_000_000_000L / bytesPerSecond
+          allocatedUntil = now + idleInNanos + byteCountNanos
+          return byteCount
+        }
+
+        // Fulfill a big-enough block without waiting.
+        if (immediateBytes >= waitByteCount) {
+          allocatedUntil = now + idleInNanos + usableNanos
+          return immediateBytes
+        }
+
+        // Wait until we can write some bytes.
+        val byteCountNanos = minOf(waitByteCount, byteCount) * 1_000_000_000L / bytesPerSecond
+        waitNanos(byteCountNanos - usableNanos)
+      }
+    }
+    throw AssertionError() // Unreachable, but synchronized() doesn't know that.
+  }
+
+  private fun waitNanos(nanosToWait: Long) {
+    val millisToWait = nanosToWait / 1_000_000L
+    val remainderNanos = nanosToWait - (millisToWait * 1_000_000L)
+    (this as Object).wait(millisToWait, remainderNanos.toInt())
+  }
+
+  /** Create a Source which honors this Throttler.  */
+  fun source(source: Source): Source {
+    return object : ForwardingSource(source) {
+      override fun read(sink: Buffer, byteCount: Long): Long {
+        try {
+          val toRead = take(byteCount)
+          return super.read(sink, toRead)
+        } catch (e: InterruptedException) {
+          Thread.currentThread().interrupt()
+          throw InterruptedIOException("interrupted")
+        }
+      }
+    }
+  }
+
+  /** Create a Sink which honors this Throttler.  */
+  fun sink(sink: Sink): Sink {
+    return object : ForwardingSink(sink) {
+      @Throws(IOException::class)
+      override fun write(source: Buffer, byteCount: Long) {
+        try {
+          var remaining = byteCount
+          while (remaining > 0L) {
+            val toWrite = take(remaining)
+            super.write(source, toWrite)
+            remaining -= toWrite
+          }
+        } catch (e: InterruptedException) {
+          Thread.currentThread().interrupt()
+          throw InterruptedIOException("interrupted")
+        }
+      }
+    }
+  }
+}
--- a/okio/jvm/src/test/java/okio/ThrottlerTest.kt
+++ b/okio/jvm/src/test/java/okio/ThrottlerTest.kt
@@ -1,0 +1,271 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import okio.TestUtil.randomSource
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import java.util.concurrent.Executors
+import kotlin.test.assertEquals
+
+class ThrottlerTest {
+  private val size = 1024L * 80L // 80 KiB
+  private val source = randomSource(size)
+
+  private val throttler = Throttler()
+  private val throttlerSlow = Throttler()
+
+  private val threads = 4
+  private val executorService = Executors.newFixedThreadPool(threads)
+  private var stopwatch = Stopwatch()
+
+  @Before fun setup() {
+    throttler.bytesPerSecond(4 * size, 4096, 8192)
+    throttlerSlow.bytesPerSecond(2 * size, 4096, 8192)
+    stopwatch = Stopwatch()
+  }
+
+  @After fun teardown() {
+    executorService.shutdown()
+  }
+
+  @Test fun sourceByteCount() {
+    throttler.bytesPerSecond(bytesPerSecond = 20, waitByteCount = 5, maxByteCount = 10)
+    val source = throttler.source(source)
+    val buffer = Buffer()
+
+    // We get the first 10 bytes immediately (that's maxByteCount).
+    assertThat(source.read(buffer, size)).isEqualTo(10)
+    stopwatch.assertElapsed(0.0)
+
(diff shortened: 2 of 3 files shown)


samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,107 @@
...
+ * <p>To clear any throttling, use {@link #clear()}.
+ */
+public final class Throttler {
+
+  private double limit = 0.0d;
+  private long last = System.currentTimeMillis();
+
+  /** Sets the desired max throughput. */
+  public void bytesPerTimePeriod(long byteCount, int duration, TimeUnit unit) {
+    Objects.requireNonNull(unit);
```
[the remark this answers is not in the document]
bnorm: My goal was to keep it Java 6 complaint but it slipped my mind `Objects` wasn't introduced until Java 7. Will remove. And convert `duration` to `long`.

samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+
+  /** Create a Source which honors this Throttler. */
+  public Source source(Source source) {
+    return new ForwardingSource(source) {
+      @Override
+      public long read(Buffer sink, long byteCount) throws IOException {
+        long read = super.read(sink, byteCount);
+        if (read > 0) delay(read);
```
bnorm: One thing I want to point out is that we are delaying after the read here because I thought it best to delay based on the amount of data that is actually read, not the requested byte count. For the Sink, we are delaying before the write so there is a little bit of an inconsistency. I can change either or neither, but I wanted to point it out specifically.

samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,110 @@
...
+ * desired throughput of any of the Throttlers.
+ *
+ * <p>To clear any throttling, use {@link #clear()} or specify a <code>byteCount</code> of <code>
+ * 0</code> when setting the throughput.
+ */
+public final class Throttler {
+  private static final int NANOS_PER_MILLI = 1000000;
+
+  private double limit = 0.0d;
+  private long last = System.nanoTime();
```
NightlyNexus: should this be initialized to some sentinel value like -1, so the first read or write is never delayed?

samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,138 @@
...
+
+  /** Sets the desired max throughput. <code>byteCount</code> of zero to clear. */
+  public void bytesPerTimePeriod(long byteCount, long duration, TimeUnit unit) {
+    if (unit == null || byteCount < 0 || duration <= 0) {
+      throw new IllegalArgumentException(
+          "byteCount=" + byteCount + " duration=" + duration + " unit=" + unit);
+    }
+
+    synchronized (this) {
+      this.limit = (double) byteCount / unit.toNanos(duration);
```
swankjesse: Nit: I don’t think this is what we want. I think want to actually implement the limit provided and not spread the limit out over the narrowest possible time unit.

In particular, if the user asks for 60 bytes per minute, I want us to:
Send 60 bytes, then sleep 60 seconds, vs.
Send 1 byte, sleep 1 second. 60 times.

samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+  }
+
+  /** Clears the max throughput. */
+  public void clear() {
+    synchronized (this) {
+      this.limit = 0.0d;
```
swankjesse: notifyAll ?

samples/src/main/java/okio/samples/Throttler.java
swankjesse: notifyAll after this?

samples/src/main/java/okio/samples/Throttler.java
```diff
@@ -0,0 +1,138 @@
...
+      try {
+        if (delay > 0) {
+          Thread.sleep(delay / NANOS_PER_MILLI, (int) (delay % NANOS_PER_MILLI));
+        }
+      } catch (InterruptedException e) {
+        Thread.currentThread().interrupt();
+        throw new InterruptedIOException("interrupted");
+      }
+
+      last = System.nanoTime();
```
swankjesse: I don’t think we should build up credits if bandwidth is unused.

If you say 60 bytes per second and you have a 5 minute period where you don’t do anything, I don’t think you get a free 300 bytes to send.

samples/src/main/java/okio/samples/Throttler.java
swankjesse: Thanks for calling this out. I think it’s a nice design but I don’t think it’s what we want!

In particular we should assume somebody could be using the throttler to avoid consuming some expensive resource. Perhaps they’re sharing a narrow connection and they want to throttle one user to prioritize another.

swankjesse:
... Thanks for building this, and thanks for thinking through lots of hard problems!

As usual, I want to do this, but bigger and more ambitious. I’d like to include this as a proper Okio API (not a sample) and I’d like to figure out all of the tricks and catches to make it work absolutely great.

Lets figure out exactly what behavior we want. I like “leaky bucket” as the abstraction for features like this, though perhaps there are others to study.

https://en.wikipedia.org/wiki/Leaky_bucket

swankjesse:
FWIW, if we get this working nicely I’d like to put it to use in Misk! We can throttle sockets that send us too much data in order to prevent other sockets from being starved out.

samples/src/main/java/okio/samples/Throttler.java
bnorm: There shouldn't be a need. This is holding the synchronized lock so no Sources or Sinks should be reading or writing.

samples/src/main/java/okio/samples/Throttler.java
bnorm: There shouldn't be a need. This is holding the synchronized lock so no Sources or Sinks should be reading or writing.

samples/src/main/java/okio/samples/Throttler.java
swankjesse: Ahh. . .  you wanna change `delay()` to call `wait()` instead of `Thread.sleep()` because the former can be woken up early with a call to `notifyAll()`. And you wanna wake it up early if more bandwidth becomes available while it’s sleeping.

samples/src/main/java/okio/samples/Throttler.java
bnorm: I use the limit to determine how long a request of a specific `byteCount` should take.

In your example, if the user asks for 180 bytes, the request should take 180 seconds. If the last request was over 180 seconds ago, there will be no waiting, but if it is less, we'll wait for the difference. Then reset the time and process the next request.

samples/src/main/java/okio/samples/Throttler.java
bnorm: Using `wait()` actually breaks the entire implementation because it would release the synchronization lock. I want the Threads to be stuck on the synchronization, I'm using it like a queue. All requests will get queued by the synchronization block and each request will be delayed as much as necessary depending on how long ago the last request was. 

In terms of the Leaky Bucket algorithm, think of the synchronization lock as the bucket and the leak is the `sleep()` and the `last` variable is the inverse of how full the bucket is.

I'm hoping the graphs that I created in the `delay()` method shed some light on how the implementation works in practice. Maybe I can get more detailed or add some descriptions to each example.

swankjesse:
For sources we’re definitely within our rights to return fewer bytes than the caller requested.

For sinks we can’t return until everything is done, but we should loop internally, writing fragments of the source data as we go. If the limit is 7 bytes per minute and we’re given 35 bytes, we should alternately write 7 bytes and wait 60 seconds until it’s all gone.

I’m not exactly sure the nicest way to handle the state of having consumed some of the allocation but not all of it. Tempting to see what Gauva does too.
https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/RateLimiter.java

swankjesse:
BTW this doc is awesome.
https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/SmoothRateLimiter.java

Tolriq:
Sorry to hijack but since I was in the opened issue I permit to talk about a related need as @swankjesse talks about seeing it bigger :p

For easier user configuration it would be really nice (even if probably out of the scope) to be able to handle percent of max BW.

This is what most users asked me, to throttle download to still have enough BW to be able to do other things at the same time. Fixed rates are problematic as user could require a different limit on 3G / 4G / 5G / Wifi ...

Being able to set limit rate to X% with a sample rate duration (duration allowed at start to burst to find the current max BW) would be a real nice addition.

swankjesse:
@bnorm I think your API is perfect but I want to build the production version of this to smooth out big requests. Here’s a work-in-progress of what I’m thinking.

https://github.com/square/okio/compare/master...jwilson.1006.bandwidthcap?expand=1#files_bucket

The biggest difference is that it returns before the full request is fulfilled. When this case it tells you how many bytes you got. I’m also using wait/notify so it should be possible to immediately respond to a rate change.

Next steps – the code above is hasty. I’d like this thing to have your API and pass your tests. Would you like to take it from here? (I’m happy to do so otherwise!)

bnorm:
I'm happy to take it from here. Your example helped clarify some things I had questions on so thank you for that. But I do have some followup questions, mostly logistical ones.

1. Would you like to have a separate class which manages the bandwidth limiting, similar to what you created, or would you like it all rolled up into a single class? I can see the Source and Sink hiding enough of the details that it might make it hard to test the bandwidth limiting behavior thoroughly so I'm a little torn between the options.

2. What's your thought on the API for setting the bandwidth limit and min and max transfer sizes? `bytesPerTimePeriod` is a little wordy for my taste but I stuck with it for the sake of the sample. Now that you want this part of the official API, I'd like to take a closer look at the control functions for the `Throttler` class.

3. What if the read size requested for a Source is smaller than the minimum transfer size? This thought makes me think we need to perform some kind of buffering of the Source? I'll play around with ideas, but I'd like to know what your thoughts.

I'll start with those questions as I start working on integrating your work.

bnorm:
@Tolriq this is an interesting request, but I'm not sure how well this fits into the API of the Throttler. You could definitely use the Throttler as part of the solution to your problem, but the measuring part seems to me like another entity. This other entity, `Measurer` for example, could keep track of the avg/min/max of things like bandwidth, latency, IOPS, etc, and be used to determine how much throttling should be put in place.

Tolriq:
The issue is about how to link things.

One could measure max BW then create a new source with the throttler, but when it's tied to a download on a non resumable source for example, This means double download of the needed data to measure it.

Don't know how dynamic the throttler could be but this request may require some API entry points to enable that if it's in core.

But yes a way to measure BW via listener + dynamic throttler could be a way. Just wanted to be sure the dynamic part was possible.

swankjesse:
> Would you like to have a separate class which manages the bandwidth limiting, similar to what you created, or would you like it all rolled up into a single class?

I like your API. A top-level `Throttler` class that can act as a factory for sources and sinks. I think it’s clever that multiple streams can all share a single throughput budget, and want to keep that.

Might be worthwhile to have a package-private method like `take()` for testing, but this need not be public API.

> What's your thought on the API for setting the bandwidth limit and min and max transfer sizes? 

Though I generally prefer `(long duration, TimeUnit unit)` APIs, I don’t think rate limits really apply to any unit other than bytes per second. Does this make me a heretic?! I’m just thinking about all of the bandwidth measurements I‘ve ever heard, and they’re always either bits per second or bytes per second.

So my preference is a 3 argument API accepting bytesPerSecond, minBlock, and maxBlock. I think both the minBlock and maxBlock abstractions are useful.

> What if the read size requested for a Source is smaller than the minimum transfer size?

I think the minBlock should only apply if it’s smaller than the requested size. If your minBlock is 1024 and you request 512, we only wait for 512. But if you request 2048, we’ll only wait for 1024.


Strawman API:

```
class Throttler { 
   fun bytesPerSecond(
      bytesPerSecond: Long, 
      minBlock: Long = 8 * 1024L, 
      maxBlock: Long = 256 * 1024L)

   fun source(Source source):  Source

   fun sink(Sink sink): Sink 
}
```

bnorm:
Turns out integrating your code was easier than expected. Took me longer than I would like to admit to figure out what `minTake` actually means. I think it was confusing because it's only a minimum value in specific situations.

I tried a new naming for `minTake` and `maxTake` let me know what you think. Documentation is not a strong suit of mine so happy to take any suggestions, even if it seems nitpicky.

okio/jvm/src/main/java/okio/Throttler.kt
```diff
@@ -0,0 +1,151 @@
...
+ * [.sink] and set the desired throughput via [.bytesPerSecond]. Multiple Sources and Sinks can be
+ * attached to a single Throttler and they will be throttled as a group, where their combined
+ * throughput will not exceed the desired throughput. The same Source or Sink can be attached to
+ * multiple Throttlers and its throughput will not exceed the desired throughput of any of the
+ * Throttlers.
+ *
+ * This class has these tuning parameters:
+ *
+ *  * `bytesPerSecond`: maximum sustained throughput.
+ *  * `waitByteCount`: when the requested byte count is greater than this many bytes and isn't
```
swankjesse: Dig.

okio/jvm/src/main/java/okio/Throttler.kt
```diff
@@ -0,0 +1,151 @@
...
+ * Enables limiting of Source and Sink throughput. Attach to this throttler via [.source] and
+ * [.sink] and set the desired throughput via [.bytesPerSecond]. Multiple Sources and Sinks can be
+ * attached to a single Throttler and they will be throttled as a group, where their combined
+ * throughput will not exceed the desired throughput. The same Source or Sink can be attached to
+ * multiple Throttlers and its throughput will not exceed the desired throughput of any of the
+ * Throttlers.
+ *
+ * This class has these tuning parameters:
+ *
+ *  * `bytesPerSecond`: maximum sustained throughput.
```
swankjesse: Specify that callers should use 0 for no limit?

okio/jvm/src/main/java/okio/Throttler.kt
```diff
@@ -0,0 +1,151 @@
...
+  /** Create a Source which honors this Throttler.  */
+  fun source(source: Source): Source {
+    return object : ForwardingSource(source) {
+      override fun read(sink: Buffer, byteCount: Long): Long {
+        try {
+          val toRead = take(byteCount)
+          return super.read(sink, toRead)
+        } catch (e: InterruptedException) {
+          Thread.currentThread().interrupt()
+          throw InterruptedIOException("interrupted")
```
swankjesse: Nice.

okio/jvm/src/main/java/okio/Throttler.kt
```diff
@@ -0,0 +1,151 @@
...
+  fun sink(sink: Sink): Sink {
+    return object : ForwardingSink(sink) {
+      @Throws(IOException::class)
+      override fun write(source: Buffer, byteCount: Long) {
+        try {
+          var remaining = byteCount
+          while (remaining > 0L) {
+            val toWrite = take(remaining)
+            super.write(source, toWrite)
+            remaining -= toWrite
```
swankjesse: 👍🏻

okio/jvm/src/test/java/okio/ThrottlerTest.kt
```diff
@@ -0,0 +1,271 @@
...
+
+    // Wait a quarter second for each subsequent 5 bytes (that's waitByteCount).
+    assertThat(source.read(buffer, size)).isEqualTo(5)
+    stopwatch.assertElapsed(0.25)
+
+    assertThat(source.read(buffer, size)).isEqualTo(5)
+    stopwatch.assertElapsed(0.25)
+
+    // Wait three quarters of a second to build up 15 bytes of potential.
+    // Since maxByteCount = 10, there will only be 10 bytes of potential.
```
swankjesse: 👍🏻

okio/jvm/src/test/java/okio/ThrottlerTest.kt
```diff
@@ -0,0 +1,271 @@
...
+    }
+    Thread.sleep(250)
+    throttler.bytesPerSecond(0)
+    for (future in futures) {
+      future.get()
+    }
+    stopwatch.assertElapsed(0.250)
+  }
+
+  @Test fun parallelFastThenSlower() {
```
swankjesse: Neat!

swankjesse:
One small request: can you double the durations (by halving the rate or doubling the data) ? We've had problems in Travis CI where timing tests flake due to us trying to measure too finely.

okio/jvm/src/main/java/okio/Throttler.kt
```diff
@@ -0,0 +1,151 @@
...
+ * attached to a single Throttler and they will be throttled as a group, where their combined
+ * throughput will not exceed the desired throughput. The same Source or Sink can be attached to
+ * multiple Throttlers and its throughput will not exceed the desired throughput of any of the
+ * Throttlers.
+ *
+ * This class has these tuning parameters:
+ *
+ *  * `bytesPerSecond`: Maximum sustained throughput. Use 0 for no limit.
+ *  * `waitByteCount`: When the requested byte count is greater than this many bytes and isn't
+ *    available, wait until we can allocate at least this many bytes. Use this to avoid repeated
```
bnorm: Instead of `Use this to avoid...`, what about `Use this to set the ideal byte count during sustained throughput.`?

swankjesse: 👍🏻

swankjesse:
# 🐢

This is great!

PR #517, opened 2018-09-30
AsyncTimeout wrapped Source needs to timeout on long close

bnorm:
The wrapping Source for an AsyncTimeout was missing a call to enter() in
close(). This means a timeout would never occur if closing the Source
took longer than then the current timeout.

Added some unit tests to cover this case as well for both Source and
Sink and close() and flush().

### DIFF
--- a/okio/jvm/src/main/java/okio/AsyncTimeout.kt
+++ b/okio/jvm/src/main/java/okio/AsyncTimeout.kt
@@ -168,6 +168,7 @@ open class AsyncTimeout : Timeout() {
 
       override fun close() {
         var throwOnTimeout = false
+        enter()
         try {
           source.close()
           throwOnTimeout = true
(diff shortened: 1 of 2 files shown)


okio/jvm/src/main/java/okio/AsyncTimeout.kt
```diff
@@ -168,6 +168,7 @@ open class AsyncTimeout : Timeout() {
 
       override fun close() {
         var throwOnTimeout = false
+        enter()
```
swankjesse: How long was this missing?!

bnorm: Git history suggests forever?
https://github.com/square/okio/commits/okio_1x/okio/src/main/java/okio/AsyncTimeout.java
https://github.com/square/okio/commit/5880e8947d65b5bdf81ea6f56aff7f5e766c5ed7#diff-cb1656dc5e360018922fbd4b9a342b70R204

Maybe worth a patch back to `okio_1x`?

swankjesse: Sure, though no release is currently planned.

PR #520, opened 2018-10-08
Backport BufferedSource.peek() and getBuffer() from Okio 2.1.

swankjesse:
I want to use these features in Moshi and I'm not ready to force Moshi users
to Okio 2.x.


okio/src/main/java/okio/PeekSource.java
```diff
@@ -0,0 +1,80 @@
...
+ * A {@link Source} which peeks into an upstream {@link BufferedSource} and allows reading and
+ * expanding of the buffered data without consuming it. Does this by requesting additional data from
+ * the upstream source if needed and copying out of the internal buffer of the upstream source if
+ * possible.
+ *
+ * <p>This source also maintains a snapshot of the starting location of the upstream buffer which it
+ * validates against on every read. If the upstream buffer is read from, this source will become
+ * invalid and throw {@link IllegalStateException} on any future reads.
+ */
+class PeekSource implements Source {
```
JakeWharton: final

okio/src/main/java/okio/PeekSource.java
```diff
@@ -0,0 +1,80 @@
...
+    long toCopy = Math.min(byteCount, buffer.size - pos);
+    if (toCopy <= 0L) return -1L;
+
+    buffer.copyTo(sink, pos, toCopy);
+    pos += toCopy;
+    return toCopy;
+  }
+
+  @Override public Timeout timeout() {
+    return upstream.timeout();
```
JakeWharton: Could be using `FowardingSource` I suppose, but for little gain

okio/src/main/java/okio/BufferedSource.java
```diff
@@ -533,6 +541,27 @@
   boolean rangeEquals(long offset, ByteString bytes, int bytesOffset, int byteCount)
       throws IOException;
 
+  /**
+   * Returns a new `BufferedSource` that can read data from this `BufferedSource` without consuming
```
hzsweers: These docs are still with kdoc formatting, is that intentional?

okio/src/main/java/okio/PeekSource.java
swankjesse: Yeah, and then it’s a bit awkward because we don’t want close() to cascade.

okio/src/main/java/okio/BufferedSource.java
swankjesse: Fixed!

okio/src/main/java/okio/PeekSource.java
swankjesse: Fixed!

hzsweers: :+1:

Issue #514, opened 2018-09-26
Buffer.truncate(size)


bnorm:
I found myself in a situation today where it was more efficient to write extra data (in the form of SegmentedByteStrings) to a buffer and truncate the buffer. Currently, I'm doing it this way:

```
long oldSize = sink.size();
long toWrite = Math.min(byteCount, SEGMENT_SIZE);
// ... maybe other writes
sink.write(segmentTemplate); // might overfill the buffer
sink.readAndWriteUnsafe(cursor);
try {
    cursor.resizeBuffer(oldSize + toWrite);
} finally {
    cursor.close();
}
```

This would also be alleviated if there was better handling of ByteStrings substrings. Either make creating substrings of a SegementedByteString share the underlying segments or add a BufferedSink.write method which allows for a begin and end index when writing a ByteString. Probably all good things regardless?

swankjesse:
As a general rule I prefer all Buffer changes to be either reading from the front or writing to the end. You can cheat to truncate like so:

```
Buffer needsToTruncate = ...
long truncatedLength = ...;
Buffer temp = new Buffer();
temp.write(needsToTruncate, truncatedLength);
needsToTruncate.clear();
needsToTruncate.writeAll(temp);
```

I wonder which of the two variants will perform better. One of ’em might share segments better than the other.

swankjesse:
Yeah, my preference is to keep the conceptual model as “read front, write back” and optimize performance for all combinations of that. If we offer APIs to do things like insert bytes in the middle of a buffer those APIs won’t be very fast on our current implementation and I don’t like that.

bnorm:
The write method I am describing would write a substring of ByteString, not to write to the middle of the BufferedSink. Exactly like how `BufferedSink.write(ByteArray, beginIndex, endIndex)` currently works, but for ByteString.

I messed around with a few optimizations for SegmentedByteString. Substring segment sharing, utf8 saving, digest and hmac via segments, etc. Is this a pull-request you would be interested in?

swankjesse:
Ahh, sorry this issue covering other things got me thinking something else. My bad.

Yeah, we definitely should have a method to write a slice of a ByteString. Starting with the basic implementation that copies, and later optimizing for SegmentedByteString!

bnorm:
No worries, I sort of jumped topics pretty quick there. Closing as the original request is not going to be implemented.

PR #524, opened 2018-10-10
Various SegmentedByteString improvements

bnorm:
Inspired after #514, I started taking a look at creating a substring for SegmentedByteString that shared segments and didn't convert to a normal first ByteString. This lead me down a road of a bunch of different optimizations that could be made to SegmentedByteString to simplify and reduce the amount of converting to a single ByteArray.

I've included each improvement as a separate commit to hopefully make reviewing it in isolation easier. I also did this in case you didn't want to include any of them, it should be pretty easy to just rebase out the commit.

### DIFF
--- a/okio/jvm/src/main/java/okio/SegmentedByteString.kt
+++ b/okio/jvm/src/main/java/okio/SegmentedByteString.kt
@@ -19,7 +19,11 @@ import java.io.IOException
 import java.io.OutputStream
 import java.nio.ByteBuffer
 import java.nio.charset.Charset
+import java.security.InvalidKeyException
+import java.security.MessageDigest
 import java.util.Arrays
+import javax.crypto.Mac
+import javax.crypto.spec.SecretKeySpec
 
 /**
  * An immutable byte string composed of segments of byte arrays. This class exists to implement
@@ -97,19 +101,26 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
 
   override fun toAsciiUppercase() = toByteString().toAsciiUppercase()
 
-  override fun md5() = toByteString().md5()
+  override fun digest(algorithm: String): ByteString {
+    val digest = MessageDigest.getInstance(algorithm)
+    processSegments { data, offset, byteCount ->
+      digest.update(data, offset, byteCount)
+    }
+    return ByteString(digest.digest())
+  }
 
-  override fun sha1() = toByteString().sha1()
-
-  override fun sha256() = toByteString().sha256()
-
-  override fun sha512() = toByteString().sha512()
-
-  override fun hmacSha1(key: ByteString) = toByteString().hmacSha1(key)
-
-  override fun hmacSha256(key: ByteString) = toByteString().hmacSha256(key)
-
-  override fun hmacSha512(key: ByteString) = toByteString().hmacSha512(key)
+  override fun hmac(algorithm: String, key: ByteString): ByteString {
+    try {
+      val mac = Mac.getInstance(algorithm)
+      mac.init(SecretKeySpec(key.toByteArray(), algorithm))
+      processSegments { data, offset, byteCount ->
+        mac.update(data, offset, byteCount)
+      }
+      return ByteString(mac.doFinal())
+    } catch (e: InvalidKeyException) {
+      throw IllegalArgumentException(e)
+    }
+  }
 
   override fun base64Url() = toByteString().base64Url()
 
--- a/okio/jvm/src/main/java/okio/SegmentedByteString.kt
+++ b/okio/jvm/src/main/java/okio/SegmentedByteString.kt
@@ -136,17 +136,11 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
   override fun getSize() = directory[segments.size - 1]
 
   override fun toByteArray(): ByteArray {
-    val result = ByteArray(directory[segments.size - 1])
-    var segmentOffset = 0
-    var s = 0
-    val segmentCount = segments.size
-    while (s < segmentCount) {
-      val segmentPos = directory[segmentCount + s]
-      val nextSegmentOffset = directory[s]
-      arraycopy(segments[s], segmentPos, result, segmentOffset,
-          nextSegmentOffset - segmentOffset)
-      segmentOffset = nextSegmentOffset
-      s++
+    val result = ByteArray(size)
+    var resultPos = 0
+    processSegments { data, offset, byteCount ->
+      arraycopy(data, offset, result, resultPos, byteCount)
+      resultPos += byteCount
     }
     return result
   }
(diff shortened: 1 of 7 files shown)


okio/jvm/src/main/java/okio/SegmentedByteString.kt
```diff
@@ -222,40 +226,65 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
...
-      fromIndex)
-
   /** Returns a copy as a non-segmented byte string.  */
   private fun toByteString() = ByteString(toByteArray())
 
   override fun internalArray() = toByteArray()
 
+  /** Processes all segments, invoking `yield` with the ByteArray and range of valid data. */
+  private inline fun processSegments(
+    yield: (data: ByteArray, offset: Int, byteCount: Int) -> Unit
```
JakeWharton: I would prefer these not be called "yield" since Kotlin uses that for suspending callbacks: http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.coroutines.experimental/-sequence-builder/yield.html. This helper is kind of like a fancy `forEach` which uses "action": http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each.html

okio/jvm/src/main/java/okio/SegmentedByteString.kt
```diff
@@ -222,40 +226,65 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
...
-  override fun lastIndexOf(other: ByteArray, fromIndex: Int) = toByteString().lastIndexOf(other,
-      fromIndex)
-
   /** Returns a copy as a non-segmented byte string.  */
   private fun toByteString() = ByteString(toByteArray())
 
   override fun internalArray() = toByteArray()
 
+  /** Processes all segments, invoking `yield` with the ByteArray and range of valid data. */
+  private inline fun processSegments(
```
JakeWharton: Can we eliminate this overload and use default values for `beginIndex` and `endIndex` on the other?

okio/jvm/src/main/java/okio/SegmentedByteString.kt
bnorm: I can change it to `action`. Maybe also change the name to `forEachSegment`?

I will also mention that I used `yield` for the UTF-8 decoding methods in the common module. I could change those to `action` as well but it feels a little out of scope. Maybe a separate pull request?

okio/jvm/src/main/java/okio/SegmentedByteString.kt
bnorm: Certainly could, but I was concerned about the performance impact as the other function does a few extra calculations this one is able to avoid.

I'm working on a performance test which will hopefully verify that we didn't lose performance on any of these functions. I'll also test if this function is required or if the other has similar performance.

okio/jvm/src/main/java/okio/SegmentedByteString.kt
JakeWharton: I'd be happy with `forEachSegment`. I really like how much this cleans up the other code!

okio/jvm/src/main/java/okio/SegmentedByteString.kt
```diff
@@ -99,24 +102,57 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
...
+    }
 
-  override fun base64Url() = toByteString().base64Url()
+    // Set the new position of the first segment
+    val segmentOffset = if (beginSegment == 0) 0 else directory[beginSegment - 1]
+    newDirectory[newSegments.size] += beginIndex - segmentOffset
 
-  override fun substring(beginIndex: Int, endIndex: Int) = toByteString().substring(beginIndex,
-      endIndex)
+    return SegmentedByteString(newSegments, newDirectory)
```
swankjesse: nice

okio/jvm/src/main/java/okio/SegmentedByteString.kt
swankjesse: Thumbs up on `forEachSegment`.

swankjesse: Its functions like this that make me want to convert **everything** to Kotlin

okio/src/main/kotlin/okio/internal/ByteString.kt
```diff
@@ -168,10 +168,42 @@ internal fun ByteString.commonEndsWith(suffix: ByteString) =
 internal fun ByteString.commonEndsWith(suffix: ByteArray) =
     rangeEquals(size - suffix.size, suffix, 0, suffix.size)
 
+internal fun ByteString.commonIndexOf(other: ByteString, fromIndex: Int): Int {
+  val limit = size - other.size
+  for (i in maxOf(fromIndex, 0)..limit) {
```
swankjesse: should this be until and not `..` ?

swankjesse: I almost always want `until` and not `..` because `..` is [inclusive, inclusive] which is rarely used in our code

okio/src/main/kotlin/okio/internal/ByteString.kt
```diff
@@ -168,10 +168,42 @@ internal fun ByteString.commonEndsWith(suffix: ByteString) =
...
+    if (rangeEquals(i, other, 0, other.size)) {
+      return i
+    }
+  }
+  return -1
+}
+
+internal fun ByteString.commonLastIndexOf(other: ByteString, fromIndex: Int): Int {
+  var fromIndex = fromIndex
+  fromIndex = minOf(fromIndex, size - other.size)
```
swankjesse: you can‘t combine this with the previous line?

bnorm: I copied it from the JVM ByteString implementation, but I'll make it more like the `commonIndexOf` functions.

okio/jvm/src/main/java/okio/SegmentedByteString.kt
```diff
@@ -99,24 +102,57 @@ internal class SegmentedByteString(buffer: Buffer, byteCount: Int) : ByteString(
...
+    }
 
-  override fun hmacSha256(key: ByteString) = toByteString().hmacSha256(key)
+    val beginSegment = segment(beginIndex) // First segment to include
+    val endSegment = segment(endIndex - 1) // Last segment to include
 
-  override fun hmacSha512(key: ByteString) = toByteString().hmacSha512(key)
+    val newSegments = segments.copyOfRange(beginSegment, endSegment + 1)
+    val newDirectory = IntArray(newSegments.size * 2)
+    for ((index, s) in (beginSegment..endSegment).withIndex()) {
```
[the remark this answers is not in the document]
bnorm: Yes, it looks like it does. I'll optimize it.

bnorm:
So after some performance testing, I'm going to revert the change in implementation for `indexOf` and `lastIndexOf`. The existing implementation is currently faster than what I attempted to change it to so I will mess with improving that in another pull request.

swankjesse:
Nice!

PR #525, opened 2018-10-18
Long division in Throttler can cause infinite wait

bnorm:
Found this issue when messing around with the Throttler. Creating a failing unit test and pull request to discuss the best possible fix.

Preconditions:
- Completely idle Thorttler
- `waitByteCount` and `maxByteCount` are equal
- `bytesPerSecond` not an even divisor of `maxByteCount` and `1e9`

Basically we are trying to induce integer truncation in the `nanosForMaxByteCount` calculation. This will find its way to the `take()` method when calculating `immediateBytes` which will be inaccurate because of the truncation and result in having to wait. The problem comes when the `byteCountNanos` for waiting is calculated, it will have the same value as `usableNanos` because the Throttler is idle (`idleInNanos=0`) and result in a wait of `0`. Thus infinite waiting.

Possible solutions:
1. I had success converting `nanosForMaxByteCount` to a Double.
2. Disallow `maxByteCount` to be equal to `waitByteCount`. I've found a few vary large `bytesPerSecond` values (like `3e9+1`) that still don't work with this solution though.
3. ?

### DIFF
--- a/okio/jvm/src/test/java/okio/ThrottlerTest.kt
+++ b/okio/jvm/src/test/java/okio/ThrottlerTest.kt
@@ -21,6 +21,7 @@ import org.junit.After
 import org.junit.Before
 import org.junit.Test
 import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
 import kotlin.test.assertEquals
 
 class ThrottlerTest {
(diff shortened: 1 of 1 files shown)


swankjesse:
Wow, awesome find. Makes me want to get back to writing tests for this that use a fake clock and sleep function for real determinism!

Here’s one basic fix:

https://github.com/square/okio/pull/526/files

bnorm:
It also looks like rounding `nanosForMaxByteCount` up during calculation works as well:
```kotlin
this.nanosForMaxByteCount = when {
  bytesPerSecond != 0L -> (maxByteCount * 1_000_000_000L + bytesPerSecond - 1) / bytesPerSecond
  else -> -1L
}
```

PR #526, opened 2018-10-19
Don't wait 0 nanoseconds

swankjesse:
That causes an infinite wait, which we never want

### DIFF
--- a/okio/jvm/src/test/java/okio/ThrottlerTakeTest.kt
+++ b/okio/jvm/src/test/java/okio/ThrottlerTakeTest.kt
@@ -96,6 +96,16 @@ class ThrottlerTakeTest {
     assertElapsed(0L)
   }
 
+  /**
+   * We had a bug where integer division truncation would cause us to call wait() for 0 nanos. We
+   * fixed it by minimizing integer division generally, and by handling that case specifically.
+   */
+  @Test fun infiniteWait() {
+    throttler.bytesPerSecond(3, maxByteCount = 4, waitByteCount = 4)
+    takeFully(7)
+    assertElapsed(1000L)
+  }
+
   /** Take at least the minimum and up to `byteCount` bytes, sleeping once if necessary. */
   private fun take(byteCount: Long): Long {
     val byteCountOrWaitNanos = throttler.byteCountOrWaitNanos(nowNanos, byteCount)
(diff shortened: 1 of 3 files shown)


JakeWharton:
I mean, I guess it's in Kotlin here, but I could see it being an error prone check...

swankjesse:
I’m not completely satisfied with this.

At 3 bytes per second and a 4-byte maxByteCount, the initial immediateBytes is 3 and not 4 due to truncation when we compute that.

Fixing that probably fixes everything.

swankjesse:
Yeah, wait(0) is bad here, but not something you never need to do. The part where this logic is unhappy is that `0` only creeps in due to my sloppy use of truncating division.

swankjesse:
Okay, refactored to avoid unnecessary truncation. It’s not much different.

okio/jvm/src/test/java/okio/ThrottlerTakeTest.kt
```diff
@@ -96,6 +96,16 @@ class ThrottlerTakeTest {
...
 
+  /**
+   * We had a bug where integer division truncation would cause us to call wait() for 0 nanos. We
+   * fixed it by minimizing integer division generally, and by handling that case specifically.
+   */
+  @Test fun infiniteWait() {
+    throttler.bytesPerSecond(3, maxByteCount = 4, waitByteCount = 4)
+    takeFully(7)
+    assertElapsed(1000L)
+  }
```
swankjesse: @bnorm this is my simplification of your infinite wait discovery.

okio/jvm/src/test/java/okio/ThrottlerTest.kt
```diff
@@ -44,74 +42,6 @@ class ThrottlerTest {
     executorService.shutdown()
   }
 
-  @Test fun sourceByteCount() {
```
swankjesse: Removing this because it suffers off-by-ones because the new code truncates differently than the old code. (And now flaky because small movements in nanoTime() can change the result from say, 5 bytes to 6 bytes.)

okio/jvm/src/test/java/okio/ThrottlerTest.kt
```diff
@@ -122,18 +52,6 @@ class ThrottlerTest {
     stopwatch.assertElapsed(0.25)
   }
 
-  @Test fun sourceAfterClear() {
```
swankjesse: these `afterClear()` tests are less valuable now that we test synchronization/source/sink separately from the byte math.

swankjesse:
@bnorm PTAL ?

bnorm:
Looks good to me!

PR #532, opened 2018-11-04
Proof of concept for de-duplicating some coroutines code.

zach-klippenstein:
This pulls the common code from the `read` and `readAsync` methods out into an inline method that takes a function parameter for the actual read call. Inline functions can take suspending lambdas when called from suspend functions, so this is re-useable from both sync and async methods. This pattern could be copied for all the other methods.

### DIFF
--- a/okio/jvm/src/main/java/okio/RealBufferedSource.kt
+++ b/okio/jvm/src/main/java/okio/RealBufferedSource.kt
@@ -34,24 +34,25 @@ internal class RealBufferedSource(
   override fun buffer() = bufferField
 
   override fun read(sink: Buffer, byteCount: Long): Long {
-    require(byteCount >= 0) { "byteCount < 0: $byteCount" }
-    check(!closed) { "closed" }
-
-    if (buffer.size == 0L) {
-      val read = source.read(buffer, Segment.SIZE.toLong())
-      if (read == -1L) return -1L
-    }
-
-    val toRead = minOf(byteCount, buffer.size)
-    return buffer.read(sink, toRead)
+    return readMaybeAsync(sink, byteCount, Source::read)
   }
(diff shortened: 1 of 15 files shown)


swankjesse:
Oooh, yes, this is a powerful technique. I think I can run with this.

okio/jvm/src/main/java/okio/RealBufferedSource.kt
```diff
@@ -34,24 +34,25 @@ internal class RealBufferedSource(
...
-      if (read == -1L) return -1L
-    }
-
-    val toRead = minOf(byteCount, buffer.size)
-    return buffer.read(sink, toRead)
+    return readMaybeAsync(sink, byteCount, Source::read)
   }
 
   override suspend fun readAsync(sink: Buffer, byteCount: Long): Long {
+    // Can't use a suspend function reference here apparently.
```
Egorand: Hmm, is this a type inference problem, or a limitation of coroutines API?

zach-klippenstein: I think it's the former. The error message is something like:
```
Type mismatch: inferred type is KSuspendFunction0<Buffer, Long, Long> but (Buffer, Long) -> Long was expected
```

Egorand: Might be worth opening an issue on Kotlin's YouTrack.

zach-klippenstein: I just filed this issue: https://youtrack.jetbrains.net/issue/KT-28014

swankjesse:
Abandoning this for now. Loom is a nicer solution!

PR #536, opened 2018-11-12
Unsigned numbers in Okio

### LINKED ISSUE #513: BufferedSource.readUByte etc.
https://blog.jetbrains.com/kotlin/2018/08/kotlin-1-3-m2/

bnorm:
Fixes #513

### DIFF
--- a/samples/src/jvmMain/kotlin/okio/samples/-Unsigned.kt
+++ b/samples/src/jvmMain/kotlin/okio/samples/-Unsigned.kt
@@ -1,0 +1,373 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+@file:JvmName("-Unsigned")
+@file:Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
+
+package okio.samples
+
+import okio.BufferedSink
+import okio.BufferedSource
+import java.io.IOException
+
+/** Writes a unsigned byte to this sink. */
+@ExperimentalUnsignedTypes
+@Throws(IOException::class)
+inline fun <S : BufferedSink> S.writeUByte(b: UByte): S {
+  writeByte(b.toByte().toInt())
+  return this
+}
+
+/**
+ * Writes a big-endian, unsigned short to this sink using two bytes.
+ * ```
+ * val buffer = Buffer()
+ * buffer.writeUShort(65534u.toUShort())
+ * buffer.writeUShort(15u.toUShort())
+ *
+ * assertEquals(4, buffer.size)
+ * assertEquals(0xff.toByte(), buffer.readByte())
+ * assertEquals(0xfe.toByte(), buffer.readByte())
+ * assertEquals(0x00.toByte(), buffer.readByte())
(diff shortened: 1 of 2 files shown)


okio/jvm/src/main/java/okio/-Unsigned.kt
```diff
@@ -0,0 +1,380 @@
...
+/** Writes a unsigned byte to this sink. */
+@OkioUnsignedApi
+@Throws(IOException::class)
+inline fun <S : BufferedSink> S.writeUByte(b: UByte): S {
+  writeByte(b.toByte().toInt())
+  return this
+}
+
+/**
+ * Writes a big-endian unsigned short to this sink using two bytes.
```
bnorm: Grammar: preference on "big-endian, unsigned short" or "unsigned, big-endian short"?

swankjesse: I think “unsigned short” is the subject, and big-endian is a representation of that subject . . . 

so I like big-endian, unsigned short

okio/jvm/src/main/java/okio/-Unsigned.kt
```diff
@@ -0,0 +1,380 @@
...
+@file:Suppress("NOTHING_TO_INLINE") // Syntactic sugar.
+@file:UseExperimental(ExperimentalUnsignedTypes::class)
+
+package okio
+
+import java.io.IOException
+
+/** Marker interface for APIs related to unsigned numbers. */
+@Experimental
+annotation class OkioUnsignedApi
```
bnorm: Do we want to have our own experimental annotation or pass through the `ExperimentalUnsignedTypes` annotation?

On one hand it would be nice to have our own annotation so we don't have to become non-experimental when unsigned types do, but on the other I don't see this API changing much. I'm leaning towards removing `OkioUnsignedApi`, let me know if you want me to keep it.

swankjesse:
The annotation is the most interesting part of this. I think my preference is to publish this as a gist or sample until unsigned types are stable.

bnorm:
I would be fine with either a sample or gist. I think I would lean to a sample as it will force us to update it if unsigned types ever change and might make it easier to find.

swankjesse:
Good call. Make is so!

build.gradle
```diff
@@ -10,7 +10,7 @@ buildscript {
       'assertj': '1.7.0',
       'shadowPlugin': '4.0.4',
       'nodePlugin': '1.2.0',
-      'ktlintPlugin': '5.1.0',
+      'ktlintPlugin': '7.1.0',
```
bnorm: Old version didn't like unsigned literals in the samples project. Not sure why it was okay before.

okio/jvm/src/main/java/okio/-Unsigned.kt
JakeWharton: I would pass it through the original because we're using their types in the public API. If we were using them as implementation detail only I would prefer our own annotation.

swankjesse:
Still waiting on signed types?

swankjesse:
Still waiting on unsigned types?

ImUrX:
:sob:

ImUrX:
whats missing from this? it would really nice to have it!

MohammedKHC:
@swankjesse @JakeWharton 

I see that this is already approved. Can it get merged?

Having unsigned functions will remove the need for all this masking
https://github.com/square/okio/blob/9fbab0f7ab00b525b0ae331c8c3ac3c645afc8c8/okio/src/zlibMain/kotlin/okio/internal/ZipFiles.kt#L198-L216

For example instead of `readShortLe().toInt() and 0xffff `  it could just be `readUShortLe()` right?

And it will generally make me happy when parsing files that store unsigned Integers.

ShozanAhmed:
@oldergod @JakeWharton @bnorm @swankjesse 
Plz merge it's been like 7 years 🥲

JakeWharton:
Note that this PR:
- Is very old and thus uses the now outdated experimental annotation.
- Is only for the JVM rather than common.
- Has its sources in a sample not the main artifact.
- Lacks the array-based unsigned types (probably because they didn't exist yet at the time).

So no, we cannot merge the PR as-is. I'm going to close it. However, I think we're ready to merge functions like these into the correct place now.

PR #540, opened 2018-11-27
Gradle 5.0

Egorand:


Egorand:
Looks like the old Kotlin Gradle plugins are incompatible with Gradle 5, so we have to move to the new plugin first..

Egorand:
Superseded by #544

PR #546, opened 2019-01-10
Adds support for Folding Pipes

oldergod:


okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -66,25 +74,81 @@ class Pipe(internal val maxBufferSize: Long) {
...
         sinkClosed = true
         (buffer as Object).notifyAll() // Notify the source that no more bytes are coming.
       }
+
+      delegate?.let { honorsTimeoutAndDeadline(it) { close() } }
     }
 
     override fun timeout(): Timeout = timeout
+
+    private fun honorsTimeoutAndDeadline(delegateSink: Sink, block: Sink.() -> Unit) {
```
swankjesse: There’s a very constrained way you always invoke this:

```
delegate?.let { honorsTimeoutAndDeadline(it) { ... } }
```

Here’s a proposal for another signature

```
fun Sink.forward(block: Sink.() -> Unit)
```

And then the callsite could be:

```
delegate?.forward { ... }
```

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -66,25 +74,81 @@ class Pipe(internal val maxBufferSize: Long) {
...
+      val sinkTimeoutNanos = this.timeout().timeoutNanos()
+      val delegateOriginalTimeoutNanos = delegateSink.timeout().timeoutNanos()
+      val minTimeoutNanos = when {
```
swankjesse: I think you want a method, `minTimeout(aNanos: Long, bNanos: Long)` that does this when block

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -66,25 +74,81 @@ class Pipe(internal val maxBufferSize: Long) {
...
+        sinkTimeoutNanos == 0L -> delegateOriginalTimeoutNanos
+        delegateOriginalTimeoutNanos == 0L -> sinkTimeoutNanos
+        sinkTimeoutNanos < delegateOriginalTimeoutNanos -> sinkTimeoutNanos
+        else -> delegateOriginalTimeoutNanos
+      }
+
+      val sinkDeadline =
+        if (this.timeout().hasDeadline()) this.timeout().deadlineNanoTime() else null
```
swankjesse: can you rewrite this without nulls ? Nulls mean boxes and boxes mean allocations

swankjesse: actually even better – just define a method on `Timeout` that takes a lambda and does this internally. When you’re inside the timeout class you don’t have to do nearly as much awkward stuff to make the timeout/deadline stuff work

okio/jvm/src/main/java/okio/Pipe.kt
swankjesse: should be inline so it doesn’t allocate

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -116,6 +180,36 @@ class Pipe(internal val maxBufferSize: Long) {
     override fun timeout(): Timeout = timeout
   }
 
+  /**
+   * Writes any buffered contents of this pipe to `sink`, then changes this pipe's sink to forward
+   * to `sink`. This pipe's source is closed and attempts to read it will throw
+   * [IllegalStateException].
+   *
+   * This method must not be called while concurrently accessing this pipe's sink. It is safe,
```
swankjesse: This method must not be called while concurrently accessing this pipe's **sink**.
 -> This method must not be called while concurrently accessing this pipe's **source**.

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -116,6 +180,36 @@ class Pipe(internal val maxBufferSize: Long) {
...
+   * however, to call this while concurrently writing this pipe's source.
```
swankjesse: and again, this is backwards.

swankjesse: Maybe the thing to write is that this method replaces the pipe’s source with `sink`

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -15,10 +15,27 @@
...
 import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import kotlin.test.assertFailsWith
 
 class PipeKotlinTest {
+  @JvmField @Rule val timeout = org.junit.rules.Timeout(5, TimeUnit.SECONDS)
+
+  private val executorService = Executors.newScheduledThreadPool(2)
```
swankjesse: should be 1 ?

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+    val pipe = Pipe(128)
+
+    val pipeSink = pipe.sink.buffer()
+    pipeSink.writeUtf8("hello")
+    pipeSink.emit()
+
+    val pipeSource = pipe.source.buffer()
+    assertEquals("hello", pipeSource.readUtf8(5))
+
+    val actualSinkBuffer = Buffer()
```
swankjesse: actualSink --> foldedSink?

(that’s the name we use elsewhere)

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+    val sink = pipe.sink.buffer()
+    sink.writeUtf8("abcdefgh") // Blocks writing 8 bytes to a 4 byte pipe.
+    sink.close()
+
+    latch.await()
+    assertEquals("abcdefgh", foldSink.readUtf8())
+  }
+
+  @Test
+  fun accessSinkAfterFold() {
```
swankjesse: accessSourceAfterFold

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+  class TimeoutFlushingSink : Sink {
+    val timeout = object : AsyncTimeout() {
+      override fun timedOut() {
+        synchronized(this@TimeoutFlushingSink) {
+          (this@TimeoutFlushingSink as Object).notifyAll()
+        }
+      }
+    }
+
+    override fun write(source: Buffer, byteCount: Long) = Unit
```
swankjesse: nit: to be correct to the Sink contract this method needs to call `source.skip(byteCount)`

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+  class TimeoutWritingSink : Sink {
+    val timeout = object : AsyncTimeout() {
+      override fun timedOut() {
+        synchronized(this@TimeoutWritingSink) {
+          (this@TimeoutWritingSink as Object).notifyAll()
+        }
+      }
+    }
+
+    override fun write(source: Buffer, byteCount: Long) {
```
swankjesse: nit: to be correct to the Sink contract this method needs to call source.skip(byteCount)

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+  class TimeoutClosingSink : Sink {
+    val timeout = object : AsyncTimeout() {
+      override fun timedOut() {
+        synchronized(this@TimeoutClosingSink) {
+          (this@TimeoutClosingSink as Object).notifyAll()
+        }
+      }
+    }
+
+    override fun write(source: Buffer, byteCount: Long) = Unit
```
swankjesse: nit: to be correct to the Sink contract this method needs to call source.skip(byteCount)

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+    val elapsed = TimeUnit.MILLISECONDS.toNanos(System.currentTimeMillis() - start)
+
+    assertEquals(smallerTimeoutNanos.toDouble(), elapsed.toDouble(),
+      TimeUnit.MILLISECONDS.toNanos(100).toDouble())
+    assertEquals(smallerTimeoutNanos, underlying.timeout().timeoutNanos())
+  }
+
+  @Test
+  fun honorsPipeSinkTimeoutOnFlushingWhenItIsSmaller() {
+    val smallerTimeoutNanos = TimeUnit.MILLISECONDS.toNanos(500L)
```
swankjesse: I wonder if it’s worth having class-level fields so we don’t need to keep saying these two

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -32,4 +49,682 @@ class PipeKotlinTest {
...
+
+    val pipe = Pipe(4)
+    val underlying = TimeoutFlushingSink()
+
+    underlying.timeout.timeout(biggerTimeoutNanos, TimeUnit.NANOSECONDS)
+    pipe.sink.timeout().timeout(smallerTimeoutNanos, TimeUnit.NANOSECONDS)
+
+    pipe.fold(underlying)
+
+    val start = System.currentTimeMillis()
```
swankjesse: maybe a method like this:

```
fun assertDuration(expected: Double, block: () -> Unit)
```

okio/jvm/src/main/java/okio/Pipe.kt
oldergod: had to move it out of the Sink class then otherwise would not be able to have access to `this@Pipe.sink`

okio/jvm/src/main/java/okio/Timeout.kt
```diff
@@ -208,5 +236,12 @@ open class Timeout {
 
       override fun throwIfReached() {}
     }
+
+    fun minTimeout(aNanos: Long, bNanos: Long) = when {
+      aNanos == 0L -> bNanos
+      bNanos == 0L -> aNanos
+      aNanos < bNanos -> aNanos
+      else -> bNanos
+    }
```
oldergod: maybe we move it into some Util top level file

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -116,6 +145,40 @@ class Pipe(internal val maxBufferSize: Long) {
...
+  /**
+   * Writes any buffered contents of this pipe to `sink`, then replace this pipe's source with
+   * `sink`. This pipe's source is closed and attempts to read it will throw an
+   * [IllegalStateException].
+   *
+   * This method must not be called while concurrently accessing this pipe's source. It is safe,
+   * however, to call this while concurrently writing this pipe's sink.
+   */
+  @Throws(IOException::class)
+  fun fold(sink: Sink) {
```
swankjesse: Just thought of a crazy and ugly edge case . . . what if the `sink.write(...)` call fails ? I think we put things in a very ugly state.

The fix should be really easy. We need a finally clause on the `sink.write()` call that sets `sourceClosed = true` if that method throws

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -116,6 +145,40 @@ class Pipe(internal val maxBufferSize: Long) {
...
+        sinkBuffer.write(buffer, buffer.size)
+        (buffer as Object).notifyAll() // Notify the sink that it can resume writing.
+        return@synchronized sinkBuffer
+      }
+
+      sink.write(sinkBuffer, sinkBuffer.size)
+    }
+  }
+
+  private inline fun Sink.forward(block: Sink.() -> Unit) {
```
swankjesse: neat

okio/jvm/src/main/java/okio/Pipe.kt
oldergod:
```
try {
  sinkBuffer.write(buffer, buffer.size)
} catch (e: IOException) {
  sourceClosed = true
  throw e
}
```
?

oldergod: btw, I `try/finally` inside `mergeOnce` so that we always set the delegate timeout to what it was, even when write/flush/close throws.

oldergod: double checking but you close the source even when it doesn't throw, so it should be inside a catch right?
```kotlin
try {
  sink.write(sinkBuffer, sinkBuffer.size)
} catch (e: IOException) {
  synchronized(buffer) {
    sourceClosed = true
    (buffer as Object).notifyAll() // Notify the sink that it can resume writing.
  }
  throw e
}

```

okio/jvm/src/main/java/okio/Timeout.kt
```diff
@@ -196,6 +196,36 @@ open class Timeout {
     }
   }
 
+  /**
+   * Applies the minimum intersection between this timeout and `other`, run `block`, then finally
+   * rollback this timeout's values.
+   */
+  inline fun mergeOnce(other: Timeout, block: () -> Unit) {
```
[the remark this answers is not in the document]
oldergod: done without fallback values nor nulls

okio/jvm/src/main/java/okio/Pipe.kt
swankjesse: try/finally is a little neater cause it'll handle the case where it's an unchecked exception

okio/jvm/src/main/java/okio/Pipe.kt
```diff
@@ -116,6 +145,48 @@ class Pipe(internal val maxBufferSize: Long) {
...
+  /**
+   * Writes any buffered contents of this pipe to `sink`, then replace this pipe's source with
+   * `sink`. This pipe's source is closed and attempts to read it will throw an
+   * [IllegalStateException].
+   *
+   * This method must not be called while concurrently accessing this pipe's source. It is safe,
+   * however, to call this while concurrently writing this pipe's sink.
+   */
+  @Throws(IOException::class)
+  fun fold(sink: Sink) {
```
bnorm: What is the expected behavior of folding a pipe a second time?

swankjesse: good catch! it should fail with an IllegalStateException

okio/jvm/src/test/java/okio/PipeKotlinTest.kt
```diff
@@ -15,11 +15,28 @@
...
+import org.junit.Rule
 import org.junit.Test
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import kotlin.test.assertFailsWith
 
 class PipeKotlinTest {
-  @Test fun test() {
+  @JvmField @Rule val timeout = org.junit.rules.Timeout(5, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
swankjesse: simple name collision!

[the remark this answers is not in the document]
JakeWharton:
```kotlin
import org.junit.rules.Timeout as JUnitTimeout
```

PR #548, opened 2019-01-15
Adds support for Folding Pipes in Java

oldergod:


okio/src/main/java/okio/Pipe.java
```diff
@@ -56,43 +60,128 @@ public final Sink sink() {
...
           }
+        }
+      }
 
-          long bytesToWrite = Math.min(bufferSpaceAvailable, byteCount);
-          buffer.write(source, bytesToWrite);
-          byteCount -= bytesToWrite;
-          buffer.notifyAll(); // Notify the source that it can resume reading.
+      if (delegate != null) {
+        timeoutFolder.push(this.timeout());
```
oldergod: Regard to `timeoutFolder`, should I do something similar to what we do for getting `folderSink` out of the synchronized block?

okio/src/main/java/okio/Pipe.java
```diff
@@ -56,43 +60,128 @@ public final Sink sink() {
...
-        while (byteCount > 0) {
-          if (sourceClosed) throw new IOException("source is closed");
-
-          long bufferSpaceAvailable = maxBufferSize - buffer.size();
-          if (bufferSpaceAvailable == 0) {
-            timeout.waitUntilNotified(buffer); // Wait until the source drains the buffer.
-            continue;
+        while (byteCount > 0 && delegate == null) {
+          if (foldedSink != null) {
+            delegate = foldedSink;
```
swankjesse: `break;` so you don’t need an else?

swankjesse: ... and then you also don’t need the delegate check on the loop’s condition

okio/src/main/java/okio/Pipe.java
swankjesse: yes please. It’s the easiest to reason about

okio/src/test/java/okio/PipeTest.java
```diff
@@ -357,6 +372,195 @@
...
+
+    final BufferedSink pipeSink = Okio.buffer(pipe.sink());
+    pipeSink.writeUtf8("hello");
+    pipeSink.emit();
+
+    final BufferedSource pipeSource = Okio.buffer(pipe.source());
+    assertEquals("hello", pipeSource.readUtf8(5));
+
+    final Buffer foldedSinkBuffer = new Buffer();
+    final boolean[] foldedSinkClosed = {false};
```
swankjesse: For what it’s worth, AtomicBoolean is much nicer for this job.

swankjesse:
Merged to land this! Travis is unhappy but I verified manually.

```

[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO]
[INFO] Okio (Parent) ...................................... SUCCESS [  0.920 s]
[INFO] Okio ............................................... SUCCESS [02:20 min]
[INFO] Okio Performance Benchmarks ........................ SUCCESS [  2.539 s]
[INFO] Okio Samples ....................................... SUCCESS [  0.410 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
```

okio/src/test/java/okio/PipeTest.java
oldergod: Thank you! The IDE suggested I turned to bool into an array and I didn't think twice

PR #572, opened 2019-04-04
Extract Sink, Source, BufferedSink and BufferedSource into common as marker interfaces

Egorand:


Egorand:
Sadly there's no `kotlin.Closeable`

PR #573, opened 2019-04-04
Copy KDoc for common interfaces from JVM

Egorand:
These are not rendered as of now but it should get fixed in the next Dokka release. Unclear what the KDoc merging strategy is going to be, hopefully they'll be merging common KDoc with platform-specific: `Source` and `Sink` have paragraphs on `InputStream` and `OutputStream` comparisons which should ideally live in the JVM module.


Egorand:
Sure, fair enough

PR #574, opened 2019-04-04
Extract expect classes for Buffer and Timeout

Egorand:


okio/js/src/main/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright (C) 2019 Square, Inc.
```
JakeWharton: I was going to say that this is derivative from the original types so it should retain the original years, but it seems like we 2019'd all the Kotlin classes so whevs.

Egorand:
I'm hoping to be able pull out some common code as well

PR #575, opened 2019-04-04
Extract all expect interface methods into common

Egorand:


rjrjr:
Did you mean to be merging this into `egorand/190404/expect-buffer-timeout` instead of master?

Egorand:
Yes!

okio/src/main/kotlin/okio/Source.kt
```diff
@@ -15,4 +15,8 @@
  */
 package okio
 
-expect interface Source
+expect interface Source {
+  fun read(sink: Buffer, byteCount: Long): Long
+
```
JakeWharton: Missing close

okio/src/main/kotlin/okio/Sink.kt
```diff
@@ -15,4 +15,8 @@
  */
 package okio
 
-expect interface Sink
+expect interface Sink {
+  fun write(source: Buffer, byteCount: Long)
+
```
JakeWharton: Missing close and flush

okio/src/main/kotlin/okio/BufferedSource.kt
```diff
@@ -15,4 +15,82 @@
...
+
+  fun readDecimalLong(): Long
+
+  fun readHexadecimalUnsignedLong(): Long
+
+  fun skip(byteCount: Long)
+
+  fun readByteString(): ByteString
+
+  fun readByteString(byteCount: Long): ByteString
```
JakeWharton: Lots of cases where default argument values with `@JvmOverloads` should be used here in common. Also in BufferedSink.

okio/src/main/kotlin/okio/Buffer.kt
```diff
@@ -15,4 +15,131 @@
...
+
+  override fun readIntLe(): Int
+
+  override fun readLongLe(): Long
+
+  override fun readDecimalLong(): Long
+
+  override fun readHexadecimalUnsignedLong(): Long
+
+  override fun readByteString(): ByteString
```
JakeWharton: None of these functions should need redeclare unless they're using covariant return types to change the return type to `Buffer`

Egorand: Oh, good call. I was getting build failures so pulled everything out without digging deep into specific methods. 

Covariant returns are a bit inconsistent due to the fact that `Buffer` omits return types in many places, so it's hard to say whether using covariant returns was intentional or not.

okio/src/main/kotlin/okio/BufferedSource.kt
Egorand: Can address in more targeted PRs

okio/src/main/kotlin/okio/Sink.kt
Egorand: Those come from Java interfaces so keeping them in JVM for now. Can also introduce our own interfaces and typealias to Java ones, would that make sense?

JakeWharton: On the JVM we use the interfaces to carry forward the declarations, but those methods are still intrinsic to the type and need defined here. What is a sink or a source that you can't call close on?

okio/src/main/kotlin/okio/BufferedSource.kt
JakeWharton: File an issue so we don't lose track? I wouldn't want to ship a version like this

okio/src/main/kotlin/okio/Buffer.kt
JakeWharton: I think it should just be for all the `write` methods to support chaining on the JVM without running into the checked exceptions from `BufferedSink`

okio/src/main/kotlin/okio/Buffer.kt
```diff
@@ -15,4 +15,135 @@
  */
 package okio
 
-expect class Buffer : BufferedSource, BufferedSink
+expect class Buffer : BufferedSource, BufferedSink {
```
rjrjr: Pretty sure you want `Buffer()` here, the zero args constructor is public.

okio/js/src/main/kotlin/okio/Buffer.kt
```diff
@@ -15,4 +15,149 @@
  */
 package okio
 
-actual class Buffer : BufferedSource, BufferedSink
+actual class Buffer : BufferedSource, BufferedSink {
```
swankjesse: FWIW I think Buffer fits better in common completely.

okio/js/src/main/kotlin/okio/Sink.kt
```diff
@@ -15,4 +15,12 @@
  */
 package okio
 
-actual interface Sink
+actual interface Sink {
```
swankjesse: can we just put these interfaces in common?

okio/src/main/kotlin/okio/Source.kt
swankjesse: Fixed.

okio/js/src/main/kotlin/okio/Sink.kt
rjrjr: I don't think so, b/c of the java.io deps.

okio/js/src/main/kotlin/okio/Buffer.kt
rjrjr: Again, lots of java deps. Unless it's possible to have a non-`expects` common class enhanced by a platform specific one? Didn't think that was the case.

Egorand: That would be the use case for concrete implementations being allowed in `expect` classes, which is not there at the moment. For now we'll have to work around it in the same way we did with `ByteString`.

okio/js/src/main/kotlin/okio/Sink.kt
Egorand: Right, we're adding supertypes for `Sink` on the JVM side and need an `actual` for this.

PR #577, opened 2019-04-04
Upgrades to 1.3.30-eap-125, and drops workarounds for KT22818.

rjrjr:


okio/js/src/main/kotlin/okio/ByteString.kt
```diff
@@ -85,14 +86,7 @@ internal actual constructor(
    */
   actual open fun toAsciiUppercase(): ByteString = commonToAsciiUppercase()
 
-  // TODO move substring() when https://youtrack.jetbrains.com/issue/KT-22818 is fixed
```
swankjesse: # 💥

rjrjr:
Update rebases to keep up w/master

okio/jvm/src/main/java/okio/ByteString.kt
```diff
@@ -141,15 +142,8 @@ internal actual constructor(
    */
   actual open fun toAsciiUppercase(): ByteString = commonToAsciiUppercase()
 
-  // TODO move substring() when https://youtrack.jetbrains.com/issue/KT-22818 is fixed
-
-  /**
```
Egorand: We should keep the KDoc in JVM module for now cause Dokka is broken. They plan to fix this in the next release, then we can move all KDoc into common.

rjrjr: Sorry, missed this comment before I merged.

PR #578, opened 2019-04-05
Multiplat read/write ByteString, ByteArray.

rjrjr:
:boom:


okio/src/main/kotlin/okio/SegmentPool.kt
```diff
@@ -28,14 +25,11 @@ internal object SegmentPool {
   const val MAX_SIZE = 64 * 1024L // 64 KiB.
 
   /** Singly-linked list of segments.  */
-  @JvmField
```
Egorand: Were these annotations unused?

rjrjr: It's an internal class, and the code's all Kotlin now, so I don't so how they could have been.

rjrjr:
Update rebases. 

- [ ] Still trying to solve the `NoSuchMethodError` from `Buffer.size`, https://youtrack.jetbrains.com/issue/KT-23699
- [ ] Still need to move related tests to common

rjrjr:
PTAL, a lot has changed.

Multi-plat support, tests for reading and writing ByteArray, ByteString.

okio/js/src/main/kotlin/okio/Buffer.kt
```diff
@@ -15,147 +15,187 @@
...
-    throw UnsupportedOperationException()
+  actual override fun require(byteCount: Long) {
+    if (size < byteCount) throw EOFException()
   }
 
-  override fun request(byteCount: Long): Boolean = throw UnsupportedOperationException()
+  actual override fun request(byteCount: Long): Boolean = size >= byteCount
 
-  override fun peek(): BufferedSource = throw UnsupportedOperationException()
+  actual override fun peek(): BufferedSource {
```
rjrjr: These first few methods aren't covered by the common test, should look for more to move over.

okio/src/main/kotlin/okio/Buffer.kt
```diff
@@ -15,13 +15,93 @@
...
+  var size: Long
+    internal set
+
   override val buffer: Buffer
 
   override fun emitCompleteSegments(): Buffer
 
   override fun emit(): Buffer
 
+  override fun exhausted(): Boolean
```
rjrjr: Oops, I stomped work jake did to kill unnecessary ones.

okio/js/src/main/kotlin/okio/Buffer.kt
rjrjr: Have to happen in follow up, and I don't think we'll miss it.

rjrjr:
K, I think this is good to go, and I'm out of hack week so I'll stop now.

okio/jvm/src/main/java/okio/Buffer.kt
```diff
@@ -259,35 +245,11 @@ actual class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
 
   /** Returns the byte at `pos`.  */
   @JvmName("getByte")
-  operator fun get(pos: Long): Byte {
-    checkOffsetAndCount(size, pos, 1L)
-    seek(pos) { s, offset ->
-      return s!!.data[(s.pos + pos - offset).toInt()]
-    }
-  }
+  operator fun get(pos: Long): Byte = commonGet(pos)
```
swankjesse: mmmmmm . . . . . . this is going to have performance costs in a place where I’ve been really anxious about performance costs. This one method gets called a lot parsing JSON, and we’re inner-loop enough that just another stack frame is substantial

I’d like to fix more of the problem by pushing more into common, so that `get` is just commonGet(). How far are we from that?

okio/jvm/src/main/java/okio/Okio.kt
```diff
@@ -34,18 +34,9 @@ import java.nio.file.Path
 import java.util.logging.Level
 import java.util.logging.Logger
 
-/**
- * Returns a new source that buffers reads from `source`. The returned source will perform bulk
- * reads into its in-memory buffer. Use this wherever you read a source to get an ergonomic and
- * efficient access to data.
- */
-fun Source.buffer(): BufferedSource = RealBufferedSource(this)
+actual fun Source.buffer(): BufferedSource = RealBufferedSource(this)
```
swankjesse: cool

okio/src/main/kotlin/okio/internal/Buffer.kt
```diff
@@ -0,0 +1,209 @@
+package okio.internal
```
swankjesse: nit: copyright

okio/src/main/kotlin/okio/internal/Buffer.kt
```diff
@@ -0,0 +1,209 @@
...
+
+import okio.Buffer
+import okio.ByteString
+import okio.EOFException
+import okio.Segment
+import okio.SegmentPool
+import okio.checkOffsetAndCount
+
+// TODO Kotlin's expect classes can't have default implementations, so platform implementations
+// have to call these functions. Remove all this nonsense when expect class allow actual code.
```
swankjesse: yeah, this is very serious nonsense

JakeWharton:
You can't make it 100% common because it implements JVM specific types and has JVM specific methods.

swankjesse:
Can we define those on JS and native? Slightly gross to redefine `java.io.Closeable` on Native, but possible?

swankjesse:
Though the more I think about it, we just want this:
https://youtrack.jetbrains.com/issue/KT-20427

okio/jvm/src/main/java/okio/Buffer.kt
swankjesse: ... I think we can workaround that by making `commonXxxx` methods inline, and avoid calling them anywhere but from `actual` functions.

Egorand:
Can we just define our own `Closeable` and `Flushable` with JVM typealiases, would that allow us to make `Buffer` 100% common? Can take a shot at it during my flight.

JakeWharton:
It wouldn't, no. It implements a NIO type and has functions for reading and
writing IO types.

Egorand:
@JakeWharton right. Re-designing those as extension functions would've been possible, but it's incompatible with Java.

rjrjr:
Follow up tracked at #589, but I can't promise to do it — I'm back to my day job.

rjrjr:
> @JakeWharton right. Re-designing those as extension functions would've been possible, but it's incompatible with Java.

Yes, this! I kept thinking how much cleaner `Buffer` could be if we didn't have to maintain backward compatibility. Is it worth thinking about a `Buffer2`?

JakeWharton:
The implications of that are dramatic and the benefit seems only to be solving a maintenance burden. I would rather us deal with pain than inflict a dramatic migration story on our users. Plus once Kotlin supports partial expects it should reduce the maintenance burden to near zero.

rjrjr:
Update adds missing copyright blocks.

Egorand:
:shipit:

PR #581, opened 2019-04-05
Replace default Gradle source directories with custom layout

### LINKED ISSUE #580: Project import into IntelliJ
Anybody else having trouble working with the project in IntelliJ? I'm seeing a lot of warnings and errors being reported but obviously the project compiles with Gradle just fine.

I've been poking at this for the last couple days and one of the bigger problems I've found is the conflict between where the project Kotlin common code is located and where IntelliJ presumes the Java code for the `jvmWithJava` code is going to be. But if the project Kolin common code is moved to `src/commonMain/kotlin` and `src/commonTest/kotlin`, a lot of the expect and actual warnings I've seen go away.

I know there's currently a lot of churn with the common code going on right now but I wanted to record my findings and see if anybody else is having the same problems I'm having.

bnorm:
Instead of _adding_ desired source directories, _replace_ the existing,
default source directories with the custom layout used by Okio. This
makes IntelliJ *much* happier about the project layout.

closes #580 - I was finally able to figure this out without having to move any files. `samples` is still causing problems but I'll keep working on it.

### DIFF
--- a/okio/build.gradle
+++ b/okio/build.gradle
@@ -5,7 +5,11 @@ kotlin {
     // egor: jvmWithJava is getting deprecated but it's not clear what the replacement is. For now 
     // gonna keep it inside targets {}
     fromPreset(presets.jvmWithJava, 'jvm') {
-      project.sourceSets.test.java.srcDir('jvm/src/test/java')
+      project.sourceSets.main.java.setSrcDirs(['jvm/src/main/java'])
+      project.sourceSets.test.java.setSrcDirs(['jvm/src/test/java'])
+      project.sourceSets.main.resources.setSrcDirs(['jvm/src/main/resources'])
+      project.sourceSets.main.kotlin.setSrcDirs(['jvm/src/main/kotlin'])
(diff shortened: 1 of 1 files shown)


okio/build.gradle
```diff
@@ -27,56 +31,61 @@ kotlin {
...
-      kotlin.srcDir('js/src/test/kotlin')
+      kotlin.setSrcDirs(['js/src/test/kotlin'])
       dependencies {
         implementation deps.kotlin.test.js
       }
     }
     nativeMain {
-      kotlin.srcDir('native/src/main/kotlin')
+      kotlin.setSrcDirs(['native/src/main/kotlin'])
+      dependsOn commonMain
```
Egorand: AFAIK these are redundant, platform specific builds depend on common by default.

bnorm: That's what I was thought as well, but I think I needed to add these so the navigation from expect to actual listed native code as one of the actual implementations. I can double check and remove if I ever figure out the `sample` project.

Egorand: Interesting! If that proves to be the case I think filing an issue will make sense.

swankjesse:
If you’d like, move files!

PR #583, opened 2019-04-05
Update japicmp plugin

JakeWharton:
This bumps japicmp to 0.13.1 which still doesn't contain the effectively-final fix, but at least provides compatibility with newer japicmp versions so that we can upgrade once that change is released.


swankjesse:
Thanks!

PR #586, opened 2019-04-07
 Move Options.kt into common

Egorand:


Egorand:
Japicmp fails with the following:
```
Comparing binary compatibility of  against 
***! MODIFIED CLASS: PUBLIC FINAL okio.Options  (not serializable)
	***! CLASS FILE FORMAT VERSION: 50.0 <- 51.0
```
Any idea what might be causing this?

bnorm:
The Japicmp failure is a bytecode format change? 50 is Java 6 and 51 is Java 7.
https://en.wikipedia.org/wiki/Java_class_file#General_layout

Egorand:
Yeah, looks like it, though I'm pretty sure I haven't seen this kind of failures before. It's also weird that our JVM module produces Java 6 bytecode, whereas MPP produces Java 7. Could it be due to the fact that I switched from Java's `java.util.RandomAccess` and `java.util.AbstractList` to Kotlin types, which are possibly only declared in `kotlin-stdlib-jdk7`?

Egorand:
Added to Japicmp excludes. Wonder why we haven't hit this problem with other classes that moved from JVM to common.

okio/jvm/japicmp/build.gradle
```diff
@@ -45,6 +45,7 @@ task japicmp(type: JapicmpTask, dependsOn: 'jar') {
       'okio.SegmentedByteString', // Internal.
       'okio.SegmentPool', // Internal.
       'okio.Util', // Internal.
+      'okio.Options', // Bytecode version changed from 51.0 to 50.0
```
JakeWharton: This is pretty drastic though as it means the API comparison will become useless as types are migrated since they'll all have their major version changed.

Egorand: Agree, it's not ideal. Any idea why this hasn't surfaced for other classes moved into common?

swankjesse:
Opened a tracking issue for the class file format warning.
https://github.com/square/okio/issues/597

Merging this ’cause it’s forward progress.

PR #588, opened 2019-04-07
Gradle 5.3.1

Egorand:


Egorand:
@JakeWharton is there a way to check? I usually just run `:check` or `:assemble` to allow the wrapper to pull updates.

JakeWharton:
I always run the wrapper task twice. First one downloads the updated jar,
second one updates the scripts.

cketti:
Run `gradle wrapper --gradle-version 5.3.1 --distribution-type all` to update all wrapper files.

Egorand:
@cketti yep, that's what I did, with the only difference that I ran it with `./gradlew`. Does this matter?

Egorand:
Running the `wrapper` task twice has indeed pulled the updated scripts. Now for some reason the Java code in `samples` can't see code defined in `common`. Investigating.

Egorand:
Possibly an issue with MPP, filed https://youtrack.jetbrains.com/issue/KT-30881

bnorm:
I think we might need to use the Kotlin multiplatform plugin in samples as well: https://github.com/h0tk3y/k-new-mpp-samples/blob/master/project-dependency/app/build.gradle

Egorand:
Hmm, interesting. Will look into it.

cketti:
@Egorand: Apparently I've been doing it wrong all this time as well :smile:  The version of the invoked Gradle binary is used to update the wrapper scripts when running the wrapper task. So, assuming you're not already using the Gradle version to which you want to upgrade the wrapper, you'll have to run the wrapper task twice to ensure the wrapper scripts are updated to the target version.

Egorand:
Yeah, I don't see why `wrapper` can't just update everything.

Issue #594, opened 2019-04-27
why peeked BufferSource.request()  blocked all the time


LightSun:
Hi , All ! You are write a very good framework of IO.
I write a program which used okio to read and write message.
but when i peek a buffer source (origin from socket).  the request method will blocked all the time.
(the client write data and flush ).
if use timeout. it will be closed by timeout.
```java
 BufferedSource peek = in.peek();
                if(!peek.request(8)){   // blocked here all the time.
                    System.out.println("message not all reached.");
                    continue;
      }
```
and if i use this . it will be ok. so . it think this is a bug.
```java
               in.skip(4);
                int len = in.readInt();
                if(!in.request(len)){
                    System.out.println("data not reached.");
                    break;
                }
                byte[] bytes = in.readByteArray(len);
                System.out.println("message reached:  " + new String(bytes, StandardCharsets.UTF_8));
```

the full sample code is here.
```java
public class Server3 {
    // initialize socket and input stream
    private Socket socket = null;
    private ServerSocket server = null;
    private BufferedSource in = null;
    private BufferedSink out = null;

    // constructor with port
    public Server3(int port) {
        //regist

        // starts server and waits for a connection
        try {
            server = new ServerSocket(port);
            System.out.println("Server started");

            System.out.println("Waiting for a client ...");

            socket = server.accept();
            System.out.println("Client accepted");

            // takes input from the client socket
            //in = Okio.buffer(Okio.source(socket.getInputStream()));
            out = Okio.buffer(Okio.sink(socket));
            in = Okio.buffer(Okio.source(socket));

            String line = "";

            // reads message from client until "Over" is sent
            while (!line.equals("Over")) {
                BufferedSource peek = in.peek();
                if(!peek.request(8)){ //blocked all the time
                    System.out.println("message not all reached.");
                    continue;
                }
                peek.skip(4);
                int len = peek.readInt();
                if(!peek.request(len)){
                    System.out.println("message len = " + len + " not all reached.");
                    continue;
                }
                in.skip(8);
                byte[] bytes = in.readByteArray(len);
                System.out.println("message reached:  " + new String(bytes, StandardCharsets.UTF_8));
            }
         //this is ok
        /* while (!line.equals("Over")) {
                in.skip(4);
                int len = in.readInt();
                if(!in.request(len)){
                    System.out.println("data not reached.");
                    break;
                }
                byte[] bytes = in.readByteArray(len);
[...]

PR #596, opened 2019-04-28
Peek the minimum number of bytes to return a result

swankjesse:
The contract for read() is to return at least 1 byte, but we were
attempting to read the full request. That caused us to prematurely
do a blocking read on the upstream source.

This fixes that code to defer the blocking read.

Closes: https://github.com/square/okio/issues/594

### DIFF
--- a/okio/src/main/kotlin/okio/PeekSource.kt
+++ b/okio/src/main/kotlin/okio/PeekSource.kt
@@ -36,6 +36,7 @@ internal class PeekSource(
   private var pos = 0L
 
   override fun read(sink: Buffer, byteCount: Long): Long {
+    require(byteCount >= 0) { "byteCount < 0: $byteCount" }
     check(!closed) { "closed" }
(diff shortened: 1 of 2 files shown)


okio/src/main/kotlin/okio/PeekSource.kt
```diff
@@ -36,15 +36,17 @@ internal class PeekSource(
...
     // Source becomes invalid if there is an expected Segment and it and the expected position
     // do not match the current head and head position of the upstream buffer
     check(expectedSegment == null ||
       expectedSegment === buffer.head && expectedPos == buffer.head!!.pos) {
       "Peek source is invalid because upstream source was used"
     }
+    if (byteCount == 0L) return 0L
+    if (!upstream.request(pos + 1)) return -1L
 
-    upstream.request(pos + byteCount)
```
swankjesse: this was the bug. We were attempting to satisfy the biggest request possible, but the caller would have been happy with just one byte.

PR #603, opened 2019-04-30
Fix snapshot publishing command

Egorand:
Note: this will publish snapshots for all targets


Egorand:
No since we're not publishing samples anymore, I'll update the PR

PR #605, opened 2019-04-30
Add RELEASING.md

Egorand:
I think it might be helpful to have to ensure we're running correct tasks when releasing the multiplatform version of the library.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,24 @@
+Releasing
+=========
+
+ 1. Change the version in `gradle.properties` to a non-SNAPSHOT version.
+ 2. Update the `CHANGELOG.md` for the impending release.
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -0,0 +1,24 @@
+Releasing
+=========
+
+ 1. Change the version in `gradle.properties` to a non-SNAPSHOT version.
+ 2. Update the `CHANGELOG.md` for the impending release.
+ 3. Update the `README.md` with the new version.
+ 4. `git commit -am "Prepare for release X.Y.Z."` (where X.Y.Z is the new version)
+ 5. `./gradlew clean publishJvmPublicationToMavenRepository`.
```
Egorand: Publishing JVM only for now, will update once we start publishing all targets

PR #616, opened 2019-06-19
Add Buffer implementation to Kotlin/JS and Kotlin/Native

bnorm:
Cut-pasted all the Buffer methods which don't rely on Java specifics to Kotlin/Common. Also copied the buffer unit tests but left the existing tests as is. Want to make sure we maintain that coverage during this transition.

 I'll go through and point out anything I did which might be questionable.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/-Util.kt
+++ b/okio/src/commonMain/kotlin/okio/-Util.kt
@@ -18,6 +18,7 @@
 
 package okio
 
+import okio.internal.HEX_DIGITS
 import kotlin.jvm.JvmName
 
 internal fun checkOffsetAndCount(size: Long, offset: Long, byteCount: Long) {
@@ -87,4 +88,29 @@ internal fun arrayRangeEquals(
     if (a[i + aOffset] != b[i + bOffset]) return false
   }
   return true
+}
+
+internal fun Byte.toHexString(): String {
+  val result = CharArray(4)
+  result[0] = '0'
+  result[1] = 'x'
+  result[2] = HEX_DIGITS[this shr 4 and 0xf]
+  result[3] = HEX_DIGITS[this       and 0xf] // ktlint-disable no-multi-spaces
+  return String(result)
+}
+
+internal fun Int.toHexString(): String {
+  // Should this trim leading 0's?
+  val result = CharArray(10)
+  result[0] = '0'
+  result[1] = 'x'
+  result[2] = HEX_DIGITS[this shr 28 and 0xf]
+  result[3] = HEX_DIGITS[this shr 24 and 0xf]
+  result[4] = HEX_DIGITS[this shr 20 and 0xf]
+  result[5] = HEX_DIGITS[this shr 16 and 0xf]
+  result[6] = HEX_DIGITS[this shr 12 and 0xf]
+  result[7] = HEX_DIGITS[this shr 8  and 0xf] // ktlint-disable no-multi-spaces
+  result[8] = HEX_DIGITS[this shr 4  and 0xf] // ktlint-disable no-multi-spaces
+  result[9] = HEX_DIGITS[this        and 0xf] // ktlint-disable no-multi-spaces
+  return String(result)
 }
--- a/okio/src/commonMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/commonMain/kotlin/okio/SegmentPool.kt
@@ -15,10 +15,13 @@
  */
 package okio
 
+import kotlin.native.concurrent.ThreadLocal
+
 /**
  * A collection of unused segments, necessary to avoid GC churn and zero-fill.
  * This pool is a thread-safe static singleton.
  */
+@ThreadLocal
 internal object SegmentPool {
   /** The maximum number of bytes to pool.  */
   // TODO: Is 64 KiB a good maximum size? Do we ever have that many idle segments?
--- a/okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
@@ -23,11 +23,12 @@ import okio.processUtf16Chars
 // to everything else. Putting them in this file, `-Utf8.kt`, makes them invisible to
 // Java but still visible to Kotlin.
 
-fun ByteArray.commonToUtf8String(): String {
-  val chars = CharArray(size)
+fun ByteArray.commonToUtf8String(offset: Int = 0, byteCount: Int = size): String {
+  require(offset + byteCount <= size) { "offset=$offset byteCount=$byteCount size=$size" }
+  val chars = CharArray(byteCount)
 
   var length = 0
-  processUtf16Chars(0, size) { c ->
+  processUtf16Chars(offset, offset + byteCount) { c ->
     chars[length++] = c
   }
 
--- a/okio/src/commonMain/kotlin/okio/internal/Buffer.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/Buffer.kt
@@ -13,15 +13,213 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// TODO move to Buffer class: https://youtrack.jetbrains.com/issue/KT-20427
+@file:Suppress("NOTHING_TO_INLINE")
+
 package okio.internal
 
 import okio.Buffer
 import okio.ByteString
 import okio.EOFException
+import okio.Options
+import okio.REPLACEMENT_CODE_POINT
 import okio.Segment
 import okio.SegmentPool
+import okio.Sink
+import okio.Source
+import okio.and
+import okio.asUtf8ToByteArray
 import okio.checkOffsetAndCount
+import okio.toHexString
 import okio.minOf
+
+private val DIGITS = "0123456789abcdef".asUtf8ToByteArray()
+
+/**
+ * Returns true if the range within this buffer starting at `segmentPos` in `segment` is equal to
+ * `bytes[bytesOffset..bytesLimit)`.
+ */
+internal fun rangeEquals(
+  segment: Segment,
+  segmentPos: Int,
+  bytes: ByteArray,
+  bytesOffset: Int,
+  bytesLimit: Int
+): Boolean {
+  var segment = segment
+  var segmentPos = segmentPos
+  var segmentLimit = segment.limit
+  var data = segment.data
+
+  var i = bytesOffset
+  while (i < bytesLimit) {
+    if (segmentPos == segmentLimit) {
+      segment = segment.next!!
+      data = segment.data
+      segmentPos = segment.pos
+      segmentLimit = segment.limit
+    }
+
+    if (data[segmentPos] != bytes[i]) {
+      return false
+    }
+
+    segmentPos++
+    i++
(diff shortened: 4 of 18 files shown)


okio/src/commonMain/kotlin/okio/SegmentPool.kt
```diff
@@ -15,10 +15,13 @@
  */
 package okio
 
+import kotlin.native.concurrent.ThreadLocal
+
 /**
  * A collection of unused segments, necessary to avoid GC churn and zero-fill.
  * This pool is a thread-safe static singleton.
  */
+@ThreadLocal
```
bnorm: Required to avoid a bunch of errors in Kotlin/Native unit tests. If this isn't desired I will look into alternatives.

okio/src/commonTest/kotlin/okio/CommonOptionsTest.kt
```diff
@@ -16,20 +16,19 @@
...
 import okio.ByteString.Companion.encodeUtf8
-import okio.TestUtil.bufferWithSegments
-import org.assertj.core.api.Assertions.assertThat
-import org.junit.Test
+import kotlin.test.Test
+import kotlin.test.assertEquals
 import kotlin.test.fail
 
-class OptionsTest {
+class CommonOptionsTest {
```
bnorm: Moving this test class to Kotlin/Common was pretty straight forward. I can create a copy instead and leave the original if we want.

okio/src/commonTest/kotlin/okio/CommonBufferTest.kt
```diff
@@ -0,0 +1,442 @@
...
+    assertEquals(2, buffer.size)
+    assertEquals("ef", buffer.readUtf8(2))
+    assertEquals(0, buffer.size)
+    assertFailsWith<EOFException> {
+      buffer.readUtf8(1)
+    }
+  }
+
+  /** Buffer's toString is the same as ByteString's.  */
+  @Ignore @Test fun bufferToString() {
```
bnorm: Buffer toString requires SegmentedByteString which hasn't moved to Kotlin/Common yet. I moved this unit test but marked it as ignored until SegmentedByteString is moved.

okio/src/commonMain/kotlin/okio/-Util.kt
```diff
@@ -88,3 +89,28 @@ internal fun arrayRangeEquals(
...
+  val result = CharArray(4)
+  result[0] = '0'
+  result[1] = 'x'
+  result[2] = HEX_DIGITS[this shr 4 and 0xf]
+  result[3] = HEX_DIGITS[this       and 0xf] // ktlint-disable no-multi-spaces
+  return String(result)
+}
+
+internal fun Int.toHexString(): String {
+  // Should this trim leading 0's?
```
bnorm: I can improve these implementations if we want to trim leading `0`s. I went for a fast and dirty implementation at first.

okio/src/commonTest/kotlin/okio/CommonBufferTest.kt
JakeWharton: Can you leave this comment in the code so it's clear what the blocker is

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -242,3 +568,667 @@ internal fun Buffer.commonRead(sink: ByteArray, offset: Int, byteCount: Int): In
...
+    return false
+  }
+  for (i in 0 until byteCount) {
+    if (this[offset + i] != bytes[bytesOffset + i]) {
+      return false
+    }
+  }
+  return true
+}
+
```
Egorand: nit: a lot of empty lines here

okio/src/commonMain/kotlin/okio/-Util.kt
swankjesse: Thumbs up. You need to position the `x` where the last leading `0` is, and call the String constructor that takes an offset and a count.

swankjesse: Also unclear whether bytes read nicer as `0x0f` or `0xf`. For ints I think trimming leading zeros is good.

okio/src/commonMain/kotlin/okio/SegmentPool.kt
swankjesse: clever fix

swankjesse: Tempted to make it use actual ThreadLocals on JVM

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
```diff
@@ -23,11 +23,12 @@ import okio.processUtf16Chars
 // to everything else. Putting them in this file, `-Utf8.kt`, makes them invisible to
 // Java but still visible to Kotlin.
 
-fun ByteArray.commonToUtf8String(): String {
-  val chars = CharArray(size)
+fun ByteArray.commonToUtf8String(offset: Int = 0, byteCount: Int = size): String {
```
swankjesse: I would have expected (begin, end) instead of (offset, count) here, but I suspect you’re tracking something else

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
```diff
@@ -23,11 +23,12 @@ import okio.processUtf16Chars
...
+  require(offset + byteCount <= size) { "offset=$offset byteCount=$byteCount size=$size" }
+  val chars = CharArray(byteCount)
 
   var length = 0
-  processUtf16Chars(0, size) { c ->
+  processUtf16Chars(offset, offset + byteCount) { c ->
```
swankjesse: .... ^ cause then this is just (begin, end)

okio/src/commonTest/kotlin/okio/util.kt
```diff
@@ -0,0 +1,50 @@
+package okio
```
swankjesse: nit: copyright thingy

okio/src/commonTest/kotlin/okio/util.kt
```diff
@@ -0,0 +1,50 @@
...
+
+import kotlin.random.Random
+
+fun Char.repeat(count: Int): String {
+  return toString().repeat(count)
+}
+
+fun segmentSizes(buffer: Buffer): List<Int> {
+  return generateSequence(buffer.head) { segment -> segment.next.takeIf { it !== buffer.head }}
```
swankjesse: for tests this is fine, but for production code I prefer the obvious performance consequences of procedural code

okio/src/commonTest/kotlin/okio/CommonOptionsTest.kt
```diff
@@ -16,20 +16,19 @@
...
   /** Confirm that options prefers the first-listed option, not the longest or shortest one. */
   @Test fun optionOrderTakesPrecedence() {
     assertSelect("abcdefg", 0, "abc", "abcdef")
     assertSelect("abcdefg", 0, "abcdef", "abc")
   }
 
   @Test fun simpleOptionsTrie() {
-    assertThat(utf8Options("hotdog", "hoth", "hot").trieString()).isEqualTo("""
+    assertEquals(utf8Options("hotdog", "hoth", "hot").trieString(), """
```
swankjesse: We need somebody to make assertJ for multiplatform!

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -242,3 +568,667 @@ internal fun Buffer.commonRead(sink: ByteArray, offset: Int, byteCount: Int): In
...
+  val scanLength = if (limit == Long.MAX_VALUE) Long.MAX_VALUE else limit + 1L
+  val newline = indexOf('\n'.toByte(), 0L, scanLength)
+  if (newline != -1L) return readUtf8Line(newline)
+  if (scanLength < size &&
+    this[scanLength - 1] == '\r'.toByte() &&
+    this[scanLength] == '\n'.toByte()) {
+    return readUtf8Line(scanLength) // The line was 'limit' UTF-8 bytes followed by \r\n.
+  }
+  val data = Buffer()
+  copyTo(data, 0, okio.minOf(32, size))
```
swankjesse: IntelliJ warns me that `okio.` is unnecessary here

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -242,3 +568,667 @@ internal fun Buffer.commonRead(sink: ByteArray, offset: Int, byteCount: Int): In
...
+    if (this[offset + i] != bytes[bytesOffset + i]) {
+      return false
+    }
+  }
+  return true
+}
+
+
+
+internal inline fun Buffer.commonEquals(other: Any?): Boolean {
```
swankjesse: JVM can delegate to this?

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -242,3 +568,667 @@ internal fun Buffer.commonRead(sink: ByteArray, offset: Int, byteCount: Int): In
...
+      sb = sb.next!!
+      posB = sb.pos
+    }
+    pos += count
+  }
+
+  return true
+}
+
+internal inline fun Buffer.commonHashCode(): Int {
```
swankjesse: JVM can delegate to this?

okio/src/commonMain/kotlin/okio/-Util.kt
swankjesse: Actually I think we probably want to do something weirder for these.

We should delegate! return Buffer().writeUtf8("0x").writeHexadecimalUnsignedLong(this).readUtf8()

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
bnorm: This is actually a different method which is already `[begin, end)`.

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
bnorm: I think this is probably a wash either way. Either I have to calculate the end index for (offset, count) or I have to calculate the size for (begin, end). I'll change it for the follow up PR though, just in case you like it better.

okio/src/commonMain/kotlin/okio/-Util.kt
bnorm: Sadly `writeHexadecimalUnsignedLong` is still TODO. It requires figuring out how many characters the number is going to be which currently relies on JVM specific methods. I'm hoping to use the same binary tree style approach used by `commonWriteDecimalLong` instead.

swankjesse: FWIW, you *might* be able to do numberOfTrailingZeros and highestOneBit yourself pretty easily, there’s a trick in [Hacker’s Delight](https://hackersdelight.org/) or related sites.

http://aggregate.org/MAGIC/

swankjesse: https://github.com/square/okio/issues/620

PR #618, opened 2019-06-20
Kotlin 1.3.40

Egorand:


okio/jvm/jmh/build.gradle
```diff
@@ -23,10 +23,10 @@ jmh {
 }
 
 dependencies {
-  compile project(path: ':okio', configuration: 'runtimeElements')
+  compile project(path: ':okio', configuration: 'jvmRuntimeElements')
```
JakeWharton: Is this needed? I thought you changed all these to not specify the configuration explicitly

Egorand: Nope, it's not needed here, fixed

PR #619, opened 2019-06-20
Kotlin/Common Buffer implementation cleanup

bnorm:


okio/src/commonMain/kotlin/okio/-Util.kt
```diff
@@ -112,5 +113,18 @@ internal fun Int.toHexString(): String {
...
+
+  // Find the first non-zero index
+  var i = 2
+  while (i < 10) {
+    if (result[i] != '0') break
+    i++
+  }
+
+  // truncate on a whole byte and shift to start of "0x"
+  i -= 2 + (i % 2)
```
bnorm: So this does something a little fancy: it only truncates the string on the byte boundary. I was just kind of having fun with this so if you just want to truncate on the first non-zero character, it's really easy to change.

For example
```kotlin
fun main() {
  println(0x12345678.toHexString())
  println(0x1234567.toHexString())
  println(0x123456.toHexString())
  println(0x12345.toHexString())
  println(0x1234.toHexString())
  println(0x123.toHexString())
  println(0x12.toHexString())
  println(0x1.toHexString())
  println(0x0.toHexString())
}
```
will produce
```
0x12345678
0x01234567
0x123456
0x012345
0x1234
0x0123
0x12
0x01
0x00
```

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -40,7 +40,7 @@ private val DIGITS = "0123456789abcdef".asUtf8ToByteArray()
  * Returns true if the range within this buffer starting at `segmentPos` in `segment` is equal to
  * `bytes[bytesOffset..bytesLimit)`.
  */
-internal fun rangeEquals(
+internal inline fun rangeEquals(
```
bnorm: I went through and marked everything as `inline`. Not sure it really makes sense for a couple of these methods (like this one) so let me know if I should remove it from any of these methods.

okio/src/commonMain/kotlin/okio/-Util.kt
```diff
@@ -91,6 +91,8 @@ internal fun arrayRangeEquals(
 }
 
 internal fun Byte.toHexString(): String {
+  if (this == 0.toByte()) return "0x00"
```
swankjesse: is this an optimization?

okio/src/commonMain/kotlin/okio/-Util.kt
```diff
@@ -100,10 +102,9 @@ internal fun Byte.toHexString(): String {
 }
 
 internal fun Int.toHexString(): String {
-  // Should this trim leading 0's?
+  if (this == 0) return "0x00"
```
swankjesse: is this an optimization?

okio/src/commonMain/kotlin/okio/-Util.kt
swankjesse: I think this is cute and clever and awesome. But I don’t think we should ship it ’cause an Int isn’t a `byte[]`, even though it serves that role well.

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
```diff
@@ -16,19 +16,22 @@
...
 // Java but still visible to Kotlin.
 
-fun ByteArray.commonToUtf8String(offset: Int = 0, byteCount: Int = size): String {
-  require(offset + byteCount <= size) { "offset=$offset byteCount=$byteCount size=$size" }
-  val chars = CharArray(byteCount)
+fun ByteArray.commonToUtf8String(beginIndex: Int = 0, endIndex: Int = size): String {
+  if (beginIndex < 0 || endIndex > size || beginIndex > endIndex) {
+    throw ArrayIndexOutOfBoundsException("size=$size beginIndex=$beginIndex endIndex=$endIndex")
+  }
+  val chars = CharArray(endIndex - beginIndex)
```
swankjesse: ahh, I see what you mean

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
swankjesse: cool. I think our rule should be “an inlined function shouldn’t be more than 5 lines or be used more than 2x”

swankjesse: ... or 10 lines, 3 times, or something. But a firm rule that’s easy to follow.

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -224,7 +224,7 @@ internal fun Buffer.selectPrefix(options: Options, selectTruncated: Boolean = fa
 // TODO Kotlin's expect classes can't have default implementations, so platform implementations
 // have to call these functions. Remove all this nonsense when expect class allow actual code.
 
-internal fun Buffer.commonCopyTo(
+internal inline fun Buffer.commonCopyTo(
```
swankjesse: all these _common_ methods show up 1x so that’s perfect

okio/src/commonTest/kotlin/okio/util.kt
```diff
@@ -7,9 +22,15 @@ fun Char.repeat(count: Int): String {
 }
 
 fun segmentSizes(buffer: Buffer): List<Int> {
-  return generateSequence(buffer.head) { segment -> segment.next.takeIf { it !== buffer.head }}
-    .map { segment -> segment.limit - segment.pos }
-    .toList()
+  var segment = buffer.head ?: return emptyList()
+
+  val sizes = mutableListOf(segment.limit - segment.pos)
```
swankjesse: thank you for indulging me

okio/src/commonMain/kotlin/okio/-Util.kt
bnorm: No problem. It was late, I was having fun. 😃

okio/src/commonMain/kotlin/okio/-Util.kt
bnorm: No, I haven't benchmarked it. Happy to remove if you think it's not needed.

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
bnorm: So I should remove `inline` from the non-commonX methods here unless they are a higher-order function?

okio/src/commonMain/kotlin/okio/-Util.kt
swankjesse: I don’t like these. Here’s why!
https://smallwig.blogspot.com/2010/03/little-optimization-that-couldnt.html

bnorm: Good to know! Thanks for the learning!

PR #621, opened 2019-06-20
Delete custom JS testing setup in favor of built-in test runner

Egorand:


Egorand:
This doesn't work at the moment, filed https://youtrack.jetbrains.com/issue/KT-32124

PR #622, opened 2019-06-21
Move SegmentedByteString implementation to Kotlin/Common

bnorm:


okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -1232,3 +1233,47 @@ internal inline fun Buffer.commonHashCode(): Int {
...
+/** Returns an immutable copy of this buffer as a byte string.  */
+internal inline fun Buffer.commonSnapshot(): ByteString {
+  check(size <= Int.MAX_VALUE) { "size > Int.MAX_VALUE: $size" }
+  return snapshot(size.toInt())
+}
+
+/** Returns an immutable copy of the first `byteCount` bytes of this buffer as a byte string. */
+internal inline fun Buffer.commonSnapshot(byteCount: Int): ByteString {
+  if (byteCount == 0) return ByteString.EMPTY
+  checkOffsetAndCount(size, 0, byteCount.toLong())
```
bnorm: Moved the implementation of `SegmentedByteString.of` directory into this method to avoid extra expect/actual's.

okio/src/commonMain/kotlin/okio/ByteString.kt
```diff
@@ -144,7 +144,6 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
   override fun toString(): String
 
   companion object {
-    // TODO null in JS and Native - https://youtrack.jetbrains.com/issue/KT-26497
```
bnorm: I was able to work around this issue by removing `COMMON_EMPTY` and constructing the ByteString directly in the actual implementations.

okio/src/commonMain/kotlin/okio/internal/SegmentedByteString.kt
```diff
@@ -0,0 +1,223 @@
...
+package okio.internal
+
+import okio.Buffer
+import okio.ByteString
+import okio.Segment
+import okio.SegmentedByteString
+import okio.arrayRangeEquals
+import okio.checkOffsetAndCount
+
+internal fun IntArray.binarySearch(value: Int, fromIndex: Int, toIndex: Int): Int {
```
bnorm: Common implementation of binary search. I can actual expect this if we want a platform native implementation.

swankjesse: I slightly prefer as you have it. It’s more code on the JVM, but we’re testing the same code everywhere.

okio/src/commonMain/kotlin/okio/internal/SegmentedByteString.kt
```diff
@@ -0,0 +1,223 @@
...
+  var left = fromIndex
+  var right = toIndex - 1
+
+  while (left <= right) {
+    val mid = (left + right) ushr 1 // protect from overflow
```
swankjesse: nice

PR #625, opened 2019-06-27
[PoC] Support for HashingSink/Source on Windows using OpenSSL

bnorm:
I was experimenting with Kotlin/Native cinterop and thought I would share what I have in case someone can learn from it. At some point if we want to support hashing and compression in Kotlin/Native this will probably need to get formalized.

This should easily generalize for Mac and Linux since both support the OpenSSL library (if this is even the library that should be used) but how the duplicate code is managed will be interesting. The Gradle file in the [libcurl sample](https://github.com/JetBrains/kotlin-native/tree/master/samples/libcurl) in the Kotlin/Native repository only configures the current platform under a constant name rather than configure all platforms to work around this. This might be something to consider.

One thing I learned is that Kotlin/Native cinterop doesn't generate macros and OpenSSL likes to alias functions using macros. Kept trying to figure out why the functions in the OpenSSL samples I found didn't exist until I actually read the header file and saw they were macros and just aliases for other functions.

I can keep going with this and get Linux supported as well if we actually want to merge it. For Mac and iOS support (which is probably what you are most interested in) someone else might have to explore that since I'm unfamiliar and don't have access to a Mac for development.


okio/src/winX64Test/kotlin/okio/HashingSourceTest.kt
```diff
@@ -0,0 +1,125 @@
...
+    assertEquals(3L, hashingSource.read(sink, Long.MAX_VALUE))
+    val hash_abc = hashingSource.hash
+    assertEquals(SHA256_abc, hash_abc, "abc")
+    source.writeUtf8("def")
+    assertEquals(3L, hashingSource.read(sink, Long.MAX_VALUE))
+    assertEquals(SHA256_def, hashingSource.hash, "def")
+    assertEquals(SHA256_abc, hash_abc, "abc")
+  }
+
+//  @Test fun multipleSegments() {
```
swankjesse: ?

okio/src/winX64Main/kotlin/okio/HashingSink.kt
```diff
@@ -0,0 +1,149 @@
...
+
+  override fun close() {
+    EVP_MD_CTX_free(ctx)
+    delegate.close()
+  }
+
+  /**
+   * Returns the hash of the bytes accepted thus far and resets the internal state of this sink.
+   *
+   * **Warning:** This method is not idempotent. Each time this method is called its
```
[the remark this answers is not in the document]
bnorm: I was wondering the same thing. Lessons learned, right?

okio/src/winX64Test/kotlin/okio/HashingSourceTest.kt
bnorm: `r32k` wasn't implemented yet. I added it in the latest update.

bnorm:
Replaced by custom hashing implementations.

PR #627, opened 2019-07-04
Share two copies of hex digits as bytes

JakeWharton:


okio/src/commonMain/kotlin/okio/internal/ByteString.kt
```diff
@@ -51,16 +51,16 @@ internal inline fun ByteString.commonBase64(): String = data.encodeBase64()
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun ByteString.commonBase64Url() = data.encodeBase64(map = BASE64_URL_SAFE)
 
-internal val HEX_DIGITS =
+internal val HEX_DIGIT_CHARS =
```
JakeWharton: Should this use `HEX_DIGIT_BYTES` and just call `toChar()`?

swankjesse: tomato/tomato

PR #631, opened 2019-07-24
Fix the deploy script to restore 1.x docs

swankjesse:


swankjesse:
Self merging docs PR.

PR #633, opened 2019-07-24
Fix the hash for the 1.x docs

swankjesse:


swankjesse:
Self-merging docs change. Reviews welcome.

PR #635, opened 2019-07-25
Promote docs to commonMain wherever possible

swankjesse:


okio/src/commonMain/kotlin/okio/BufferedSink.kt
```diff
@@ -15,44 +15,298 @@
...
+   * assertEquals(0, b1.buffer().size());
+   * assertEquals(0, b0.buffer().size());
+   *
+   * b2.flush();
+   * assertEquals(0, b2.buffer().size());
+   * assertEquals(0, b1.buffer().size());
+   * assertEquals(5, b0.buffer().size());
+   * ```
+   */
+  override fun flush()
```
swankjesse: Behavior change! This is newly added to common

okio/src/commonMain/kotlin/okio/BufferedSource.kt
```diff
@@ -15,84 +15,520 @@
...
+   * line break is optional.
+   */
   fun readUtf8Line(): String?
 
+  /**
+   * Removes and returns characters up to but not including the next line break. A line break is
+   * either `"\n"` or `"\r\n"`; these characters are not included in the result.
+   *
+   * **On the end of the stream this method throws.** Every call must consume either
+   * '\r\n' or '\n'. If these characters are absent in the stream, an [java.io.EOFException]
```
JakeWharton: Exception type

okio/src/commonMain/kotlin/okio/BufferedSource.kt
```diff
@@ -15,84 +15,520 @@
...
+   * assertEquals("12345", buffer.readUtf8LineStrict(5));
+   * ```
+   */
   fun readUtf8LineStrict(limit: Long): String
 
+  /**
+   * Removes and returns a single UTF-8 code point, reading between 1 and 4 bytes as necessary.
+   *
+   * If this source is exhausted before a complete code point can be read, this throws an
+   * [java.io.EOFException] and consumes no input.
```
JakeWharton: Exception type

swankjesse:
Thanks and agreed. We also use backticks in a few places where square braces are available.

I'll fix the Java stuff in follow-up!

PR #637, opened 2019-07-26
Try expect interfaces for Closeable, Flushable

swankjesse:
This fails due to a crashing bug in the Kotlin/JS compiler.

https://youtrack.jetbrains.com/issue/KT-28927


JakeWharton:
What's the motivation for this change? Is there Okio-specific common code that abstracts over a Closeable or Flushable?

In SQLDelight we did Closeable but now I regret it and wish we just offered a specialized `SqlStatement.use` which delegated to `Closeable` on the JVM and did the dance manually on JS/native.

swankjesse:
I wanna move Source and Sink to common properly rather than using expect classes.

JakeWharton:
Is there an advantage to doing so aside from our own maintenance burden?

swankjesse:
Just our maintenance burden! Closing for now.

PR #638, opened 2019-07-28
Copy RealBufferedSink and RealBufferedSource to Kotlin/Common

bnorm:
Copied the BufferedSourceTest and BufferedSinkTest files to Kotlin/Common. This involved converting them to Kotlin (thanks IntelliJ), changing parameterized unit test mechanism, removing Java specific unit tests, adding the `Buffer#copy()` method as an alternative to Java specific `Buffer#clone()`, and fixing a bug.

I've also moved RealBufferedSource and RealBufferedSink to Kotlin/Common but wanted to start this pull request with the unit test move.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/-Util.kt
+++ b/okio/src/commonMain/kotlin/okio/-Util.kt
@@ -91,36 +91,62 @@ internal fun arrayRangeEquals(
 }
 
 internal fun Byte.toHexString(): String {
-  val result = CharArray(4)
-  result[0] = '0'
-  result[1] = 'x'
-  result[2] = HEX_DIGIT_CHARS[this shr 4 and 0xf]
-  result[3] = HEX_DIGIT_CHARS[this       and 0xf] // ktlint-disable no-multi-spaces
+  val result = CharArray(2)
+  result[0] = HEX_DIGIT_CHARS[this shr 4 and 0xf]
+  result[1] = HEX_DIGIT_CHARS[this       and 0xf] // ktlint-disable no-multi-spaces
   return String(result)
 }
 
 internal fun Int.toHexString(): String {
-  val result = CharArray(10)
-  result[2] = HEX_DIGIT_CHARS[this shr 28 and 0xf]
-  result[3] = HEX_DIGIT_CHARS[this shr 24 and 0xf]
-  result[4] = HEX_DIGIT_CHARS[this shr 20 and 0xf]
-  result[5] = HEX_DIGIT_CHARS[this shr 16 and 0xf]
-  result[6] = HEX_DIGIT_CHARS[this shr 12 and 0xf]
-  result[7] = HEX_DIGIT_CHARS[this shr 8  and 0xf] // ktlint-disable no-multi-spaces
-  result[8] = HEX_DIGIT_CHARS[this shr 4  and 0xf] // ktlint-disable no-multi-spaces
-  result[9] = HEX_DIGIT_CHARS[this        and 0xf] // ktlint-disable no-multi-spaces
+  if (this == 0) return "0" // Required as code below does not handle 0
+
+  val result = CharArray(8)
+  result[0] = HEX_DIGIT_CHARS[this shr 28 and 0xf]
+  result[1] = HEX_DIGIT_CHARS[this shr 24 and 0xf]
+  result[2] = HEX_DIGIT_CHARS[this shr 20 and 0xf]
+  result[3] = HEX_DIGIT_CHARS[this shr 16 and 0xf]
+  result[4] = HEX_DIGIT_CHARS[this shr 12 and 0xf]
+  result[5] = HEX_DIGIT_CHARS[this shr 8  and 0xf] // ktlint-disable no-multi-spaces
+  result[6] = HEX_DIGIT_CHARS[this shr 4  and 0xf] // ktlint-disable no-multi-spaces
+  result[7] = HEX_DIGIT_CHARS[this        and 0xf] // ktlint-disable no-multi-spaces
 
   // Find the first non-zero index
-  var i = 2
-  while (i < 10) {
+  var i = 0
+  while (i < result.size) {
     if (result[i] != '0') break
     i++
   }
 
-  // shift and insert "0x"
-  i -= 2
-  result[i] = '0'
-  result[i + 1] = 'x'
+  return String(result, i, result.size - i)
+}
 
-  return String(result, i, 10 - i)
+internal fun Long.toHexString(): String {
+  if (this == 0L) return "0" // Required as code below does not handle 0
+
+  val result = CharArray(16)
+  result[ 0] = HEX_DIGIT_CHARS[(this shr 60 and 0xf).toInt()] // ktlint-disable no-multi-spaces
+  result[ 1] = HEX_DIGIT_CHARS[(this shr 56 and 0xf).toInt()] // ktlint-disable no-multi-spaces
+  result[ 2] = HEX_DIGIT_CHARS[(this shr 52 and 0xf).toInt()] // ktlint-disable no-multi-spaces
+  result[ 3] = HEX_DIGIT_CHARS[(this shr 48 and 0xf).toInt()] // ktlint-disable no-multi-spaces
+  result[ 4] = HEX_DIGIT_CHARS[(this shr 44 and 0xf).toInt()] // ktlint-disable no-multi-spaces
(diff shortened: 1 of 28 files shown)


okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -1408,6 +1408,22 @@ internal inline fun Buffer.commonHashCode(): Int {
   return result
 }
 
+internal inline fun Buffer.commonCopy(): Buffer {
+  val result = Buffer()
+  if (size == 0L) return result
+
+  result.head = head!!.sharedCopy()
```
JakeWharton: This problem persists in a few places, but I'd love to see locals being leveraged here to avoid so much bang-bangs.
```kotlin
val myHead = head!!
val copyHead = myHead.sharedCopy()
// .. use myHead and copyHead without !!
```
Ideally we bang-bang each unique reference once. It clutters up the source code a bit, but has _zero_ impact on bytecode.

(I mean it actually has a negative size impact on bytecode since we don't emit redundant calls to `Intrinsics.checkNotNull` et. al., but compared to Java is has zero impact)

okio/src/commonTest/kotlin/okio/BufferedSinkTest.kt
```diff
@@ -0,0 +1,268 @@
...
+package okio
+
+import okio.ByteString.Companion.decodeHex
+import kotlin.math.pow
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.fail
+
+class CommonBufferedSinkTest {
+  class Buffer : AbstractBufferedSinkTest(BufferedSinkFactory.BUFFER)
```
JakeWharton: Move these lines to just be the first inside `AbstractBufferedSInkTest`? Or just put them top-level?

okio/src/commonTest/kotlin/okio/BufferedSinkTest.kt
```diff
@@ -0,0 +1,268 @@
...
+  // TODO implement RealBufferedSink
+  // class RealBufferedSink : AbstractBufferedSinkTest(BufferedSinkFactory.REAL_BUFFERED_SINK)
+}
+
+internal interface BufferedSinkFactory {
```
JakeWharton: Would make this its own file

okio/src/commonTest/kotlin/okio/BufferedSourceTest.kt
```diff
@@ -0,0 +1,1586 @@
...
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+import kotlin.test.fail
+import okio.ByteString.Companion.decodeHex
+import okio.ByteString.Companion.encodeUtf8
+import kotlin.test.Ignore
+
+class CommonBufferedSourceTest {
+  class Buffer : AbstractBufferedSourceTest(BufferedSourceFactory.BUFFER)
```
JakeWharton: Same recommendation here

okio/src/commonTest/kotlin/okio/BufferedSourceTest.kt
```diff
@@ -0,0 +1,1586 @@
...
+  // class OneByteAtATimeBufferedSource : AbstractBufferedSourceTest(BufferedSourceFactory.ONE_BYTE_AT_A_TIME_BUFFERED_SOURCE)
+  // TODO implement RealBufferedSource
+  // class OneByteAtATimeBuffer : AbstractBufferedSourceTest(BufferedSourceFactory.ONE_BYTE_AT_A_TIME_BUFFER)
+  // TODO implement RealBufferedSource
+  // class PeekBuffer : AbstractBufferedSourceTest(BufferedSourceFactory.PEEK_BUFFER)
+  // TODO implement RealBufferedSource
+  // class PeekBufferedSource : AbstractBufferedSourceTest(BufferedSourceFactory.PEEK_BUFFERED_SOURCE)
+}
+
+internal interface BufferedSourceFactory {
```
JakeWharton: Ditto, seems like it should be top-level file

okio/src/commonTest/kotlin/okio/util.kt
```diff
@@ -33,6 +34,17 @@ fun segmentSizes(buffer: Buffer): List<Int> {
   return sizes
 }
 
+fun assertByteArraysEquals(a: ByteArray, b: ByteArray) {
```
JakeWharton:
```suggestion
fun assertArrayEquals(a: ByteArray, b: ByteArray) {
```

Matches JUnit. I think I'll send a Kotlin change to add these methods to kotlin.test...

JakeWharton: https://youtrack.jetbrains.com/issue/KT-32996

okio/src/commonMain/kotlin/okio/Okio.kt
```diff
@@ -29,4 +29,11 @@ expect fun Source.buffer(): BufferedSource
 expect fun Sink.buffer(): BufferedSink
 
 /** Returns a sink that writes nowhere. */
-expect fun blackholeSink(): Sink
+expect fun blackholeSink(): Sink // expect/actual required for Java interop
```
bnorm: There might be a way to avoid the expect/actual and still conform to the required Java interop but I wasn't able to figure it out. My quick attempt didn't seem to work because of no file merging between this file and the Kotlin/JVM Okio.kt file.

bnorm:
Rebased on latest to include `BufferedSink.write(ByteString, offset, byteCount)` change and fix a few ktlint issues.

okio/src/commonMain/kotlin/okio/Okio.kt
```diff
@@ -29,4 +29,11 @@ expect fun Source.buffer(): BufferedSource
...
+
+internal class BlackholeSink : Sink {
```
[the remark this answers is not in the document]
swankjesse: Sure!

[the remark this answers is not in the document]
swankjesse: ... for a follow-up PR!

okio/src/commonMain/kotlin/okio/-Util.kt
```diff
@@ -91,36 +91,62 @@ internal fun arrayRangeEquals(
 }
 
 internal fun Byte.toHexString(): String {
-  val result = CharArray(4)
```
swankjesse: note to self : behavior change here

okio/src/commonTest/kotlin/okio/ByteStringTest.kt
```diff
@@ -29,45 +29,12 @@ import kotlin.test.assertSame
...
-    val OKIO_ENCODER: ByteStringFactory = object : ByteStringFactory {
-      override fun decodeHex(hex: String) = hex.decodeHex()
-      override fun encodeUtf8(s: String) = ByteString.of(*s.commonAsUtf8ToByteArray())
-    }
-  }
-}
+class ByteStringTest : AbstractByteStringTest(ByteStringFactory.BYTE_STRING)
+class SegmentedByteStringTest : AbstractByteStringTest(ByteStringFactory.SEGMENTED_BYTE_STRING)
+class ByteStringOneBytePerSegmentTest : AbstractByteStringTest(ByteStringFactory.ONE_BYTE_PER_SEGMENT)
+class OkioEncoderTest : AbstractByteStringTest(ByteStringFactory.OKIO_ENCODER)
```
swankjesse: Nice!

okio/src/commonMain/kotlin/okio/Buffer.kt
```diff
@@ -109,6 +109,9 @@ expect class Buffer() : BufferedSource, BufferedSink {
 
   override fun writeHexadecimalUnsignedLong(v: Long): Buffer
 
+  /** Returns a deep copy of this buffer.  */
+  fun copy(): Buffer
```
swankjesse: Neat!

okio/src/commonMain/kotlin/okio/-Util.kt
swankjesse: the referencing exceptions all fixed. Dig.

okio/src/commonMain/kotlin/okio/Buffer.kt
swankjesse: ... I assume this is better than doing clone() across platforms?

PR #640, opened 2019-07-29
Remove Windows support, change iOS to klib

Egorand:


3flex:
Any particular reason support for Windows was removed?

I was hoping to use okio in a multiplatform project where Windows is a target.

PR #646, opened 2019-08-07
Convert Utf8Test to kotlin

louistsaitszho:
It's pretty much a line-to-line conversion. The 1 assert that got drop is in `sizeBoundsCheck()`, since `utf8Size` now takes a kotlin non null `String`.

```java
try {
  Utf8.size(null, 0, 0);
  fail();
} catch (IllegalArgumentException expected) {
}
```


Egorand:
We decided to not convert existing Java tests to Kotlin, they're still valuable as they help us ensure our API makes sense for Java consumers.

PR #648, opened 2019-08-15
Kotlin 1.3.50-eap

Egorand:
Wire would benefit from an Okio snapshot built against this version, so as long as there are no Okio releases planned in the immediate future, I'd love to merge this.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   ext.versions = [
-      'kotlin': '1.3.40',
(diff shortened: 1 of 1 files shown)


JakeWharton:
I don't quite get how Wire benefits since it's about to release and will need a stable version.

Egorand:
I mostly meant the WIP PRs that are blocked by bugs that were fixed in 1.3.50. We'll definitely release with 1.3.40.

PR #651, opened 2019-08-23
Add BufferedSink.write(ByteString, offset, byteCount)

bnorm:


bnorm:
I opted to keep the public signature like the existing `ByteArray` methods for consistency. As for the internal write method, I didn't really see a need for default parameters as updating all calling locations was just as trivial.

Happy to add default parameters if that is the long term goal.

swankjesse:
I’m not particularly worried about backporting. Anyone who builds with 2.x should deploy with 2.x. We just wanna make sure that if you build with 1.x you can drop in 2.x without recompiling.

PR #661, opened 2019-10-02
Use inline function to simplify AsyncTimeout enter-exit

bnorm:


okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -195,17 +149,35 @@ open class AsyncTimeout : Timeout() {
...
    */
   internal fun exit(cause: IOException): IOException {
     return if (!exit()) cause else newTimeoutException(cause)
   }
 
+  /**
+   * Surrounds [block] with calls to [enter] and [exit], throwing an exception from
+   * [newTimeoutException] if a timeout occurred.
+   */
+  internal inline fun <T> withTimeout(block: () -> T): T {
```
swankjesse: Love this

swankjesse: Should we make this public?

swankjesse: (Lets do the public/internal decision in follow-up)

bnorm: If this is made public, as an inline method, the 2 `exit` methods it calls will also need to be made public. Or they can be inlined into `withTimeout` as that is the only place they are now used.

Is it common for users to use `enter` and `exit` manually? Or is use of the prebuilt `source` and `sink` methods more common? It looks like OkHttp uses `enter` and `exit` in a couple different places; would OkHttp benefit from a `withTimeout` method?

swankjesse: Definitely don’t want too many exit methods on the public API. If everyone can use withTimeout() that’s ideal.

Whatever we do I think we’d want to do it so that OkHttp could take advantage.

PR #662, opened 2019-10-04
Make AsyncTimeout.withTimeout public

bnorm:
Follow up to #661 

It is possible to make internal functions available to an inline'd function using the `@PublishedApi` annotation. This effectively makes them part of the public API of the project but only usable through the the inline function.

We could take this one step farther and remove the 2 internal `exit` functions since they are only used from the `withTimeout` function. However, this would mean making `newTimeoutException` public as well.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -151,19 +151,22 @@ open class AsyncTimeout : Timeout() {
       throwOnTimeout = true
       return result
     } catch (e: IOException) {
-      throw if (!exit()) e else newTimeoutException(e)
+      throw if (!exit()) e else `access$newTimeoutException`(e)
     } finally {
       val timedOut = exit()
-      if (timedOut && throwOnTimeout) throw newTimeoutException(null)
+      if (timedOut && throwOnTimeout) throw `access$newTimeoutException`(null)
     }
   }
+
+  @PublishedApi // Binary compatible trampoline function
+  internal fun `access$newTimeoutException`(cause: IOException?) = newTimeoutException(cause)
 
   /**
    * Returns an [IOException] to represent a timeout. By default this method returns
    * [InterruptedIOException]. If [cause] is non-null it is set as the cause of the
    * returned exception.
    */
-  open fun newTimeoutException(cause: IOException?): IOException {
+  protected open fun newTimeoutException(cause: IOException?): IOException {
     val e = InterruptedIOException("timeout")
     if (cause != null) {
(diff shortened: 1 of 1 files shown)


swankjesse:
Wanna manually inline the two exit calls and the newTimeoutException() method too? That way the function just exercises public APIs.

bnorm:
Unfortunately we can't inline `newTimeoutException()` as it is overriden by `SocketAsyncTimeout` so a `SocketTimeoutException` can be thrown. I suspect other libraries outside of Okio also override `newTimeoutException()` to throw custom exceptions.

bnorm:
I think I'm actually in favor of inlining the two internal `exit` functions and making `newTimeoutException()` public. It's technically part of the API even though it's protected because `AsyncTimeout` is non-final. OkHttp overrides it to customize the exception thrown.

JakeWharton:
Agree

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -180,7 +163,7 @@ open class AsyncTimeout : Timeout() {
    * [InterruptedIOException]. If [cause] is non-null it is set as the cause of the
    * returned exception.
    */
-  protected open fun newTimeoutException(cause: IOException?): IOException {
+  open fun newTimeoutException(cause: IOException?): IOException {
```
swankjesse: I love this change in its current state.

But there’s a problem. I don‘t think this is a backwards-compatible change. Subclasses might have `protected` visibility, and widening it to `public` will break those.

The fix is pretty easy, just create a binary-public trampoline?

```
@Deprecated(level = DeprecationLevel.HIDDEN)
@JvmName("-newTimeoutException")
fun dashNewTimeoutException(cause: IOException) = newTimeoutException(cause)
```

bnorm: Good catch, thanks!

I had to do something a little different as `HIDDEN` level also makes it hidden to the current module.

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -139,42 +139,28 @@ open class AsyncTimeout : Timeout() {
...
-      throw exit(e)
+      throw if (!exit()) e else `access$newTimeoutException`(e)
     } finally {
-      exit(throwOnTimeout)
+      val timedOut = exit()
+      if (timedOut && throwOnTimeout) throw `access$newTimeoutException`(null)
     }
   }
 
+  @PublishedApi // Binary compatible trampoline function
```
swankjesse: perfect

PR #664, opened 2019-10-08
Add how to use snapshot section to readme

StefMa:
I found out that some people aren't that experienced with Gradle and don't know how to add the snapshot version to their project.
Therefore I've a little bit modified the section which gives the users a small hint how to add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -645,7 +645,20 @@ Our [change log][changelog] has release history.
 implementation("com.squareup.okio:okio:2.4.1")
 ```
 
-Snapshot builds are [available][snap].
+<details>
+   <summary>Snapshot builds are also available</summary>
+      
+```kotlin
+repositories {
+   maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
+}
+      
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -645,7 +645,20 @@ Our [change log][changelog] has release history.
 implementation("com.squareup.okio:okio:2.4.1")
 ```
 
-Snapshot builds are [available][snap].
+<details>
+   <summary>Snapshot builds are also available</summary>
+      
+```kotlin
+repositories {
+   maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
```
JakeWharton: Does this syntax work in Groovy?

StefMa: Jep, you're right.
This was only possible with the Kotlin DSL.
I've just updated the syntax which works for both Gradle DSLs. See https://github.com/square/okio/pull/664/commits/69866a72b2fb9291177422d8cb38f3bebd8e4b1a

swankjesse:
Thanks!

Issue #54, opened 2014-04-28
A seekable store


swankjesse:
We don't support anything like [SeekableByteChannel](http://docs.oracle.com/javase/7/docs/api/java/nio/channels/SeekableByteChannel.html). Would potentially be useful for random access. Could be very useful to navigate ZIP files, [dex files](https://android.googlesource.com/platform/libcore/+/master%5E2/dex/src/main/java/com/android/dex/Dex.java), etc.

Colin recommended a `Store` interface that permitted read and write access to a random access file.

rzimmer:
We needed random access as well for our [xar Library for Java](https://github.com/sprylab/xar/). It might not be the most elaborated implementation. but for an inspiration, you may want to have a look at [RandomAccessFileSource](https://github.com/sprylab/xar/blob/master/xar/src/main/java/com/sprylab/xar/utils/RandomAccessFileSource.java).

swankjesse:
Implemented in FileHandle
https://github.com/square/okio/issues/797

PR #680, opened 2019-11-15
[PoC] Initial ideas for an Okio Store

bnorm:
Looking at FileOperator from OkHttp and potentially my own need for random access file operations, I've put together some ideas for an Okio native Store.

Still a bunch to do like documentation and multiplatform support, but I thought this would be a good place to discuss. Do you still want this abstraction in Okio? What should the API require things like `size()` and `truncate()`?

Feel free to close this pull request if this is not something you want in Okio. Otherwise, let me know what changes you would like to see and I will work toward getting this pull request ready for merge.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Okio.kt
+++ b/okio/src/jvmMain/kotlin/okio/Okio.kt
@@ -26,17 +26,50 @@ import java.io.FileOutputStream
 import java.io.IOException
 import java.io.InputStream
 import java.io.OutputStream
+import java.io.RandomAccessFile
 import java.net.Socket
 import java.net.SocketTimeoutException
+import java.nio.channels.Channel
+import java.nio.channels.FileChannel
 import java.nio.file.Files
 import java.nio.file.OpenOption
 import java.nio.file.Path
+import java.nio.file.StandardOpenOption
 import java.util.logging.Level
 import java.util.logging.Logger
 
 actual fun Source.buffer(): BufferedSource = RealBufferedSource(this)
 
 actual fun Sink.buffer(): BufferedSink = RealBufferedSink(this)
+
+fun Store.source(pos: Long): Source {
+  return object : Source {
+    private var position = pos
+    override fun read(sink: Buffer, byteCount: Long): Long {
+      val read = read(position, sink, byteCount)
+      if (read == -1L) return -1L
+      position += read
+      return read
+    }
+
+    override fun timeout(): Timeout = timeout
+    override fun close() {}
+  }
+}
+
+fun Store.sink(pos: Long): Sink {
+  return object : Sink {
+    private var position = pos
+    override fun write(source: Buffer, byteCount: Long) {
+      write(position, source, byteCount)
+      position += byteCount
+    }
+
+    override fun flush() = this@sink.flush()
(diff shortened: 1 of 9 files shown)


okio/src/jvmMain/kotlin/okio/Okio.kt
```diff
@@ -108,6 +141,12 @@ private class InputStreamSource(
   override fun toString() = "source($input)"
 }
 
+fun FileChannel.store(): Store {
```
JakeWharton: Should be `asStore`.

"as" is for encapsulation, "to" is for transformation. We messed up with existing source and buffer functions.

bnorm: I agree. If a new convention is to be created - which it sounds like there is agreement - should the old convention be deprecated with a warning? I can create a separate pull-request for the existing functions.

bnorm:
I used the name Store as suggested by #54. Naming is hard and I tend not to be very good at it so happy to change to whatever is suggested.

I definitely like the idea of separating out a read and write interface like Source and Sink. I'll go with the names Input and Output respectively unless something else is suggested. Some additional ideas:

- It might be helpful to have a union interface for things like `FileChannel` that can act as both (like Store is currently) but maybe the responsibility should be on the caller to create an Input and Output if they are both wanted.
- Like `Buffer`, there could be a Packet type which is a union of Input and Output and uses `Segments` to read and write data.
- `ByteString` could implement Output.

swankjesse:
Suppose we assume SQLite is the best choice for its use case: mutable structured data stored in a file.

Is this true? I think it might be, and if so we should do things like make OkHttp’s overdue cookie storage depend on it.

If it's true what use cases are up for grabs here?

swankjesse:
I hate files and filesystems. 

Every interaction with files is difficult:

 - It’s hard or impossible to do anything atomically. Partial success is the worst.
 - Nothing is portable. Filesystems have inconsistent semantics between Windows and UNIX. On Windows you can’t delete a file if somebody is reading it!
 - Permissions exist and must be managed.
 - Is it a local SSD? Or a spinning floppy disk on an NFS mount? No natural way to tell, but the consequences are real.
 - You can fill up the disk.

I think there’s only a few sane ways to do files:
 - In managed cloud databases like S3, DynamoDB, and TiDB.
 - In local sqlite databases
 - Doing a streaming write of `file.tmp` from beginning to end, atomically renaming `file.tmp` to `file` upon success
 - Reading any local file that is immutable

I genuinely think your `Store` API is great! It’s simple and exactly what I’d reach for if I needed to do random-access edits of a file. But I don’t think most developers should attempt to do random-access edits of files. It’s too difficult to get right, and there are too many failure modes.

Would you consider hosting this in a separate project? I don’t think this code is bad; I just don’t want to support it!

bnorm:
As someone who had a job testing a NAS, I concur; file systems are very difficult.

I've got an [okiox](https://github.com/bnorm/okiox) project I started a long time ago which I've started to pick back up again recently to experiment with a coroutines based API. Maybe I'll add a `okiox-store` module to it and actually release it.

swankjesse:
Awesome, yep that sounds good.

PR #686, opened 2019-12-11
Remove empty tail Segment

### LINKED ISSUE #685: throw IOException when using okio.Buffer.inputStream() with InputStreamReader
Class **InputStreamReader** does not allow the read method of the underlying **InputStream** to return 0. If the content size of the incoming buffer of **okio.Buffer** is just a multiple of 8192 bytes, the penultimate read call (from okio.Buffer.inputStream()) will return 0.

Reproducible code:

```java
        byte[] target = new byte[8192];  // or 8192 * N
        Buffer buf = new Buffer();
        buf.readFrom(new ByteArrayInputStream(target));
        try (InputStream inputStream = buf.inputStream()) {
            InputStreamReader reader = new InputStreamReader(inputStream);
            char[] buf2 = new char[512];
            int n;
            while ((n = reader.read(buf2)) != -1) {
                // ...
            }
            reader.close();
        } catch (IOException e) {
            e.printStackTrace();       // always here
        }
```

Error stack trace:

```
java.io.IOException: Underlying input stream returned zero bytes
	at sun.nio.cs.StreamDecoder.readBytes(StreamDecoder.java:288)
	at sun.nio.cs.StreamDecoder.implRead(StreamDecoder.java:326)
	at sun.nio.cs.StreamDecoder.read(StreamDecoder.java:178)
	at java.io.InputStreamReader.read(InputStreamReader.java:184)
	at java.io.Reader.read(Reader.java:140)
        ...
```

Reproducible version：all (**2.4.1**, **1.17.4** tried)

cketti:
Fixes #685

### DIFF
--- a/okio/src/jvmTest/kotlin/okio/TestUtil.kt
+++ b/okio/src/jvmTest/kotlin/okio/TestUtil.kt
@@ -23,6 +23,7 @@ import java.io.Serializable
 import java.util.Random
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
+import kotlin.test.assertNotEquals
 import kotlin.test.assertTrue
 
 object TestUtil {
@@ -41,6 +42,13 @@ object TestUtil {
       }
     }
     return result
+  }
+
+  @JvmStatic
+  fun assertNoEmptySegment(buffer: Buffer) {
+    // This assertion uses the implementation detail that Buffer.read(ByteArray) returns 0 if there's an empty Segment.
+    assertNotEquals(0, buffer.read(byteArrayOf(0)),
+      "Expected Buffer to not contain an empty Segment; but Buffer.read(ByteArray) returned 0")
   }
 
   @JvmStatic
--- a/okio/src/jvmTest/kotlin/okio/TestUtil.kt
+++ b/okio/src/jvmTest/kotlin/okio/TestUtil.kt
@@ -23,7 +23,6 @@ import java.io.Serializable
 import java.util.Random
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
-import kotlin.test.assertNotEquals
 import kotlin.test.assertTrue
 
 object TestUtil {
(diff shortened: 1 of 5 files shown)


cketti:
> LGTM. I'm not so sure about the test approach. Can we reach inside and check internal state?

We could copy https://github.com/square/okio/blob/34e5eb0b3f4fdface44ca0c4aeabfafcab8c5bbc/okio/src/commonTest/kotlin/okio/util.kt#L25-L35 to `okio/src/jvmTest/kotlin/okio/TestUtil.kt` and use that.

okio/src/jvmTest/kotlin/okio/TestUtil.kt
```diff
@@ -33,14 +33,11 @@ object TestUtil {
...
-      while (cursor.next() > 0) {
-        result.add(cursor.end - cursor.start)
-      }
-    }
-    return result
+  fun segmentSizes(buffer: Buffer): List<Int> = okio.segmentSizes(buffer)
+
+  @JvmStatic
+  fun assertNoEmptySegments(buffer: Buffer) {
+    assertTrue(segmentSizes(buffer).all { it != 0 }, "Expected all segments to be non-empty")
```
swankjesse: I like this better.

swankjesse:
Released! https://github.com/square/okio/blob/master/CHANGELOG.md#version-242

bnorm:
As a follow on, as this pattern is used 4 times within Okio now, it might be nice to have a higher order function which handles this kind of unknown size segment write pattern. Something like the following might work nicely:

```kotlin
internal fun Buffer.readIntoSegment(block: (tail: Segment) -> Int): Int {
  val tail = writableSegment(1)
  val bytesRead = block(tail)
  if (bytesRead <= 0) {
    if (tail.pos == tail.limit) {
      // We allocated a tail segment, but didn't end up needing it. Recycle!
      head = tail.pop()
      SegmentPool.recycle(tail)
    }
  } else {
    tail.limit += bytesRead
    size += bytesRead.toLong()
  }
  return bytesRead
}
```

swankjesse:
Awesome idea. Taking it further: is there a way to use fancy Kotlin functions to make segment management foolishly easy? I'm imagining something where the writableSegment method itself is invisible it automatic.

PR #693, opened 2020-01-13
Add missing @Override to Watchdog.run() method.

edenrox:


JakeWharton:
Yikes. Error-prone should have caught this. Maybe we don't have the check turned on...

edenrox:
I think the Travis build is mis-configured.

I don't think jdk8 is supported on xenial:
https://travis-ci.community/t/expected-feature-release-number-in-range-of-9-to-12-but-got-8-installing-oraclejdk8/1345/8

edenrox:
Ok Travis checks are passing after rebase.  Should be good to merge.

PR #695, opened 2020-01-28
Migrate bug bounty URL to bugcrowd

swankjesse:


JakeWharton:
You should have them set up a redirect from `square.com/bugbounty` or something so we never have to do this again across 50 projects.

PR #697, opened 2020-02-14
Throw new InterruptedIOException subtype on timeout

dmgd:
Introduces TimeoutException which a caller can catch if it just wants to just deal with timeout and deadline reached, but not interrupts.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -16,7 +16,6 @@
 package okio
 
 import java.io.IOException
-import java.io.InterruptedIOException
 import java.util.concurrent.TimeUnit
 
 /**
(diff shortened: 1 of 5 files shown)


dmgd:
The specific case I'm dealing with is something that wants to transform timeouts into an http response, but leave interrupts alone.  It _could_ do this by catching InterruptedIOException and then working out whether it's a timeout or not + rethrowing if it's not, but this change would make the handling code _much_ cleaner

swankjesse:
Unfortunately you’re competing with `SocketTimeoutException` which also attempts to signal whether the failure was caused by timeout.

Introducing our own timeout exception feels like a complicated feature that we’d need to do everywhere, including significant tests in Okio and OkHttp.

dmgd:
> Unfortunately you’re competing with `SocketTimeoutException` which also attempts to signal whether the failure was caused by timeout.
> 
> Introducing our own timeout exception feels like a complicated feature that we’d need to do everywhere, including significant tests in Okio and OkHttp.

Good point about `SocketTimeoutException`.  We already catch and convert that, so reusing that would work perfectly for us.  I didn't reuse it just because these timeouts are not really socket related, so it felt a bit wrong . . .

In terms of backwards compatibility, `TimeoutException` is still an `InterruptedIOException` and has the same message it already had, so the only thing that I can think of that could break would be comparisons using `getClass()` rather than `instanceof`

swankjesse:
I definitely like subtyping `InterruptedIOException`. The holdup is more about finding all of hte places where we should be throwing `TimeoutException`.

dmgd:
So I think what I've just pushed might be everywhere that's relevant in okio (based on searches for `"timeout"` `"deadline reached"` and `InterruptedIOException`.)  Is it the sort of thing you had in mind?

I'm taking a look at OkHttp now :)

dmgd:
https://github.com/dmgd/okhttp/commit/71fa6b5e7d625fd6258a1b0c963f784aec42ac5b
and
https://github.com/dmgd/okhttp/commit/a9fb839811e615c81ad3b27949876c3cad71f8a4
look like the okhttp changes that would go with this

PR #698, opened 2020-02-16
Add NSData.toByteString()

Egorand:


okio/build.gradle
```diff
@@ -66,13 +66,25 @@ kotlin {
     nativeTest {
       dependsOn commonTest
     }
+    appleMain {
```
Egorand: Not sure if "apple" is the right term for iOS + macOS

okio/src/appleMain/kotlin/okio/ByteString.kt
```diff
@@ -0,0 +1,30 @@
...
+package okio
+
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSData
+import platform.posix.memcpy
+
+fun NSData.toByteString(): ByteString {
+  val data = this
+  return ByteString(ByteArray(data.length.toInt()).apply {
```
JakeWharton: Why is `toInt()` needed? Is it unsigned normally?

Egorand: Yeah,
>  inferred type is NSUInteger /* = ULong */

okio/src/appleMain/kotlin/okio/ByteString.kt
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
```
swankjesse: NEAT

okio/src/appleTest/kotlin/okio/AppleByteStringTest.kt
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import platform.Foundation.NSData
```
swankjesse: COOOL

JakeWharton:
Darwin?

swankjesse:
Darwin’s mostly a coincidence here. The AppKit / UiKit / NextStep APIs could target Linux or Windows, and I suspect there’s code in iTunes and Safari for Windows that does so.

swankjesse:
oooooh http://cocotron.org/

JakeWharton:
You could use that argument against Linux because it runs on Windows or Windows because it runs on Linux.

Egorand:
Are we likely to use any types outside of `Foundation`? Would `foundation` make sense then?

swankjesse:
I think Apple is the best name.

PR #699, opened 2020-02-16
Add Buffer extensions to read and write NSData

Egorand:
TODO: more tests

### DIFF
--- a/okio/src/appleMain/kotlin/okio/Buffer.kt
+++ b/okio/src/appleMain/kotlin/okio/Buffer.kt
@@ -24,6 +24,7 @@ import platform.Foundation.NSData
 import platform.Foundation.NSMutableData
 import platform.Foundation.create
 import platform.darwin.NSInteger
+import platform.darwin.NSUInteger
 import platform.posix.memcpy
 
 fun Buffer.write(
@@ -54,13 +55,12 @@ fun Buffer.write(
   size += byteCount.toLong()
 }
 
-fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
-  require(byteCount >= 0 && byteCount <= Int.MAX_VALUE) { "byteCount: $byteCount" }
-  if (size < byteCount) throw EOFException()
+fun Buffer.readNSData(byteCount: NSUInteger = size.toULong()): NSData {
+  require(byteCount >= 0u && byteCount <= NSUInteger.MAX_VALUE) { "byteCount: $byteCount" }
+  if (size.toULong() < byteCount) throw EOFException()
 
-  val length = byteCount.toULong()
-  val result = NSMutableData.create(length = length)
-    ?: throw IOException("Failed to create NSMutableData of length $length")
+  val result = NSMutableData.create(length = byteCount)
+    ?: throw IOException("Failed to create NSMutableData of length $byteCount")
   readFully(result)
   return result
 }
--- a/okio/src/appleMain/kotlin/okio/Buffer.kt
+++ b/okio/src/appleMain/kotlin/okio/Buffer.kt
@@ -22,7 +22,6 @@ import kotlinx.cinterop.addressOf
 import kotlinx.cinterop.usePinned
 import platform.Foundation.NSData
 import platform.Foundation.NSMutableData
-import platform.Foundation.appendBytes
 import platform.Foundation.create
 import platform.darwin.NSInteger
 import platform.posix.memcpy
--- a/okio/src/appleMain/kotlin/okio/Buffer.kt
+++ b/okio/src/appleMain/kotlin/okio/Buffer.kt
@@ -1,0 +1,101 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import kotlinx.cinterop.ByteVarOf
+import kotlinx.cinterop.CPointer
+import kotlinx.cinterop.plus
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSData
+import platform.Foundation.NSMutableData
+import platform.Foundation.appendBytes
+import platform.Foundation.create
+import platform.darwin.NSInteger
+import platform.posix.memcpy
+
+fun Buffer.write(
+  source: NSData,
+  offset: Int = 0,
+  byteCount: Int = source.length.toInt()
+): Buffer = apply {
+  var offset = offset
+  checkOffsetAndCount(source.length.toLong(), offset.toLong(), byteCount.toLong())
+
+  val limit = offset + byteCount
+  while (offset < limit) {
+    val tail = writableSegment(1)
+
+    val toCopy = minOf(limit - offset, Segment.SIZE - tail.limit)
+    tail.data.usePinned { pinned ->
+      memcpy(
+        __dst = pinned.addressOf(tail.limit),
+        __src = (source.bytes as CPointer<ByteVarOf<*>>) + offset,
+        __n = toCopy.toULong()
+      )
+    }
+
+    offset += toCopy
+    tail.limit += toCopy
+  }
+
+  size += byteCount.toLong()
+}
+
+fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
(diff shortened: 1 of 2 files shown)


okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+import kotlinx.cinterop.plus
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSData
+import platform.Foundation.NSMutableData
+import platform.Foundation.create
+import platform.darwin.NSInteger
+import platform.posix.memcpy
+
+fun Buffer.write(
```
JakeWharton: What does this look like from Obj-C or Swift?

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+  size += byteCount.toLong()
+}
+
+fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
+  require(byteCount >= 0 && byteCount <= Int.MAX_VALUE) { "byteCount: $byteCount" }
+  if (size < byteCount) throw EOFException()
+
+  val length = byteCount.toULong()
+  val result = NSMutableData.create(length = length)
+    ?: throw IOException("Failed to create NSMutableData of length $length")
```
JakeWharton: Should this be an OOM? When would this fail?

swankjesse: Yep, should be OOM. (Is there a multiplatform OOM type?)

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  size += byteCount.toLong()
+}
+
+fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
+  require(byteCount >= 0 && byteCount <= Int.MAX_VALUE) { "byteCount: $byteCount" }
+  if (size < byteCount) throw EOFException()
+
+  val length = byteCount.toULong()
```
swankjesse: should the parameter be a ULong, if that’s NSData’s length type?

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    offset += toCopy
+    tail.limit += toCopy
+  }
+
+  size += byteCount.toLong()
+}
+
+fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
+  require(byteCount >= 0 && byteCount <= Int.MAX_VALUE) { "byteCount: $byteCount" }
```
swankjesse: I don’t think this constraint holds? We could read more than Int.MAX_INT if NSData will take it.

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+  if (size < byteCount) throw EOFException()
+
+  val length = byteCount.toULong()
+  val result = NSMutableData.create(length = length)
+    ?: throw IOException("Failed to create NSMutableData of length $length")
+  readFully(result)
+  return result
+}
+
+fun Buffer.read(sink: NSMutableData, offset: Int = 0, byteCount: Int = size.toInt()): Int {
```
swankjesse: byteCount should be a long?

swankjesse: result should be a long?

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    offset += toCopy
+    tail.limit += toCopy
+  }
+
+  size += byteCount.toLong()
+}
+
+fun Buffer.readNSData(byteCount: NSInteger = size): NSData {
```
swankjesse: should the parameter types be NSInteger or Int or Long?

okio/src/appleMain/kotlin/okio/Buffer.kt
swankjesse: should this method loop? Why read fewer bytes than byteCount ?

swankjesse: ... doing less-than-max reads is more useful for sources that might need to block for sufficient data. In this case we should only be limited by the number of bytes in this buffer, not in any individual segment.

okio/src/appleMain/kotlin/okio/Buffer.kt
JakeWharton: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-out-of-memory-error/index.html

Egorand: `NSMutableData.create(length)` returns null when [`malloc` fails internally](http://www.cplusplus.com/reference/cstdlib/malloc/):

> If the function failed to allocate the requested block of memory, a null pointer is returned.

It's unclear whether it's always be caused by OOM, probably not.

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: Good call regarding types. I copied method signatures from `commonMain/Buffer`, but it does make sense to use types that `NSData` uses.

In general, Foundation numeric types are typealiased to Kotlin types, e.g.:

```kotlin
package platform.darwin

public typealias NSInteger = kotlin.Long
public typealias NSUInteger = kotlin.ULong
```

So when this code is called from Kotlin there should be no difference in whether `NSInteger` or `Long` was in the method signature. However, using darwin types should make Swift/ObjC call sites more ergonomic.

Egorand: afaff8f

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: Changed to `NSUInteger.MAX_VALUE`: afaff8f

okio/src/appleMain/kotlin/okio/Buffer.kt
JakeWharton: That sounds like OOM conditions to me.

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: a2877ce

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: afaff8f

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: Yeah, this is just a copy of [`Buffer.commonRead()`](https://github.com/square/okio/blob/cedf01bbf7e7144105db653f247eda0dbe4a15ae/okio/src/commonMain/kotlin/okio/internal/Buffer.kt#L612) adapted to work with `NSMutableData`. Let's review the implementation and fix in a followup?

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand: 2b457f4

okio/src/appleMain/kotlin/okio/Buffer.kt
Egorand:
```swift
import Foundation
import Okio

let hello = "Hello".data(using: .utf8)!

let buffer = Buffer()
buffer.write(source: hello, offset: 0, byteCount_: Int32(hello.count))

let data = buffer.readNSData(byteCount: UInt64(hello.count))
print(String.init(data: data, encoding: .utf8)!)
```

Egorand: 1. Weird that `byteCount` required different conversions for write and read, will look into changing signatures to not require any conversions.
2. Should probably drop `NS` to follow Swift conventions, maybe even switch to just `read()`.

okio/src/appleMain/kotlin/okio/Buffer.kt
```diff
@@ -0,0 +1,112 @@
...
+
+  val limit = offset + byteCount
+  while (offset < limit) {
+    val tail = writableSegment(1)
+
+    val toCopy = minOf(limit - offset, Segment.SIZE - tail.limit)
+    tail.data.usePinned { pinned ->
+      memcpy(
+        __dst = pinned.addressOf(tail.limit),
+        __src = (source.bytes as CPointer<ByteVarOf<*>>) + offset,
```
varahash, line 47: Also, it worth notice, that [`NSData.enumerateBytes(_:)`](https://developer.apple.com/documentation/foundation/nsdata/1408400-enumeratebytes) should be used instead of `NSData.bytes` in order to make it more memory efficient. Since the NSData can be backed by multiple not-contiguous buffers under-the-hood, and when you access such NSData using `bytes` property, all buffers will be flattened into one (creates new single contiguous buffer).

Egorand: Good call! Want to submit a PR?

PR #707, opened 2020-03-10
Kotlin 1.3.70

Egorand:
Changelog: https://github.com/JetBrains/kotlin/blob/1.3.70/ChangeLog.md

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 buildscript {
   ext.versions = [
-      'kotlin': '1.3.61',
+      'kotlin': '1.3.70',
       'jmhPlugin': '0.4.8',
       'animalSnifferPlugin': '1.5.0',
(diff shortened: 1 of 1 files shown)


martinbonnin:
Am I correct that if I want to use okio with kotlin 1.3.70 I'd need a new release of okio > 2.4.3 ? 

I currently get the following error:

``` 
The abi versions don't match. Expected '[22]', found '17'
```

JakeWharton:
Yes

On Thu, Mar 19, 2020, at 6:39 PM, Martin Bonnin wrote:
> 

> Am I correct that if I want to use okio with kotlin 1.3.70 I'd need a new release of okio > 2.4.3 ?

> I currently get the following error:

> `The abi versions don't match. Expected '[22]', found '17'
`

Issue #260, opened 2016-09-07
CipherSink/CipherSource


cketti:
What are your thoughts on CipherSink and CipherSource analogous to CipherInputStream/CipherOutputStream? Is that something that would be included in Okio or should it go into a separate library?

swankjesse:
The cipher streams are pretty hazardous to use directly – it’s so easy to do the IV or MAC wrong and write broken crypto. I’d prefer to recommend specific high-level abstractions like TLS. Is there a good format for encrypting data at rest? GPG?

cketti:
Fair enough. It's probably a good idea not to include foot guns.

TLS and OpenPGP both come with enough parameters that using them isn't as risk-free as one would hope. Still, better than rolling your own cryptosystem.

However, I'll soon need CipherSource/CipherSink to create an implementation of a proprietary crypto system. It'll be a separate library with a big fat warning message then.

erickok:
For future visitors that are interested, I have a CiperSource/CipherSink implementation (as well as Base64Source/Sink) on https://github.com/erickok/okio-extensions which, without warrety, might help you out.

PR #711, opened 2020-05-02
CipherSink & CipherSource

MartinDevi:
I know #260 was closed, but I don't really understand the issue, given that the initialization of the cipher is outside of the responsibility of the stream implementation. Let me know if I this is a fool's errand, but I figure this was at least worth a shot. This would be very useful for applications which treat large amounts of ciphered data, because the usage of internal segments would be a big improvement for memory usage, not to mention the simple fact of having cryptography properly included in the okio ecosystem for piping.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -106,7 +106,7 @@ private class CipherSink internal constructor(
     var thrown: Throwable? = null
     val buffer = sink.buffer
 
-    // For block cipher, output size cannot block size in doFinal
+    // For block cipher, output size cannot exceed block size in doFinal
     val s = buffer.writableSegment(outputSize)
 
     try {
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -1,0 +1,93 @@
+package okio
+
+import java.io.IOException
+import javax.crypto.Cipher
+
+class CipherSink internal constructor(private val sink: BufferedSink, private val cipher: Cipher) : Sink {
+  constructor(sink: Sink, cipher: Cipher) : this(sink.buffer(), cipher)
+
+  private var closed = false
+
+  @Throws(IOException::class)
+  override fun write(source: Buffer, byteCount: Long) {
+    checkOffsetAndCount(source.size, 0, byteCount)
+    check(!closed) { "closed" }
+
+    var remaining = byteCount
+    while (remaining > 0) {
+      val head = source.head!!
+      var toCipher = minOf(remaining, head.limit - head.pos).toInt()
+      var outputSize = cipher.getOutputSize(toCipher)
+      // TODO: Consider using fixed block size (imposed and verified in constructor)
+      while (outputSize > Segment.SIZE) {
+        toCipher /= 2
+        outputSize = cipher.getOutputSize(toCipher)
+      }
+
+      val buffer = sink.buffer
+      val s = buffer.writableSegment(outputSize)
+      val ciphered = cipher.update(head.data, head.pos, toCipher, s.data, s.limit)
+      s.limit += ciphered
+      buffer.size += ciphered
+
+      if (s.pos == s.limit) {
+        buffer.head = s.pop()
+        SegmentPool.recycle(s)
+      }
+
+      source.size -= toCipher
+      head.pos += toCipher
+      if (head.pos == head.limit) {
+        source.head = head.pop()
+        SegmentPool.recycle(head)
+      }
+
+      remaining -= toCipher
+    }
+  }
+
+  override fun flush() {}
+
+  override fun timeout(): Timeout =
+    sink.timeout()
+
+  @Throws(IOException::class)
+  override fun close() {
+    if (closed) return
+
+    var thrown: Throwable? = null
+
+    val outputSize = cipher.getOutputSize(0)
+    if (outputSize != 0) {
+      val buffer = sink.buffer
+      val s = buffer.writableSegment(outputSize)
+
+      try {
+        val ciphered = cipher.doFinal(s.data, s.limit)
+
+        s.limit += ciphered
+        buffer.size += ciphered
+      } catch (e: Throwable) {
+        thrown = e
+      }
+
+      if (s.pos == s.limit) {
+        buffer.head = s.pop()
+        SegmentPool.recycle(s)
+      }
+    }
+
+    try {
+      sink.close()
+    } catch (e: Throwable) {
+      if (thrown == null) thrown = e
+    }
+
+    closed = true
+
+    if (thrown != null) throw thrown
+  }
+}
+
+fun Sink.cipherSink(cipher: Cipher): CipherSink =
+  CipherSink(this, cipher)
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -22,6 +22,9 @@ package okio
 import java.io.IOException
 import javax.crypto.Cipher
 
+/**
+ * A sources that uses a [Cipher] to process data read from another source.
+ */
 class CipherSink internal constructor(private val sink: BufferedSink, private val cipher: Cipher) : Sink {
   constructor(sink: Sink, cipher: Cipher) : this(sink.buffer(), cipher)
 
@@ -124,5 +127,10 @@ class CipherSink internal constructor(private val sink: BufferedSink, private va
   }
 }
 
+/**
+ * Returns a [CipherSink] that processes data to this [Sink] using [cipher] while writing.
+ *
+ * @see CipherSink
+ */
 inline fun Sink.cipherSink(cipher: Cipher): CipherSink =
   CipherSink(this, cipher)
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -22,7 +22,7 @@ package okio
 import java.io.IOException
 import javax.crypto.Cipher
 
-private class CipherSink internal constructor(
+private class CipherSink(
   private val sink: BufferedSink,
   private val cipher: Cipher
 ) : Sink {
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -130,6 +130,9 @@ private class CipherSink(
 /**
  * Returns a [Sink] that processes data using this [Cipher] while writing to
  * [sink].
+ *
+ * @throws IllegalArgumentException
+ *  If this isn't a block cipher.
  */
 fun Cipher.sink(sink: Sink): Sink =
   CipherSink(sink, this)
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -27,6 +27,7 @@ class CipherSink internal constructor(private val sink: BufferedSink, private va
       val buffer = sink.buffer
       val s = buffer.writableSegment(outputSize)
       val ciphered = cipher.update(head.data, head.pos, toCipher, s.data, s.limit)
+
       s.limit += ciphered
       buffer.size += ciphered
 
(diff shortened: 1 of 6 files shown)


swankjesse:
I like the design of leaving cipher set up out of scope. Good idea!

swankjesse:
Can you make CipherSource and CipherSink non-public types? I'd like to have this API from Kotlin:

```
val cipher: Cipher = ...
val cipherSink: Sink = cipher.sink(sink) // returns a Sink
val cipherSource: Sourfe = cipher.source(souce) // returns a Source
```

For Java it'd be exposed as static methods on Okio.

MartinDevi:
Done. 
- `CipherSource and CipherSink` are private classes. Changed the signature of the methods, to create instances, which means they are no longer inline.
- Java static methods in the `Okio` class. 
```kotlin
@file:JvmMultifileClass
@file:JvmName("Okio")
```

erickok:
I did some testing (on cases which were not fully covered by the well-done written tests) and this looks really good to me. 

Only thing I noticed is that requesting X bytes from the source will potentially read more from _its_ source than would stricktly be required (given the cipher block size) but that seems to be library-standard, using the segments and their size instead.

MartinDevi:
👋 Any news on this? Any chance this could make it into an upcoming release? Thanks.

MartinDevi:
@swankjesse @Egorand Is this something that I'm likely to ever see merged to be included natively within the library?

swankjesse:
Apologies for radio silence here. I've been busy. Will try to get to this soon, but it's a lot of code and I'm still unsure about the approach. In particular I like to support application-ready standards like TLS or SHA-256. In this case it’s just a building block.

MartinDevi:
Thanks for your response. Supporting a standard like TLS out of the box does sound great, although that's just one application for this type of block cipher stream. In my case, the application for this is to handle encryption of user data protected by a password within an app, so there's really no standard layer to add on top of the cipher.

swankjesse:
Can you explain more about your use case? Doing anything fancy to go from password to a symmetric key? I think I wish there was a format like gpg that was good.

MartinDevi:
Actually to be more exact the user data generally is ciphered using a randomly generated key, and it's that key which itself is ciphered by the derived password (this allows the key to be reused, so that key derivation is only performed once, since that's the most costly operation). Key derivation of the password is done most often using [Argon2](https://en.wikipedia.org/wiki/Argon2) in most cases, but [PBKDF2](https://en.wikipedia.org/wiki/PBKDF2) is also supported. Let me know if you're curious for more details. 

This is all pretty standard, but the actual details are essentially different in each implementation. Unfortunately there's no universal norm.

MartinDevi:
I tried adding a few examples in the README, to make things a bit clearer.

I also did some thinking about what could potentially be provided to make this easier to use. The only thing I could think of would be some sort of builder for the cipher itself.

```kotlin
class CipherBuilder {

  private var encrypt: Boolean? = null
  private var transformation: String? = null
  private var mode: String? = null
  private var iv: ByteArray? = null
  private var padding: String? = null
  private var key: ByteArray? = null

  fun encrypt() = apply { encrypt = true }
  fun decrypt() = apply { encrypt = false }

  fun aes() = apply { transformation = "AES" }
  fun des() = apply { transformation = "DES" }
  fun tripleDes() = apply { transformation = "DESede" }

  fun cbc(iv: ByteArray) = apply { mode = "CBC"; this.iv = iv }
  fun ecb() = apply { mode = "ECB" }

  fun pkcs5Padding() = apply { mode = "PKCS5Padding" }
  fun noPadding() = apply { mode = "des" }

  fun key(key: ByteArray) = apply { this.key = key }

  fun build(): Cipher {
    val encrypt = checkNotNull(encrypt) { "encrypt == null" }
    val algorithm = checkNotNull(transformation) { "transformation == null" }
    val mode = checkNotNull(mode) { "mode == null" }
    val padding = checkNotNull(padding) { "padding == null" }
    val key = checkNotNull(key) { "key == null" }

    val opMode = if (encrypt) Cipher.ENCRYPT_MODE else Cipher.DECRYPT_MODE
    val secretKeySpec = SecretKeySpec(key, algorithm)
    return Cipher.getInstance("$algorithm/$mode/$padding").apply {
      when (mode) {
        "CBC" -> init(opMode, secretKeySpec, IvParameterSpec(checkNotNull(iv) { "iv == null" }))
        "ECB" -> init(opMode, secretKeySpec)
        else -> error("Unknown mode $mode")
      }
    }
  }
}
```

With this builder, this is what the example that I added in the README would look like.

```kotlin
fun encryptAes(bytes: ByteString, file: File, key: ByteArray, iv: ByteArray) {
  val cipher = CipherBuilder()
    .encrypt()
    .aes()
    .cbc(iv)
    .pkcs5Padding()
    .key(key)
    .build()
  cipher.sink(file.sink()).buffer().use { it.write(bytes) }
}
```

I'm not sure this adds a whole lot TBH, and it could be argued that this level of specification for crypto falls outside the scope of this library. The `javax.crypto.Cipher` class is designed in a very extensible way, to support additional algorithms that may be available at runtime, and I don't see a way of providing a primitive without limiting that, so I don't really like this idea of adding a wrapper.

I also imagine that a possible concern is the possibility to eventually extend this to multiplatform (which I agree is a very good concern to have, and something that I could very well be interested in). Currently the part of the API that we use from the `Cipher` class looks like this.

```kotlin
interface Cipher {
  val blockSize: Int
  fun getOutputSize(inputLen: Int): Int
  fun update(input: ByteArray, inputOffset: Int, inputLen: Int, output: ByteArray, outputOffset: Int): Int
  fun doFinal(output: ByteArray, outputOffset: Int): Int
} 
```

This seems like a reasonable contract for any cipher implementation, so a similar JS class could be created using `expect/actual`. That being said, I'm don't really have experience with Kotlin multiplatform, and I'm not very familiar with the crypto implementations available in JS.

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
```
swankjesse, line 2: supernit: 2020

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
...
+package okio
+
+import java.io.IOException
+import javax.crypto.Cipher
+
+private class CipherSource(
+  private val source: BufferedSource,
+  private val cipher: Cipher
+) : Source {
+  constructor(source: Source, cipher: Cipher) : this(source.buffer(), cipher)
```
swankjesse, line 29: Same . . . just one constructor? The reason I don’t like the two is that `Source` and `BufferedSource` are subtypes of each other so it’s easy to accidentally call the wrong one

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    val ciphered =
+      cipher.update(head.data, head.pos, size, s.data, s.pos)
+
+    source.skip(size.toLong())
+
+    s.limit += ciphered
+    buffer.size += ciphered
+
+    if (s.pos == s.limit) {
```
swankjesse, line 81: Add this comment, 
```
          // We allocated a tail segment, but didn't end up needing it. Recycle!
```

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
...
+    val head = source.buffer.head!!
+    val size = head.limit - head.pos
+
+    // For block cipher, output size cannot exceed input size in update
+    val s = buffer.writableSegment(size)
+
+    val ciphered =
+      cipher.update(head.data, head.pos, size, s.data, s.pos)
+
+    source.skip(size.toLong())
```
swankjesse, line 76: nice

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
...
+        update()
+      }
+    }
+  }
+
+  private fun update() {
+    val head = source.buffer.head!!
+    val size = head.limit - head.pos
+
+    // For block cipher, output size cannot exceed input size in update
```
swankjesse, line 70: this comment is great

okio/src/jvmMain/kotlin/okio/CipherSource.kt
```diff
@@ -0,0 +1,123 @@
...
+
+    // For block cipher, output size cannot exceed block size in doFinal
+    val s = buffer.writableSegment(outputSize)
+
+    val ciphered = cipher.doFinal(s.data, s.pos)
+
+    s.limit += ciphered
+    buffer.size += ciphered
+
+    if (s.pos == s.limit) {
```
swankjesse, line 99: last time with this comment, 

```
          // We allocated a tail segment, but didn't end up needing it. Recycle!
```

okio/src/jvmTest/kotlin/okio/CipherFactory.kt
```diff
@@ -0,0 +1,69 @@
+package okio
+
+import javax.crypto.Cipher
+import javax.crypto.spec.IvParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.random.Random
+
+class CipherFactory(
```
swankjesse, line 8: Totally rad.

okio/src/jvmTest/kotlin/okio/CipherSinkTest.kt
```diff
@@ -0,0 +1,161 @@
...
+
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.junit.runners.Parameterized
+import kotlin.random.Random
+
+@RunWith(Parameterized::class)
+class CipherSinkTest(private val cipherAlgorithm: CipherAlgorithm) {
+  companion object {
+    @get:Parameterized.Parameters(name = "{0}")
```
swankjesse, line 11: fancy

swankjesse (approved):
I love it.

Apologies for sitting on this for so long. I’ve got a bunch of nitpicks. You can address ’em, but if you don’t I’ll do it and merge.

Thank you!

swankjesse:
Contributor’s agreement is complete.

okio/src/jvmMain/kotlin/okio/CipherSink.kt
```diff
@@ -0,0 +1,138 @@
...
+package okio
+
+import java.io.IOException
+import javax.crypto.Cipher
+
+private class CipherSink(
+  private val sink: BufferedSink,
+  private val cipher: Cipher
+) : Sink {
+  constructor(sink: Sink, cipher: Cipher) : this(sink.buffer(), cipher)
```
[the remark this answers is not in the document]
bnorm: Do we want this to be a public type? Should this follow a similar patter as GzipSource with this as the only public constructor?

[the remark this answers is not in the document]
swankjesse: It couldn’t really hurt, and it’ll give us flexibility if ever we need a method like Inflater’s readOrInflate().

MartinDevi:
Thanks for all your feedback, I applied the changes.

MartinDevi:
On the topic of the platform cipher API, I was wondering how difficult it would actually be to implement AES in Kotlin and ship it as part of the library, to have a directly multiplatform implementation. To answer that, I did a [POC](https://github.com/MartinDevi/okio/commit/c8a79eb08bddfba4176c0294d880619a017fa5c8) of by copying the BouncyCastle implementation in Java, which could easily be converted to Kotlin. It's tested for correctness against the JVM implementation. So this is definitely an option. The big loss of this approach is extensibility, in this POC only ECB and CBC modes are supported for AES, so anything else have to be added in order to be supported. That being said, I think it would be fairly easy to allow implementations of the block cipher interface which would map to the JVM cipher.

Looking at the possibilities for having a JS version of the stream cipher, it seems like in any case the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) doesn't support streaming, so a different implementation would have to be used.

swankjesse (approved):

swankjesse:
I’m reluctant to add our own set of cipher suites, just because that project has unbounded scope! But I think it’d be pretty rad to see a library that integrates with Okio types to do multiplatform encryption.

swankjesse:
Small follow-up to this PR here: https://github.com/square/okio/pull/780

MartinDevi:
Yeah, I totally follow that starting to support the cipher suites themselves could quickly go sideways and bloat the library with stuff which strays from its core purpose.

Happy to see this integrated into the library, thanks again!

PR #717, opened 2020-05-23
Make the segment pool lock-free on the JVM

swankjesse:
Under a high contention environment we'll do more zero-fill and GC
instead of locking.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
@@ -1,0 +1,104 @@
+/*
+ * Copyright (C) 2014 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import okio.SegmentPool.LOCK
+import okio.SegmentPool.MAX_SIZE
+import okio.SegmentPool.recycle
+import okio.SegmentPool.take
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * This class pools segments in a lock-free singly-linked stack. Though this code is lock-free it
+ * does use a sentinel [LOCK] value to defend against races.
+ *
+ * When popping, a caller swaps the stack's next pointer with the [LOCK] sentinel. If the stack was
+ * not already locked, the caller replaces the head node with its successor.
+ *
+ * When pushing, a caller swaps the head with a new node whose successor is the replaced head.
+ *
+ * If operations conflict, segments are not pushed into the stack. A [recycle] call that loses a
+ * race will not add to the pool, and a [take] call that loses a race will not take from the pool.
+ * Under significant contention this pool will have fewer hits and the VM will do more GC and zero
+ * filling of arrays.
+ *
+ * Note that the [MAX_SIZE] may be exceeded if multiple calls to [recycle] race. Exceeding the
+ * target pool size by a few segments doesn't harm performance, and imperfect enforcement is less
+ * code.
+ */
+internal actual object SegmentPool {
+  /** The maximum number of bytes to pool.  */
+  // TODO: Is 64 KiB a good maximum size? Do we ever have that many idle segments?
+  actual val MAX_SIZE = 64 * 1024L // 64 KiB.
+
+  /** A sentinel segment to indicate that the linked list is currently being modified. */
+  private val LOCK = Segment(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
+
+  /** Singly-linked list of segments. */
+  private var firstRef = AtomicReference<Segment?>()
+
+  /** Total bytes in this pool. */
+  private var atomicByteCount = AtomicLong()
+
+  actual val byteCount: Long
+    get() = atomicByteCount.get()
+
+  @JvmStatic
+  actual fun take(): Segment {
+    val first = firstRef.getAndSet(LOCK)
+    when {
+      first === LOCK -> {
+        // We didn't acquire the lock. Don't take a pooled segment.
+        return Segment()
+      }
+      first == null -> {
+        // We acquired the lock but the pool was empty. Unlock and return a new segment.
+        firstRef.set(null)
+        return Segment()
+      }
+      else -> {
+        // We acquired the lock and the pool was not empty. Pop the first element and return it.
+        firstRef.set(first.next)
+        first.next = null
+        atomicByteCount.addAndGet(-Segment.SIZE.toLong())
+        return first
+      }
+    }
+  }
+
+  @JvmStatic
+  actual fun recycle(segment: Segment) {
+    require(segment.next == null && segment.prev == null)
+    if (segment.shared) return // This segment cannot be recycled.
+    if (atomicByteCount.get() >= MAX_SIZE) return // Pool is full.
+
+    val first = firstRef.get()
+    if (first === LOCK) return // A take() is currently in progress.
+
+    segment.next = first
+    segment.limit = 0
+    segment.pos = 0
+
+    if (firstRef.compareAndSet(first, segment)) {
+      // We successfully recycled this segment. Adjust the pool size.
+      atomicByteCount.addAndGet(Segment.SIZE.toLong())
+    } else {
+      // We raced another operation. Don't recycle this segment.
+      segment.next = null
+    }
+  }
+}
--- a/okio/src/jsMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jsMain/kotlin/okio/SegmentPool.kt
@@ -15,19 +15,13 @@
  */
 package okio
 
-/**
- * A collection of unused segments, necessary to avoid GC churn and zero-fill.
- * This pool is a thread-safe static singleton.
- */
-internal expect object SegmentPool {
-  val MAX_SIZE: Long
+internal actual object SegmentPool {
+  actual val MAX_SIZE: Long = 0L
 
-  /** For testing only. Returns a snapshot of the number of bytes currently in the pool. */
-  val byteCount: Long
+  actual val byteCount: Long = 0L
 
-  /** Return a segment for the caller's use. */
-  fun take(): Segment
+  actual fun take(): Segment = Segment()
 
-  /** Recycle a segment that the caller no longer needs. */
-  fun recycle(segment: Segment)
+  actual fun recycle(segment: Segment) {
+  }
 }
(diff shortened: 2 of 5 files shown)


okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  /** Total bytes in this pool. */
+  private var atomicByteCount = AtomicLong()
+
+  actual val byteCount: Long
+    get() = atomicByteCount.get()
+
+  actual fun take(): Segment {
+    return when (val first = firstRef.getAndSet(LOCK)) {
+      LOCK -> {
```
Egorand, line 62: I wonder if `when` will do a ref check or an equality check here? Will this code break if it's an equality check?

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+        firstRef.set(first.next)
+        first.next = null
+        atomicByteCount.addAndGet(-Segment.SIZE.toLong())
+        first
+      }
+    }
+  }
+
+  actual fun recycle(segment: Segment) {
+    require(segment.next == null && segment.prev == null)
```
Egorand, line 82: nit: maybe add an error message? Or even split this into two requires?

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+    if (segment.shared) return // This segment cannot be recycled.
+    if (atomicByteCount.get() >= MAX_SIZE) return // Pool is full.
+
+    val first = firstRef.get()
+    if (first == LOCK) return // A take() is currently in progress.
```
Egorand, line 87: `first === LOCK`?

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+  /** A sentinel segment to indicate that the linked list is currently being modified. */
+  private val LOCK = Segment(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
+
+  /** Singly-linked list of segments. */
+  private var firstRef = AtomicReference<Segment>()
+
+  /** Total bytes in this pool. */
+  private var atomicByteCount = AtomicLong()
+
+  actual val byteCount: Long
```
JakeWharton, line 57: What is this used for? The order of operations means that you can observe this being larger than what's actually available.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
JakeWharton, line 87: `===`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
JakeWharton, line 62: This will do `==` but we probably want `===`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+ * filling of arrays.
+ *
+ * Note that the [MAX_SIZE] may be exceeded if multiple calls to [recycle] race. Exceeding the
+ * target pool size by a few segments doesn't harm performance, and imperfect enforcement is less
+ * code.
+ */
+internal actual object SegmentPool {
+  /** The maximum number of bytes to pool.  */
+  // TODO: Is 64 KiB a good maximum size? Do we ever have that many idle segments?
+  actual val MAX_SIZE = 64 * 1024L // 64 KiB.
```
JakeWharton, line 46: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  /** A sentinel segment to indicate that the linked list is currently being modified. */
+  private val LOCK = Segment(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
```
JakeWharton, line 49: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  /** Singly-linked list of segments. */
+  private var firstRef = AtomicReference<Segment>()
```
JakeWharton, line 52: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  /** Total bytes in this pool. */
+  private var atomicByteCount = AtomicLong()
```
JakeWharton, line 55: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
JakeWharton, line 57: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+
+  actual val byteCount: Long
+    get() = atomicByteCount.get()
+
+  actual fun take(): Segment {
```
JakeWharton, line 60: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,101 @@
...
+        // We acquired the lock and the pool was not empty. Pop the first element and return it.
+        firstRef.set(first.next)
+        first.next = null
+        atomicByteCount.addAndGet(-Segment.SIZE.toLong())
+        first
+      }
+    }
+  }
+
+  actual fun recycle(segment: Segment) {
```
JakeWharton, line 81: `@JvmStatic`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: I like your idea of using `===`. Lemme switch to that. (They’ll do the same thing regardless, because Segment doesn’t override equals()).

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: It’s more of an assert than anything. This isn’t public API. I’d prefer to not split it up because its primary purpose is documentation.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: Good call. Fixed.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: Just testing.

swankjesse: Lemme document that.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: Fixed.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: Yep, good call!

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: Since it’s an `object` I get that for free.

swankjesse: Actually, only the fields, not the methods!

swankjesse: Fixed all to be `@JvmStatic`. Can’t hurt.

swankjesse:
Agreed 100% on wanting to give this lots of exercise before shipping it to everyone!

swankjesse:
@Egorand I can’t do `commonMain` ’cause JVM/AtomicReference and Kotlin/AtomicReference don’t have an `expect` definition in common. When I tried to hack one myself, KMP got angry at the type parameters on `AtomicReference`.

swankjesse:
This branch released as `2.7.0-alpha.lockfree.1`.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
Egorand: Why care about `@JvmStatic` on private fields though? To avoid calling getters internally? Won't it be `@JvmField` then?

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,109 @@
...
+  @JvmStatic
+  actual val MAX_SIZE = 64 * 1024L // 64 KiB.
+
+  /** A sentinel segment to indicate that the linked list is currently being modified. */
+  @JvmStatic
+  private val LOCK = Segment(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
+
+  /** Singly-linked list of segments. */
+  @JvmStatic
+  private var firstRef = AtomicReference<Segment>()
```
bnorm, line 55: `AtomicReference<Segment?>` will get you some null type safety. Otherwise `firstRef.getAndSet(LOCK)` and `firstRef.get()` will return platform types.

swankjesse: Oooh I like it.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: I decided to go with `@JvmStatic` on the methods and not on the fields. That’s the least amount of code and the simplest dispatch for callers.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: (done)

swankjesse:
AtomicFU looks brilliant. That’d be an excellent tool for our toolchain if/when we get serious about OkHttp in Kotlin/Native. For now I’m reluctant only because it’s experimental magic, and Okio is no place for experimental magic.

adriancole:
added some notes after reviewing this incredibly late https://github.com/square/okio/pull/831

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -0,0 +1,104 @@
...
+  @JvmStatic
+  actual fun recycle(segment: Segment) {
+    require(segment.next == null && segment.prev == null)
+    if (segment.shared) return // This segment cannot be recycled.
+    if (atomicByteCount.get() >= MAX_SIZE) return // Pool is full.
+
+    val first = firstRef.get()
+    if (first === LOCK) return // A take() is currently in progress.
+
+    segment.next = first
```
adriancole, line 92: sorry if I'm showing my lack of kotlin knowledge, but I'm wondering how these 3 changes will be guaranteed to be visible in the right order to other threads? We aren't using locks or volatile inside Segment as far as I can tell.. what am I missing? (I want to follow up with a comment to explain this later unless it is generally understood kotlin thing)

adriancole: ah right. below we are using `compareAndSet`, not `weakCompareAndSet`, and that is what's allowing these changes to be safely published to the consumer of the pool (happens-before relationship) https://docs.oracle.com/javase/6/docs/api/java/util/concurrent/atomic/package-summary.html

PR #721, opened 2020-05-26
Set up MkDocs publishing on CI

Egorand:


.github/workflows/build.yml
```diff
@@ -85,12 +85,24 @@ jobs:
...
\ No newline at end of file
+      - name: Prepare docs
+        run: .buildscript/prepare_mkdocs.sh
+
+      - name: Deploy docs
+        uses: mhausenblas/mkdocs-deploy-gh-pages@v1.12
+        env:
+          PERSONAL_TOKEN: ${{ secrets.GH_CLIPPY_TOKEN }}
+
+      - name: Checkout gh-pages
```
Egorand, line 96: Super not sure whether this works, and no idea how to test it...

swankjesse:
Delete `deploy_website.sh` while you’re at it?

Egorand:
Felt that it might still be useful if we need to force publish the docs manually. Or would we then just execute the action manually? Gonna delete the script.

PR #723, opened 2020-05-27
Spin 10 times before giving up looking for a Segment

swankjesse:


okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -59,25 +59,27 @@ internal actual object SegmentPool {
...
   @JvmStatic
   actual fun take(): Segment {
-    val first = firstRef.getAndSet(LOCK)
-    when {
-      first === LOCK -> {
-        // We didn't acquire the lock. Don't take a pooled segment.
-        return Segment()
-      }
-      first == null -> {
+    for (i in 0 until 10) {
```
JakeWharton, line 62: You could do `repeat(10) { .. }` when you don't care about loop var

swankjesse:
Obsoleted by https://github.com/square/okio/pull/725

PR #724, opened 2020-05-27
Use 16 different linked lists to reduce contention

swankjesse:
This may also increase locality for NUMA.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
@@ -48,8 +48,17 @@ internal actual object SegmentPool {
   /** A sentinel segment to indicate that the linked list is currently being modified. */
   private val LOCK = Segment(ByteArray(0), pos = 0, limit = 0, shared = false, owner = false)
 
-  /** Singly-linked list of segments. */
-  private var firstRef = AtomicReference<Segment?>()
+  /**
+   * 16 hash buckets, each containing a singly-linked list of segments. We use multiple hash buckets
+   * so different threads don't race each other. We use thread IDs as hash keys because they're
+   * handy, and because it may increase locality.
+   *
+   * We don't use [ThreadLocal] because we don't know how many threads the host process has and we
+   * don't want to leak memory for the duration of a thread's life.
+   */
+  private var hashBuckets: Array<AtomicReference<Segment?>> = Array(16) {
+    AtomicReference<Segment?>()
+  }
 
   /** Total bytes in this pool. */
   private var atomicByteCount = AtomicLong()
@@ -59,6 +68,8 @@ internal actual object SegmentPool {
 
   @JvmStatic
   actual fun take(): Segment {
+    val firstRef = firstRef()
+
     val first = firstRef.getAndSet(LOCK)
     when {
       first === LOCK -> {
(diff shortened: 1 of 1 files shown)


okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -48,8 +48,17 @@ internal actual object SegmentPool {
...
-  private var firstRef = AtomicReference<Segment?>()
+  /**
+   * 16 hash buckets, each containing a singly-linked list of segments. We use multiple hash buckets
+   * so different threads don't race each other. We use thread IDs as hash keys because they're
+   * handy, and because it may increase locality.
+   *
+   * We don't use [ThreadLocal] because we don't know how many threads the host process has and we
+   * don't want to leak memory for the duration of a thread's life.
+   */
+  private var hashBuckets: Array<AtomicReference<Segment?>> = Array(16) {
```
bnorm, line 59: nit: `val`

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -48,8 +48,17 @@ internal actual object SegmentPool {
...
 
-  /** Singly-linked list of segments. */
-  private var firstRef = AtomicReference<Segment?>()
+  /**
+   * 16 hash buckets, each containing a singly-linked list of segments. We use multiple hash buckets
+   * so different threads don't race each other. We use thread IDs as hash keys because they're
+   * handy, and because it may increase locality.
+   *
+   * We don't use [ThreadLocal] because we don't know how many threads the host process has and we
+   * don't want to leak memory for the duration of a thread's life.
```
JakeWharton, line 57: Can we solve this with a `WeakHashMap<Thread, SegmentPool>`? I've never looked into how `ThreadLocal` works... why isn't it this?

swankjesse: That’s approximately it. The problem is we have no idea whether a thread that just used 64 KiB of segments is ever coming back. It might read a file when it starts up, and then never do any Okio work for the rest of time. So the 64 KiB is effectively leaked.

JakeWharton: I see. It's a live leak, or whatever, to make up a term. Maybe the watchdog thread could help?

cketti: The worst case for this "16 hash buckets" method seems worse. If an app uses 16 threads and they happen to get consecutive thread IDs, 15 of them allocate a lot of segments while doing their thing and then stop. One remains and no new threads will be created during the lifetime of the app. The one remaining thread won't be able to access the segments in other buckets and those segments won't be garbage collected even after the threads are long gone.

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -101,4 +114,9 @@ internal actual object SegmentPool {
       segment.next = null
     }
   }
+
+  private fun firstRef(): AtomicReference<Segment?> {
+    val hashBucket = (Thread.currentThread().id and 0b1111).toInt() // Get a value in [0..16).
```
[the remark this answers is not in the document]
swankjesse: Yeah, we might be able to ask for either ID!

okio/src/jvmMain/kotlin/okio/SegmentPool.kt
swankjesse: @cketti Yep. But I’m happier to waste 16 x 64 KiB than _N_ x 64 KiB for _N_ threads.

swankjesse:
@bnorm Yeah there’s two benefits of the hash. One is to avoid contention on accessing the segments; the other is to encourage locality, where a thread that releases a segment gets it back later.

swankjesse:
Obsoleted by https://github.com/square/okio/pull/725

PR #725, opened 2020-05-29
hashBucketCount = availableProcessors; segment.freeSegmentCount

rewlad:
optimizing SegmentPool:
hashBucketCount = availableProcessors
counter moved to segment.freeSegmentCount

PR #727, opened 2020-05-31
Use multiple linked lists to reduce contention in the segment pool

swankjesse:
This adds a commit to @rewlad's [PR 725](https://github.com/square/okio/pull/725).

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
@@ -16,7 +16,6 @@
 package okio
 
 import okio.SegmentPool.LOCK
-import okio.SegmentPool.MAX_SIZE
 import okio.SegmentPool.recycle
 import okio.SegmentPool.take
 import java.util.concurrent.atomic.AtomicReference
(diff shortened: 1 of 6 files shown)


okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -36,29 +34,47 @@ import java.util.concurrent.atomic.AtomicReference
...
+  private val hashBuckets: Array<AtomicReference<Segment?>> = Array(HASH_BUCKET_COUNT) {
+    AtomicReference<Segment?>()
+  }
 
-  actual val byteCount: Long
-    get() = atomicByteCount.get()
+  actual val byteCount: Int
+    get() {
+      val first = firstRef().get() ?: return 0
+      return first.limit
```
swankjesse, line 71: @rewlad rather than trying to make this correct, I’m just redefining the field to only return the byte count accessible to the current thread. That’s certainly enough to make our tests happy!

swankjesse:
I benchmarked 3 segment pool implementations. I used 128 threads on my 16 core MacBook Pro.

### master
memory: 64 KiB
CPU: 3.2M ops/sec

### hash buckets (this PR)
memory: 16 cores * 8 KiB = 1 MiB
CPU: 9.8M ops/sec

### thread local segment pool
memory: 128 threads * 8 KiB = 8 MiB
CPU: 11.4M ops/sec

This implementation is definitely a compromise on memory vs. CPU.

swankjesse:
Forgot to link the benchmark!
https://gist.github.com/swankjesse/95bf5687ad40b6dc47386d348079a623

PR #729, opened 2020-06-01
New APIs: ByteString.toIndex() and ByteString.toFraction()

swankjesse:
The first one may be useful with hashing to put byte strings in
partitioning buckets for scaling. For example, to divide a dataset
into 32 partitions, hash the key then use toIndex(32) to map the
key to its partition.

The second one may be useful with dynamic experiments and A/B
tests. For example, to assign a control group to 5% of customers
hash the customer key then check if toFraction() is less than 0.05.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/ByteString.kt
+++ b/okio/src/commonMain/kotlin/okio/ByteString.kt
@@ -138,6 +138,98 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
   override fun compareTo(other: ByteString): Int
 
   /**
+   * Projects this value to the range `[0..size)` using linear interpolation. This is equivalent to
+   * a sorted partitioning of all possible byte strings across [size] equally-sized buckets and
+   * returning the index of the bucket that this byte string fits in.
+   *
+   * For example, the byte string `8000` is the median of all 2-element byte strings, and calling
+   * `toIndex(100)` on it returns 50. Some other examples:
+   *
+   * | Byte String (hex)  | `toIndex(100)` | `toIndex(256)` | `toIndex(Int.MAX_VALUE)` |
+   * | :----------------- | -------------: | -------------: | -----------------------: |
+   * | (empty)            |              0 |              0 |                        0 |
+   * | 00                 |              0 |              0 |                        0 |
+   * | 0000               |              0 |              0 |                        0 |
+   * | 000000             |              0 |              0 |                        0 |
+   * | 0000000001         |              0 |              0 |                        0 |
+   * | 00000001           |              0 |              0 |                        0 |
+   * | 00000002           |              0 |              0 |                        0 |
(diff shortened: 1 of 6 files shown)


okio/src/commonTest/kotlin/okio/ByteStringTest.kt
```diff
@@ -457,4 +457,72 @@ abstract class AbstractByteStringTest internal constructor(
...
+    assertEquals(0x7fffff7f, factory.decodeHex("ffffff").toIndex(Int.MAX_VALUE))
+    assertEquals(0x7ffffffe, factory.decodeHex("ffffffff").toIndex(Int.MAX_VALUE))
+  }
+
+  /**
+   * Our math is incorrect for values that round differently depending on data beyond the first 4
+   * bytes. For example, "aaaaaaaaab".toIndex(3) is 1, but if we did arbitrary-precision math the
+   * result would be 2.
+   */
+  @Test fun toIndexHonorsFirstFourBytesOnly() {
```
swankjesse, line 491: of everything, this part hurts me the most

okio/src/commonMain/kotlin/okio/ByteString.kt
```diff
@@ -137,6 +137,98 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
...
+   * [compareTo]. The returned value is also consistent with [compareTo] though the dynamic range
+   * is compressed. For two byte strings `a` and `b`, if `a < b`, then
+   * `a.toFraction() <= b.toFraction()`.
+   *
+   * This examines at most the first 7 bytes of this byte string. Data beyond the first 7 bytes is
+   * not used to compute the result.
+   *
+   * @return a value that is greater than or equal to `0.0` and less than `1.0`.
+   */
+  fun toFraction(): Double
```
zhxnlai, line 230: Great comments!

zhxnlai (approved):

JakeWharton:
Recording some out-of-band discussion points: we're looking for a word to replace "to" here since the index and fraction concepts are not intrinsic to domain of how you think of a bytestring. Jesse had a fun analogy where it's like having `String.toIndex()` which returned the absolute position of a string in a dictionary of all words. It's not perfect, of course, since strings can be arbitrary characters that aren't words. It is, however, a useful mental model since `String.toIndex()` would be a weird API but `String.dictionaryIndex()` makes sense. So what's the "dictionary" of the set of all possible bytes such that we can name these methods something like `thingIndex()` and `thingFraction()`?

swankjesse:
I've been thinking about the names here. My biggest concern with the current names is they could be interpreted as decoding ASCII, like readDecimalLong.

My new recommendation is to drop the index function, and rename toFraction() to unitIntervalDouble(). You can get the result of toIndex by dividing that by the target size.

PR #732, opened 2020-06-12
New API: Pipe.cancel()

swankjesse:
I wanted this when attempting to implement java.io.Socket on top of a
pair of Pipe instances.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Pipe.kt
+++ b/okio/src/jvmMain/kotlin/okio/Pipe.kt
@@ -15,8 +15,6 @@
  */
 package okio
 
-import java.io.IOException
-
 /**
  * A source and a sink that are attached. The sink's output is the source's input. Typically each
  * is accessed by its own thread: a producer thread writes data to the sink and a consumer thread
@@ -32,9 +30,12 @@ import java.io.IOException
  * When the sink is closed, source reads will continue to complete normally until the buffer has
  * been exhausted. At that point reads will return -1, indicating the end of the stream. But if the
  * source is closed first, writes to the sink will immediately fail with an [IOException].
+ *
+ * A pipe may be canceled to immediately fail writes to the sink and reads from the source.
  */
 class Pipe(internal val maxBufferSize: Long) {
   internal val buffer = Buffer()
+  internal var canceled = false
   internal var sinkClosed = false
   internal var sourceClosed = false
   internal var foldedSink: Sink? = null
@@ -52,6 +53,7 @@ class Pipe(internal val maxBufferSize: Long) {
       var delegate: Sink? = null
       synchronized(buffer) {
         check(!sinkClosed) { "closed" }
+        if (canceled) throw IOException("canceled")
 
         while (byteCount > 0) {
           foldedSink?.let {
@@ -64,6 +66,7 @@ class Pipe(internal val maxBufferSize: Long) {
           val bufferSpaceAvailable = maxBufferSize - buffer.size
           if (bufferSpaceAvailable == 0L) {
             timeout.waitUntilNotified(buffer) // Wait until the source drains the buffer.
+            if (canceled) throw IOException("canceled")
             continue
           }
 
@@ -81,6 +84,7 @@ class Pipe(internal val maxBufferSize: Long) {
       var delegate: Sink? = null
       synchronized(buffer) {
         check(!sinkClosed) { "closed" }
+        if (canceled) throw IOException("canceled")
 
         foldedSink?.let {
           delegate = it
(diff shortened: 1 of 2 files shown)


okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
```diff
@@ -609,6 +609,157 @@ class PipeKotlinTest {
...
 
+  @Test fun cancelPreventsSinkWrite() {
+    val pipe = Pipe(8)
+    pipe.cancel()
+
+    val pipeSink = pipe.sink.buffer()
+    pipeSink.writeUtf8("hello world")
+
+    try {
+      pipeSink.emit()
```
JayNewstrom, line 620: Are these tests missing `fail()` in order to ensure the catch block is run?

okio/src/jvmMain/kotlin/okio/Pipe.kt
```diff
@@ -32,9 +30,12 @@ import java.io.IOException
  * When the sink is closed, source reads will continue to complete normally until the buffer has
  * been exhausted. At that point reads will return -1, indicating the end of the stream. But if the
  * source is closed first, writes to the sink will immediately fail with an [IOException].
+ *
+ * A pipe may be canceled to immediately fail writes to the sink and reads from the source.
  */
 class Pipe(internal val maxBufferSize: Long) {
   internal val buffer = Buffer()
+  internal var canceled = false
```
Egorand, line 38: Should we care about any of these fields being `@Volatile`?

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
Egorand: @swankjesse even better, use `assertThrows`.

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
```diff
@@ -609,6 +609,157 @@ class PipeKotlinTest {
...
+      pipe.sink.flush()
+    } catch (e: IOException) {
+      assertEquals("canceled", e.message)
+    }
+  }
+
+  @Test fun cancelDoesNotPreventSinkClose() {
+    val pipe = Pipe(8)
+    pipe.cancel()
+    pipe.sink.close()
```
Egorand, line 640:
```kotlin
try {
  pipe.sink.close()
} catch (_: IOException) {
  fail()
}
```
?

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
```diff
@@ -609,6 +609,157 @@ class PipeKotlinTest {
...
+    val pipeSource = pipe.source.buffer()
+
+    try {
+      pipeSource.require(1)
+    } catch (e: IOException) {
+      assertEquals("canceled", e.message)
+    }
+  }
+
+  @Test fun cancelDoesNotPreventSourceClose() {
```
Egorand, line 675: `DoesNotPrevent` means it shouldn't throw?

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
swankjesse: Nah, they’re missing `fail()` because I’m clumsy. Fixed!

okio/src/jvmMain/kotlin/okio/Pipe.kt
swankjesse: No, we use `synchronized` instead.

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
swankjesse: the exception will fail the test anyway

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
swankjesse: Yeah, lemme rephrase that.

okio/src/jvmTest/kotlin/okio/PipeKotlinTest.kt
swankjesse: Can’t use `assertThrows` because we don’t depend on kotlin.test. Should do that in a follow-up change!

okio/src/jvmMain/kotlin/okio/Pipe.kt
```diff
@@ -15,8 +15,6 @@
  */
 package okio
 
-import java.io.IOException
```
PromanSEW, line 18: Why?

swankjesse: We’ve got our own `okio.IOException` that’s aliased to `java.io.IOException`. No behavior change here.

Issue #740, opened 2020-07-15
Implementing ByteString hashing algorithms in common Kotlin


kevincianfarini:
I have a need for hashing algorithms available on `ByteString` outside of the JVM. Specifically, the BitTorrent protocol uses SHA-1 to verify that file pieces contain the correct data. Right now, it seems that hashing relies on the JVM `MessageDigest` implementation. 

As a proof of concept, I have an implementation of SHA-1 implemented on my own version of ByteString in my project Corbit. You can see it [here](https://github.com/kevincianfarini/corbit/blob/master/corbit-binary/src/commonMain/kotlin/com/corbit/binary/internal/-Sha1.kt).

swankjesse:
Yeah that'd be great. Need to do this.

kevincianfarini:
I could port my implementation of SHA-1 to Okio ByteString if you'd like

JakeWharton:
Are we going to use it on the JVM, too?

Perhaps it is time to create a source set that corresponds to all non-JVM platforms to host shared, non-JVM code.

kevincianfarini:
My intention while creating this issue was to implement our own hashes that supplement the JVM on other platforms. 

Is it even possible to create a source set that's something like nonJvmCommon?

JakeWharton:
Yep!

Example: https://android-review.googlesource.com/c/platform/frameworks/support/+/1269842/18/collection/collection2/build.gradle#79

kevincianfarini:
Whoa, sweet. 
I remember reading something a wile back that allows you to establish a hierarchy for muiltiplatform source sets in kotlin. Might be part if the 1.4 release. 

That could be of some interest too. Let me see if I can dig it up.

JakeWharton:
Yeah that linked example is using 1.3.60 or 1.3.70 so you can do it with any existing version.

The 1.4 feature lets you build up trees of shared code which is much more common whereas that link is using arbitrary shared source sets where you can build any DAG you want, not just a tree.

kevincianfarini:
Yup that's exactly right 😁

As far as implementation goes, do we think it makes sense to do something like `expect class MessageDigest` and then manually implement for non JVM platforms? 

Hashing in common Kotlin could also some day be used in `HashingSink` so it probably doesn't make sense to tightly couple it with `ByteString`

kevincianfarini:
Or maybe even have own idiomatic implementation where on JVM we just delegate to JVM MessageDigest

JakeWharton:
[...]

PR #741, opened 2020-07-17
DRAFT: Add common MessageDigest implementation 

kevincianfarini:
This PR is currently a work in progress to be developed out in the open. 

Tasks to be done:

* make a `nonJvmMain` sourceSet for this to reside in
* make typeAlias of `MessageDigest` in JVM module
* implement SHA-1 algorithm
* add tests

Relevant: #740


okio/src/commonMain/kotlin/okio/internal/OkioMessageDigest.kt
```diff
@@ -0,0 +1,24 @@
...
+
+  companion object {
+    fun getInstance(algorithm: OkioMessageDigestAlgorithm): OkioMessageDigest = when (algorithm) {
+      OkioMessageDigestAlgorithm.SHA_1 -> Sha1MessageDigest()
+      else -> TODO()
+    }
+  }
+}
+
+internal expect class Sha1MessageDigest constructor(): OkioMessageDigest
```
JakeWharton, line 24: We shouldn't need one class per hash, at least not on the JVM.

I think the shape of this should be
```kotlin
interface OkioMessageDigest {
  fun update(..)
  fun digest(..)
}

expect fun newMessageDigest(algorithm: String): OkioMessageDigest
```
Then JVM can just delegate to `MessageDigest.getInstance` and non-JVM can just do the switch to whatever it needs.

kevincianfarini: Love it. Will make that change.

kevincianfarini: stubs updated.

okio/src/commonMain/kotlin/okio/OkioMessageDigestAlgorithm.kt
```diff
@@ -0,0 +1,8 @@
+package okio
+
+enum class OkioMessageDigestAlgorithm(val value: String) {
```
JakeWharton, line 3: I would drop this. While the type safety is nice, we only expose the available options as functions rather than accept it as a parameter in the public API. We can just use strings internally and save the class.

kevincianfarini: This scratches the surface of something I'm slightly concerned about. 

Because were separating modules of `commonMain`, `commonTest`, `nonJvmMain`, etc. We can't directly test `OkioMessageDigest` if its marked internal. 

Currently, `newMessageDigest` and `OkioMessageDigest` are public. I don't know if a better approach is to test this message digest by proxy via `ByteString` and `HashingSink`. 

Thoughts?

JakeWharton: Internal visibility should be available from tests. The IDE probably just isn't smart enough, but it should compile from the command line.

With Kotlin 1.4 the idea of a non-JVM source set will get more first-class support.

JakeWharton: Oh but yeah to answer your question, using the public API from commonTest is far-preferred. We want to use the same codepaths as our users, not rely on implementation details.

kevincianfarini: Okay, I'll give this a go tomorrow and see how it plays out. Thanks 🙂

kevincianfarini: > Oh but yeah to answer your question, using the public API from commonTest is far-preferred. We want to use the same codepaths as our users, not rely on implementation details.

To clarify, meaning the tests backing this should go through ByteString?

JakeWharton: Yep!

okio/src/commonMain/kotlin/okio/internal/ByteString.kt
```diff
@@ -288,6 +289,10 @@ internal fun ByteString.commonWrite(buffer: Buffer, offset: Int, byteCount: Int)
   buffer.write(data, offset, byteCount)
 }
 
+internal fun ByteString.commonDigest(algorithm: String) = ByteString(
+  newMessageDigest(algorithm).apply { update(data) }.digest()
```
swankjesse, line 293: FWIW, this is too much stuff in one line for me

okio/src/commonTest/kotlin/okio/OkioMessageDigestTest.kt
```diff
@@ -0,0 +1,81 @@
...
+import okio.internal.commonAsUtf8ToByteArray
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class OkioMessageDigestTest {
+
+  // region SHA-1
+
+  @Test fun sha1EmptyValueHashIsCorrect() {
+    // arrange act
```
swankjesse, line 13: omit these comments? it feels like training wheels

okio/src/nonJvmMain/kotlin/okio/ByteString.kt
```diff
@@ -73,7 +88,10 @@ internal actual constructor(
   actual open fun substring(beginIndex: Int, endIndex: Int): ByteString =
     commonSubstring(beginIndex, endIndex)
 
-  internal actual open fun internalGet(pos: Int) = commonGetByte(pos)
+  internal actual open fun internalGet(pos: Int): Byte {
+    if (pos >= size || pos < 0) throw ArrayIndexOutOfBoundsException("size=$size pos=$pos")
```
swankjesse, line 92: why is this changing? I think the whole motivation of _internal_ in the name is it’s not going to do a bounds check?

okio/src/nonJvmMain/kotlin/okio/OkioMessageDigest.kt
```diff
@@ -0,0 +1,13 @@
+package okio
```
swankjesse, line 1: Mind adding copyright headers?

okio/src/nonJvmMain/kotlin/okio/internal/Sha1MessageDigest.kt
```diff
@@ -0,0 +1,149 @@
...
+    else -> error("$index is out of bounds")
+  }
+}
+
+private fun UInt.getByte(index: Int): Byte {
+  require(index < 4)
+  return ((this shr ((3 - index) * 8)) and 0xff.toUInt()).toByte()
+}
+
+private fun ByteArray.chunked(chunkSize: Int): List<ByteArray> {
```
swankjesse, line 116: The implementation of this shouldn’t allocate

swankjesse:
Can you please complete Square’s contributor agreement? We want to be able to redistribute your hard work!

swankjesse:

kevincianfarini:
> Multiplatform stuff looks good. I think we should rewrite the actual hash function to follow ‘inner loop’ programming practices, which is often quite the opposite of nice maintainable code. If you’d like, we can merge first and optimize in follow-up.

I'm actually doing some work right now on 256 impl. Could you link to what "inner loop" programming is and some examples? 

I could try and dip my toes in this PR but i'd like to read up on it first!

kevincianfarini:
> Can you please complete Square’s contributor agreement? We want to be able to redistribute your hard work!

Absolutely!

okio/src/nonJvmMain/kotlin/okio/ByteString.kt
kevincianfarini: This was a change made when consolidating all non jvm implementations of bytestring. This was directly copied from the (now deleted) js implementation of ByteString. I had to add this bounds check to throw the appropriate error. 

Without this, tests were failing. I can't remember the exact assertion error before, I'm going to replicate and see what it was.

okio/src/nonJvmMain/kotlin/okio/internal/Sha1MessageDigest.kt
kevincianfarini: I agree. As far as I'm aware, there's no way to do views into arrays that dont copy. If there is please point it out. 

Given that there's not, we could introduce a `ByteArraySlice` type that uses some offsets for this purpose. 

I imagine we also shouldn't be using `ByteArray.plus` like shown for this reason. 

```kotlin
for (chunk in (unprocessed + input).chunked(...))
```

okio/src/commonMain/kotlin/okio/internal/ByteString.kt
kevincianfarini: I'll break it up

okio/src/commonTest/kotlin/okio/OkioMessageDigestTest.kt
kevincianfarini: Sure thing.

kevincianfarini:
I'm getting this failure. 

```
okio.AsyncTimeoutTest[jvm] > sinkSplitsLargeWrites[jvm] FAILED
    java.io.InterruptedIOException at AsyncTimeoutTest.java:357
```

I imagine that this is expected to happen from time to time?

swankjesse:
Yeah, flaky tests are so annoying!

kevincianfarini:
I'm going to assume that this is a flaky test as well until told otherwise. 

```
okio.PipeTest[jvm] > sourceBlocksOnSlowWriter[jvm] FAILED
    java.lang.AssertionError at PipeTest.java:370
```

SHA-512 implementation pushed!

okio/src/nonJvmMain/kotlin/okio/OkioMessageDigest.kt
kevincianfarini: added to all files

kevincianfarini:
Alright. I'd like to request some full feature review now. The following algorithms have been ported to common kotlin:

* SHA-1
* SHA-256
* SHA-512
* MD5

I've addressed some of the in flight review comments. To address the bigger ones, I've added an internal `Bytes` type which allows us to do allocation free operations on a collections of bytes like slicing. 

I look forward to the feedback :)

kevincianfarini:
I've also completed the contributor agreement. Please let me know if there's anything else I need to do.

kevincianfarini:
> ‘inner loop’ programming practices, which is often quite the opposite of nice maintainable code. If you’d like, we can merge first and optimize in follow-up.

I agree. Looking at a lot of reference solutions for the algorithms, I saw that manual loop unrolling was common. This makes sense to avoid branchy code in loops like the following. 

```kotlin
    for (i in 0 until 80) {
      val (f, k) = when (i) {
        in 0..19 -> (d xor (b and (c xor d))) to 0x5A827999.toUInt()
        in 20..39 -> (b xor c xor d) to 0x6ED9EBA1.toUInt()
        in 40..59 -> ((b and c) or (b and d) or (c and d)) to 0x8F1BBCDC.toUInt()
        in 60..79 -> (b xor c xor d) to 0xCA62C1D6.toUInt()
        else -> error("Index is wonky, this should never happen")
      }
```

swankjesse:
Will review enthusiastically! Please don’t unroll too much yet; it’s easier to review for correctness in the first pass!

kevincianfarini:
I wasn't planning on it. Preferably, I'd have that be a follow up PR. This one has had some scope creep :)

okio/src/nonJvmMain/kotlin/okio/internal/AbstractMessageDigest.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  override fun digest(): ByteArray {
+    val finalMessageLength = messageLength + unprocessed.size
+
+    val finalMessage = byteArrayOf(
+      *unprocessed.toByteArray(),
+      0x80.toByte(),
+      *ByteArray(((56 - (finalMessageLength + 1) % 64) % 64).toInt()),
```
swankjesse, line 45: what’s the size of this array if `finalMessageLength` is 56 ? Ie. if we attempt to process a single 56-element array?

okio/src/nonJvmMain/kotlin/okio/internal/Bytes.kt
```diff
@@ -0,0 +1,127 @@
...
+ * limitations under the License.
+ */
+
+package okio.internal
+
+import kotlin.math.min
+
+/**
+ * A convenience wrapper around [ByteArray] to allow for
+ * zero allocation operations.
```
swankjesse, line 23: I suspect this is likely a performance regression because you’re replacing array accesses with polymorphic method calls. Optimizing this kind of thing is not intuitive!

swankjesse: (In any case, we should fix in follow-up)

swankjesse: ... also this says “zero allocation” but it allocates slices all over the place.

okio/src/nonJvmMain/kotlin/okio/internal/Bytes.kt
```diff
@@ -0,0 +1,127 @@
...
+    accumulator = accumulator or ((byte.toULong() and 0xffUL) shl ((7 - index) * 8))
+  }
+
+  return accumulator
+}
+
+internal fun Bytes.chunked(chunkSize: Int): List<Bytes> {
+  val result = mutableListOf<Bytes>()
+
+  val lastIndex = if (size % chunkSize == 0) this.size else (size / chunkSize) + size
```
swankjesse, line 119: nit: weird to mix `this.size` and `size` on the same line – as far as I can tell they’re the same?

okio/src/nonJvmMain/kotlin/okio/internal/MD5MessageDigest.kt
```diff
@@ -0,0 +1,133 @@
...
+      0x80.toByte(),
+      *ByteArray(((56 - (finalMessageLength + 1) % 64) % 64).toInt()),
+      *(finalMessageLength * 8L).toLittleEndianByteArray()
+    ).toBytes()
+
+    finalMessage.chunked(64).forEach { chunk ->
+      currentDigest = processChunk(chunk, currentDigest)
+    }
+
+    println(currentDigest)
```
swankjesse, line 84: ick

okio/src/nonJvmMain/kotlin/okio/internal/MD5MessageDigest.kt
```diff
@@ -0,0 +1,133 @@
...
+        64 -> {
+          currentDigest = processChunk(chunk, currentDigest)
+          messageLength += 64
+        }
+        else -> unprocessed = chunk
+      }
+    }
+  }
+
+  override fun digest(): ByteArray {
```
swankjesse, line 70: duplicated from AbstractMessageDigest?

okio/src/nonJvmMain/kotlin/okio/internal/MD5MessageDigest.kt
kevincianfarini: whoops!

okio/src/nonJvmMain/kotlin/okio/internal/MD5MessageDigest.kt
kevincianfarini: I wrestled with how I should handle this as well. All message digest algorims besides MD5 use big endian byte ordering. MD5 on the other hand uses little endian. 

I figured at the time that abstracting that away and into a parent class was over abstraction, and instead preferred the duplication. You'll notice bits like padding the final message chunk use little endian instead of big endian.

kevincianfarini: If you think this should be abstracted that can be worked in :)

swankjesse: Nah, the explanation was all I needed.

okio/src/nonJvmMain/kotlin/okio/internal/AbstractMessageDigest.kt
kevincianfarini: The resulting array would be 63 bytes. Given `finalMessageLength` is 55 bytes, the resulting array would be 0 bytes. 

Written more verbosely, the expression in question can be written as the following.

```kotlin
while (finalMessageLength != 56) {
    finalMessage.append(0.toByte())
}
```

We do this because the message needs to be padded with 0 bytes such that we have 8 bytes left over for the length of the message to be appended in byte representation. 

The wikipedia explanation might make more sense than my explanation.

> append 0 ≤ k < 512 bits '0', such that the resulting message length in bits is congruent to −64 ≡ 448 (mod 512)

kevincianfarini:
I will work on breaking this PR up a bit tonight :)

kevincianfarini:
> In the first PR we should land implementations only that aren’t exposed in any public API. These implementations should be in common and run on any platform including JVM. This PR should take this implementation as it is now and strip out API changes.

I take it then that we're abandoning the `nonJvm` sourcesets? If that's the case, I might close this draft PR and open a PR that's more focused.

okio/src/nonJvmMain/kotlin/okio/internal/Bytes.kt
kevincianfarini: "zero copy" it is.

I wouldn't be surprised if this is a performance problem. In the event that it is, I will probably abandon the notion of a `chunk` altogether.

kevincianfarini:
> In this phase I’d like to abandon MessageDigest and do something closer to what the hashing algorithms want. Perhaps something like this?

> ```
> interface HashAlgorithm {
>   val chunkSize: Int
>   fun process(bytes: ByteArray, offset: Int, digest: HashDigest)
> }
> ```

Could you explain this more clearly? I don't follow. What would offset do here?

okio/src/nonJvmMain/kotlin/okio/internal/AbstractMessageDigest.kt
swankjesse: I think it's gonna get -1 instead of 63 here.

swankjesse:
I think the code might be simpler if it doesn't have to consider slicing or state management. Instead pass in the internal state (HashDigest) and also do the chunking externally.

But this might be better guided by benchmarks.

okio/src/nonJvmMain/kotlin/okio/internal/AbstractMessageDigest.kt
kevincianfarini: Aaaaaaaand you're absolutely correct. Will fix in update.

swankjesse: You can probably find a way to do `% 64` only once!

kevincianfarini: Yeah, did some research. The python implementation of mod (which i was using as a reference) by default has absolute modulus and instead returns the positive complement. 

I have an `absoluteMod` function written now which seems to do the job. Will write an accompanying test too :)

kevincianfarini:
On the topic of separating this into separate PRs, does it make sense to have a `common-hashing` branch which these PRs would merge into? That way we could have the unit tests going through the public api path like @JakeWharton mentioned above. 

> Oh but yeah to answer your question, using the public API from commonTest is far-preferred. We want to use the same codepaths as our users, not rely on implementation details.

okio/src/nonJvmMain/kotlin/okio/internal/AbstractMessageDigest.kt
kevincianfarini: Fixed :)

swankjesse:
I only like merging to master. Feature branches are a bad idea.

kevincianfarini:
I'm in the process of splitting this up into separate PRs. As part of that, I'm shifting all of the unit tests to directly call internal code. I'm having trouble getting the compiler to recognize that internal code in `commonMain` should be visible in `commonTest`. In a personal project, I've been able to do it perfectly fine without any additional setup. 

```kotlin
package okio.internal

internal fun newMessageDigest(algorithm: String) = when (algorithm) {
  "SHA-1" -> Sha1MessageDigest()
  "SHA-256" -> Sha256MessageDigest()
  "SHA-512" -> Sha512MessageDigest()
  "MD5" -> MD5MessageDigest()
  else -> error("Invalid algorithm $algorithm")
}
```

When trying to call this function from `commonTest` I get the following compiler error on the command line with gradle. 

```
e: /home/kevin/code/okio/okio/src/commonTest/kotlin/okio/OkioMessageDigestTest.kt: (29, 18): Unresolved reference: newMessageDigest
```

[This issue](https://youtrack.jetbrains.com/issue/KT-38842) seems to be loosely tied to what I'm experiencing, but I'm unsure because I'm unable to see the sample project the reporter mentions. 

Any guidance would help a bunch :)

kevincianfarini:
#758 adds just an internal message digest as well corresponding tests

swankjesse:
This has all landed on master. 🎉 Thanks Kevin!

PR #746, opened 2020-08-09
Make samples runnable again

Egorand:


samples/build.gradle
```diff
@@ -1,4 +1,7 @@
 apply plugin: 'org.jetbrains.kotlin.multiplatform'
+apply plugin: 'application'
+
+mainClassName = System.getProperty("mainClass")
```
Egorand, line 4: Unless I'm missing something, the `application` plugin doesn't have a built in way to pass in the main class from the CLI, this line makes it possible to do:
```
./gradlew -DmainClass=okio.samples.Test run
```

swankjesse: I'd just been running these from IntelliJ.

Egorand: That didn't work for me for some reason, nothing happens when I tap the Run button.

PR #747, opened 2020-08-09
BitmapEncoder sample in Kotlin

Egorand:


samples/src/jvmMain/kotlin/okio/samples/BitmapEncoder.kt
```diff
@@ -0,0 +1,114 @@
...
+import okio.sink
+import java.io.File
+import java.io.IOException
+import kotlin.math.hypot
+
+class KotlinBitmapEncoder {
+  class Bitmap(
+    private val pixels: Array<IntArray>
+  ) {
+    fun width(): Int = pixels[0].size
```
swankjesse, line 29: These want to be vals with `get()` methods, right?

Egorand: Yep yep, good call! Will fix.

Egorand: I actually went with computed vals, since `pixels` never change.

PR #748, opened 2020-08-09
ReadFileLineByLine example in Kotlin

Egorand:


README.md
```diff
@@ -162,6 +197,19 @@ public void readLines(File file) throws IOException {
 }
 ```
 
+Here's a similar example written in Kotlin:
+
+```kotlin
+@Throws(IOException::class)
+fun readLines(file: File) {
+  file.source().buffer().use { source ->
+    generateSequence { source.takeUnless { it.exhausted() }?.readUtf8LineStrict(1024L) }
```
swankjesse, line 206: wouldn’t it be simpler to use a loop?

The example above is nice, but this one is worse than loops. Especially this awkward line with two lambdas and an Elvis.

swankjesse: (The Java looks _so_ much simpler than the Kotlin here)

bnorm: If you wanted to keep the Sequence, you could use something like this instead.
```kotlin
sequence { while (!source.exhausted()) yield(source.readUtf8LineStrict(1024L) }
```

Egorand: Nice! Will try Brian's suggestion and revert to the loop if it still feels wrong.

PR #754, opened 2020-08-12
GoldenValue sample in Kotlin

Egorand:


Egorand:
Oh crap, thanks for flagging @eapen. I'll redeploy the site to make sure it picked up the MkDocs config changes I made to support tabs.

Egorand:
Should be fixed now.

PR #758, opened 2020-08-17
Part 1/3: Add internal OkioMessageDigest

kevincianfarini:


okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
```diff
@@ -0,0 +1,58 @@
...
+
+internal fun Long.toLittleEndianByteArray(): ByteArray = ByteArray(8) { index ->
+  ((this shr (index * 8)) and 0xffL).toByte()
+}
+
+/**
+ * Left rotate an unsigned 32 bit integer by [bitCount] bits
+ */
+internal infix fun UInt.leftRotate(bitCount: Int): UInt {
+  return ((this shl bitCount) or (this shr (UInt.SIZE_BITS - bitCount))) and UInt.MAX_VALUE
```
swankjesse, line 36: what does the last `and` operation do? Shouldn’t be necessary?

okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * Right rotate an unsigned 32 bit integer by [bitCount] bits
+ */
+internal infix fun UInt.rightRotate(bitCount: Int): UInt {
+  return (((this shr bitCount) or (this shl (UInt.SIZE_BITS - bitCount))) and UInt.MAX_VALUE)
```
swankjesse, line 43: same question

okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
```diff
@@ -0,0 +1,58 @@
...
+}
+
+internal infix fun ULong.rightRotate(bitCount: Int): ULong {
+  return (((this shr bitCount) or (this shl (ULong.SIZE_BITS - bitCount))) and ULong.MAX_VALUE)
```
swankjesse, line 47: same question

okio/src/commonMain/kotlin/okio/internal/MD5MessageDigest.kt
```diff
@@ -0,0 +1,129 @@
...
+
+  override fun digest(): ByteArray {
+    val finalMessageLength = messageLength + unprocessed.size
+
+    val finalMessage = byteArrayOf(
+      *unprocessed.toByteArray(),
+      0x80.toByte(),
+      *ByteArray((56 - (finalMessageLength + 1) absMod 64).toInt()),
+      *(finalMessageLength * 8L).toLittleEndianByteArray()
+    ).toBytes()
```
swankjesse, line 76: this awkward line is probably better written as Buffer operations

```
val finalMessage = Buffer()
    .write(unprocessed)
    .writeByte(0x80)
    .writeInt(56 - (finalMessageLength + 1) absMod 64)
    .writeLongLe(finalMessageLength * 8L)
    .readByteArray()
```

okio/src/commonMain/kotlin/okio/internal/MD5MessageDigest.kt
```diff
@@ -0,0 +1,129 @@
...
+        )
+        in 32 until 48 -> Pair(
+          b xor c xor d,
+          ((3 * i) + 5) % 16
+        )
+        in 48 until 64 -> Pair(
+          c xor (b or d.inv()),
+          (7 * i) % 16
+        )
+        else -> error("Index is wonky, this should never happen")
```
swankjesse, line 112: just omit the message? error() or error("unexpected") is enough

okio/src/commonMain/kotlin/okio/internal/OkioMessageDigest.kt
```diff
@@ -0,0 +1,16 @@
+package okio.internal
```
swankjesse, line 1: nit: copyright header

swankjesse:
Great PR! I’m merging as-is though I’d prefer for you to address the feedback in the next PR!

kevincianfarini:
Thanks! Theres some 1.4 stuff I can use to replace some helper functions I've written as well.

okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
kevincianfarini: I was planning on removing `left/rightRotate` altogether in favor of the stdlib `rotateRight` and `rotateLeft`. These require stdlib opt in. Are we allowed to do that in Okio?

JakeWharton: No. We can only depend on stable API.

okio/src/commonMain/kotlin/okio/internal/MD5MessageDigest.kt
kevincianfarini: Love it!

PR #762, opened 2020-09-06
Optimize SHA-1.

swankjesse:
```
Benchmark                    (algorithm)  (messageSize)  Mode  Cnt     Score     Error  Units
MessageDigestBenchmark.jvm         SHA-1            100  avgt    3     0.408 ±   0.152  us/op
MessageDigestBenchmark.jvm         SHA-1        1048576  avgt    3  2867.138 ± 326.541  us/op
MessageDigestBenchmark.okio        SHA-1            100  avgt    3     0.546 ±   0.029  us/op
MessageDigestBenchmark.okio        SHA-1        1048576  avgt    3  2968.167 ± 430.262  us/op
```

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
@@ -36,6 +36,10 @@ internal infix fun UInt.leftRotate(bitCount: Int): UInt {
   return ((this shl bitCount) or (this shr (UInt.SIZE_BITS - bitCount))) and UInt.MAX_VALUE
 }
 
+internal inline infix fun Int.leftRotate(bitCount: Int): Int {
+  return (this shl bitCount) or (this ushr (32 - bitCount))
+}
+
 /**
  * Right rotate an unsigned 32 bit integer by [bitCount] bits
  */
--- a/okio/src/commonMain/kotlin/okio/internal/Sha1MessageDigest.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/Sha1MessageDigest.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Square, Inc.
+ * Copyright (C) 2020 Square, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


swankjesse:
FYI @kevincianfarini

okio/src/commonMain/kotlin/okio/internal/Sha256MessageDigest.kt
```diff
@@ -16,6 +16,7 @@
 
 package okio.internal
 
+@ExperimentalUnsignedTypes
```
JakeWharton, line 19: We can't ship with these, so they'll have to migrate to signed values or some other way of representing the constants eventually.

JakeWharton (approved):

okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
```diff
@@ -36,6 +36,10 @@ internal infix fun UInt.leftRotate(bitCount: Int): UInt {
   return ((this shl bitCount) or (this shr (UInt.SIZE_BITS - bitCount))) and UInt.MAX_VALUE
 }
 
+internal inline infix fun Int.leftRotate(bitCount: Int): Int {
```
kevincianfarini, line 39: Are we intending to move away from unsigned typed because they're experimental? 

We might be able to run some quick calculations to map our initial hash values and round constants from `UInt` to `Int` if that's the case.

okio/src/commonMain/kotlin/okio/internal/Sha1MessageDigest.kt
```diff
@@ -13,63 +13,174 @@
...
-internal class Sha1MessageDigest : AbstractMessageDigest() {
+@ExperimentalUnsignedTypes
+internal class Sha1MessageDigest : OkioMessageDigest {
+  private var messageLength = 0L
+  private val unprocessed = ByteArray(64)
+  private var unprocessedLimit = 0
+  private val words = IntArray(80)
+
+  private var h0: Int = 0x67452301
+  private var h1: Int = 0xEFCDAB89u.toInt()
```
kevincianfarini, line 26: For example from the comment above...we could open a scratch file and calculate this value and replace it with the proper `Int` value. That way we could drop unsigned typed altogether. 

However, I'm not sure if that's your intention!

kevincianfarini (commented):
Just a few thoughts :)

okio/src/commonMain/kotlin/okio/internal/Sha256MessageDigest.kt
swankjesse: Agreed. I’m going to keep ’em as-is for now for readability, and I’ll strip em out very soon!
Tracking so I don’t miss it. https://github.com/square/okio/issues/763

okio/src/commonMain/kotlin/okio/internal/HashUtil.kt
swankjesse: Yep.

okio/src/commonMain/kotlin/okio/internal/Sha1MessageDigest.kt
swankjesse: Yeah, I’d like to go away from unsigned types. Rather annoyingly there’s few nice ways to do it until unsigned types become a part of the language.

PR #764, opened 2020-09-07
MessageDigestConsistencyTest

swankjesse:
Confirm we're consistent with the JDK's behavior for different
interaction patterns.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/AbstractMessageDigest.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/AbstractMessageDigest.kt
@@ -28,6 +28,7 @@ internal abstract class AbstractMessageDigest : OkioMessageDigest {
     byteCount: Int
   ) {
     val bytes = unprocessed + input.toBytes().slice(offset until offset + byteCount)
+    unprocessed = Bytes.EMPTY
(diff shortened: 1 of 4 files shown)


okio/src/commonMain/kotlin/okio/internal/AbstractMessageDigest.kt
```diff
@@ -28,6 +28,7 @@ internal abstract class AbstractMessageDigest : OkioMessageDigest {
     byteCount: Int
   ) {
     val bytes = unprocessed + input.toBytes().slice(offset until offset + byteCount)
+    unprocessed = Bytes.EMPTY
```
swankjesse, line 31: FYI @kevincianfarini I found a bug where we weren’t clearing unprocessed if the unprocessed and new data added up to an exact multiple of 64 !

kevincianfarini: Ouch! It had actually occured to me earlier today that my PR was lacking tests that called update multiple times. 

Great catch!

swankjesse:
Obsoleted by https://github.com/square/okio/pull/779

PR #765, opened 2020-09-07
Take the allocations out of our SHA-256 implementation

swankjesse:
Performance is still quite a ways off OpenJDK.

```
Benchmark                    (algorithm)  (messageSize)  Mode  Cnt     Score     Error  Units
MessageDigestBenchmark.jvm       SHA-256            100  avgt    3     0.364 ±   0.085  us/op
MessageDigestBenchmark.jvm       SHA-256        1048576  avgt    3  2620.091 ± 645.504  us/op
MessageDigestBenchmark.okio      SHA-256            100  avgt    3     0.793 ±   0.137  us/op
MessageDigestBenchmark.okio      SHA-256        1048576  avgt    3  4468.063 ± 327.770  us/op
```

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/Sha256MessageDigest.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/Sha256MessageDigest.kt
@@ -16,63 +16,139 @@
 
 package okio.internal
 
+import okio.and
+
 @ExperimentalUnsignedTypes
-private val k = uintArrayOf(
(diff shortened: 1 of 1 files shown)


okio/src/commonMain/kotlin/okio/internal/Sha256MessageDigest.kt
```diff
@@ -16,63 +16,139 @@
 
 package okio.internal
 
+import okio.and
```
JakeWharton, line 19: Littering and... Littering and... Littering and...

JakeWharton (approved):

PR #777, opened 2020-09-25
Added OSGi metadata via build in a non intrusive way

### LINKED ISSUE #776: Include OSGi metadata
I would like to discuss adding OSGi metadata to the square libraries (okio/okhttp/okhttp "addon" libraries). I have found references that adding OSGi support has been tried already here:
https://github.com/square/okio/pull/179
https://github.com/square/okio/pull/183

https://github.com/square/okhttp/pull/2192
https://github.com/square/okhttp/pull/2246

There the discussions seem to be about that the build included more classes than needed and "the whole thing lost my trust". Not providing OSGi metadata wastes many hours in many enterprises around the world especially because your library got so popular that many libraries in the enterprise world have a dependency on okio/okhttp in some way. Repacking the libraries or relying on thirdparty repacked libraries is a waste of time for many developers relying on OSGi based frameworks.

I would like to propose a way to provide OSGi metadata in a way that does not touch any of your classes nor the jar generation process itself. OSGi itself does not need "special" class files or "more classes than needed" what OSGi needs is a META-INF/MANIFEST.MF file with additional properties - that's it. The apache felix bundle plugin used in previous attempts to add OSGi metadata do more than what's required for an OSGi compatible jar file. 

Therefore I propose the usage of the bnd gradle plugin instead of the apache felix bundle plugin. BND is the de facto standard for generating OSGi metadata in META-INF/MANIFEST.MF files. One of the developers of BND itself is the CTO of the OSGi Alliance. I have added OSGi metadata to many projects in the past in a non intrusive way like in this pull request: https://github.com/zeroturnaround/zt-exec/pull/85/files. In this pull request the way the jar file is built is NOT touched. What the plugin (maven) does in this case is simply generating a MANIFEST.MF file and we are instructing the "maven-jar-plugin" to simply use the generated MANIFEST.MF file from the bnd maven plugin that's it. Therefore in the worst case the MANIFEST.MF is broken somehow but no class files are touched/included/excluded or whatever.

I still have to figure out if there is a way to use the gradle bnd plugin in a way like I did with the maven plugin for "zt-exec" to simply generate a MANIFEST.MF file that will be used by the jar tasks (analogous to the way maven way in the mentioned pull request).

Let me know what you think about it and have an open, constructive discussion about it.

swimmesberger:
Implementation of the build discussed in the issue #776

Fixes #776

### DIFF
--- a/okio/jvm/jvm.gradle
+++ b/okio/jvm/jvm.gradle
@@ -4,40 +4,20 @@ apply plugin: 'ru.vyarus.animalsniffer'
 kotlin.targets.matching { it.platformType.name == 'jvm' }.all { target ->
   target.project.sourceCompatibility = JavaVersion.VERSION_1_7
   target.project.targetCompatibility = JavaVersion.VERSION_1_7
-
-  // task to generate a OSGi compatible JAR file
-  tasks.register('bundle', rootProject.bndBundleTaskClass) {
-    from sourceSets.main.output
-    destinationDir = file("$buildDir/bundle")
-    bnd = '''
+  
+  // configure the kotlin JVM jar task
+  tasks['jvmJar'].configure { t ->
+    // the bnd task conention modifies this jar task accordingly
+    def bndConvention = bndBundleTaskConventionClass.newInstance(t);
+    // modify these lines for MANIFEST.MF properties or for specific bnd instructions
+    bndConvention.bnd = '''
     Export-Package: okio
     Automatic-Module-Name: okio
     Bundle-SymbolicName: com.squareup.okio
     '''
-  }
-  
-  // extract only the MANIFEST.MF file from the OSGi compatible JAR file
-  tasks.register('bundleManifest', Copy) {
-     dependsOn bundle
-     // filter for the manifest
-     from zipTree(bundle.outputs.files.singleFile).matching {
-        include 'META-INF/MANIFEST.MF'
-     }
-     // flatten
-     eachFile {
-       path = name
-     }
-     includeEmptyDirs = false
-     // output directory for the manifest
-     into "$buildDir/bundle" 
-  }
-
-  // configure the kotlin JVM jar task
-  tasks['jvmJar'].configure {
-    dependsOn bundleManifest
-    // merge the manifest from the BND plugin
-    manifest {
-      from("$buildDir/bundle/MANIFEST.MF")
+    // call the convention when the task has finished to modify the jar to contain OSGi metadata
+    t.doLast {
+      bndConvention.buildBundle()
     }
   }
 
--- a/okio/jvm/jvm.gradle
+++ b/okio/jvm/jvm.gradle
@@ -5,9 +5,8 @@ kotlin.targets.matching { it.platformType.name == 'jvm' }.all { target ->
   target.project.sourceCompatibility = JavaVersion.VERSION_1_7
   target.project.targetCompatibility = JavaVersion.VERSION_1_7
   
-  // configure the kotlin JVM jar task
   tasks['jvmJar'].configure { t ->
-    // the bnd task conention modifies this jar task accordingly
+    // the bnd task convention modifies this jar task accordingly
     def bndConvention = bndBundleTaskConventionClass.newInstance(t);
     // modify these lines for MANIFEST.MF properties or for specific bnd instructions
     bndConvention.bnd = '''
--- a/okio/jvm/jvm.gradle
+++ b/okio/jvm/jvm.gradle
@@ -1,14 +1,3 @@
-buildscript {
-  repositories {
-    maven {
-      url "https://plugins.gradle.org/m2/"
-    }
-  }
-  dependencies {
-    classpath deps.bnd
-  }
-}
-
 apply plugin: 'java-library'
 apply plugin: 'ru.vyarus.animalsniffer'
 
@@ -17,7 +6,7 @@ kotlin.targets.matching { it.platformType.name == 'jvm' }.all { target ->
   target.project.targetCompatibility = JavaVersion.VERSION_1_7
 
   // task to generate a OSGi compatible JAR file
-  tasks.register('bundle', aQute.bnd.gradle.Bundle) {
+  tasks.register('bundle', project.bndBundleTaskClass) {
     from sourceSets.main.output
     destinationDir = file("$buildDir/bundle")
     bnd = '''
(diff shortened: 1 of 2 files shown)


okio/jvm/jvm.gradle
```diff
@@ -1,13 +1,54 @@
+buildscript {
+  repositories {
+    maven {
+      url "https://plugins.gradle.org/m2/"
```
JakeWharton, line 4: This is already declared in the root `build.gradle` as `gradlePluginPortal()`.

okio/jvm/jvm.gradle
```diff
@@ -1,13 +1,54 @@
...
+    }
+  }
+  dependencies {
+    classpath deps.bnd
```
JakeWharton, line 8: Please add this to the root `build.gradle`. The buildscript classpath is global and it's much easier to deal with when there is only a single place where dependencies are added.

swimmesberger: For some reason gradle does not allow to access buildscript classes directly in a gradle file that is externally applied. Therefore as a workaround I have added the class as a property to the root project to make it accessible from the jvm.gradle file.

swimmesberger: Applying plugins in applied scripts works as intended but I don't want to apply the bnd plugin because on default the bnd plugin overwrites the jar plugin to provide the updated MANIFEST.MF and that could be counted as "intrusive" again.

swankjesse:
```
Manifest-Version: 1.0
Automatic-Module-Name: okio
Bnd-LastModified: 1601068645975
Bundle-ManifestVersion: 2
Bundle-Name: com.squareup.okio
Bundle-SymbolicName: com.squareup.okio
Bundle-Version: 2.9.0.SNAPSHOT
Created-By: 11.0.7 (AdoptOpenJDK)
Export-Package: okio;uses:="kotlin,kotlin.collections,kotlin.jvm,kotli
 n.jvm.functions,kotlin.jvm.internal";version="2.9.0"
Import-Package: javax.crypto,javax.crypto.spec,kotlin,kotlin.collectio
 ns,kotlin.jvm,kotlin.jvm.functions,kotlin.jvm.internal,kotlin.text
Private-Package: okio.internal
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.6))"
Tool: Bnd-5.1.2.202007211702
```

swankjesse (approved):
Works for me. Thanks for doing this.

okio/jvm/jvm.gradle
```diff
@@ -5,9 +5,39 @@ kotlin.targets.matching { it.platformType.name == 'jvm' }.all { target ->
...
+    from sourceSets.main.output
+    destinationDir = file("$buildDir/bundle")
+    bnd = '''
+    Export-Package: okio
+    Automatic-Module-Name: okio
+    Bundle-SymbolicName: com.squareup.okio
+    '''
+  }
+  
+  // extract only the MANIFEST.MF file from the OSGi compatible JAR file
```
swankjesse, line 19: Does this part do anything meaningful, or is it just because with the old plugin things went sideways here? If this is just ceremony to make me feel better, I think we should remove it.

swankjesse (approved):

okio/jvm/jvm.gradle
swimmesberger: @swankjesse to be honest it's ceremony to make you feel better. I wanted to reduce the surface area the bnd plugins is working on (if you do not trust what the plugin does) therefore my idea was to let the plugin generate the jar file and simply use only the META-INF/MANIFEST.MF file for the jar file generated by the jvmJar task. In practice for the bnd instructions used at the moment the resulting jar file of the bnd plugin will be exactly the same as the resulting jar from the jvmJar task except that the MANIFEST.MF contains the proper OSGi metadata.

swimmesberger:
@swankjesse I have changed the code to be simpler but potentially more intrusive.

okio/jvm/jvm.gradle
```diff
@@ -4,10 +4,20 @@ apply plugin: 'ru.vyarus.animalsniffer'
...
   target.project.sourceCompatibility = JavaVersion.VERSION_1_7
   target.project.targetCompatibility = JavaVersion.VERSION_1_7
-
-  tasks['jar'].configure {
-    manifest {
-      attributes('Automatic-Module-Name': 'okio')
+  
+  // configure the kotlin JVM jar task
+  tasks['jvmJar'].configure { t ->
+    // the bnd task conention modifies this jar task accordingly
```
swankjesse, line 10: not: conention

okio/jvm/jvm.gradle
```diff
@@ -4,10 +4,20 @@ apply plugin: 'ru.vyarus.animalsniffer'
 kotlin.targets.matching { it.platformType.name == 'jvm' }.all { target ->
   target.project.sourceCompatibility = JavaVersion.VERSION_1_7
   target.project.targetCompatibility = JavaVersion.VERSION_1_7
-
-  tasks['jar'].configure {
-    manifest {
-      attributes('Automatic-Module-Name': 'okio')
+  
+  // configure the kotlin JVM jar task
```
swankjesse, line 8: not: delete comment

okio/jvm/jvm.gradle
```diff
@@ -4,10 +4,20 @@ apply plugin: 'ru.vyarus.animalsniffer'
...
+    // the bnd task conention modifies this jar task accordingly
+    def bndConvention = bndBundleTaskConventionClass.newInstance(t);
+    // modify these lines for MANIFEST.MF properties or for specific bnd instructions
+    bndConvention.bnd = '''
+    Export-Package: okio
+    Automatic-Module-Name: okio
+    Bundle-SymbolicName: com.squareup.okio
+    '''
+    // call the convention when the task has finished to modify the jar to contain OSGi metadata
+    t.doLast {
```
swankjesse, line 19: One more question: what's different between this approach and typical use of bnd? The way this stuff is hooked up seems pretty atypical! doLast and bndBundleTaskConventionClass both seem awkward.

swankjesse (commented):
If we're gonna support OSGi, I'd like to do it in the least creative, least invasive way possible.

okio/jvm/jvm.gradle
swimmesberger: Typically you would apply the plugin itself which will overwrite the jar task and read a file in the projects directory called bnd.bnd where the multiline string should be in. With the build of okio this is not possible because the jar task itself doesn't really do anything the real "jar" task is the "jvmJar" therefore I applied the stuff the plugin does normally to the jar task to the jvmJar task instead.

swankjesse: Thanks for the explanation

swankjesse (approved):

PR #778, opened 2020-09-28
Fix SegmentedByteString substring write

bnorm:
Only occurs when performing a partial write of a SegmentedByteString into a Buffer. Partial write of a ByteString was added in 2.4.1 by #651 (yep, my bad 😬 ).


swankjesse (approved):

swankjesse:
Confirmed locally ’cause the multiplatform build was flaking on CI.

swankjesse:
Thanks @bnorm!

PR #779, opened 2020-10-04
Don't use hash function names in non-JVM platforms

swankjesse:
Also replace HashFunctionTest.kt with MessageDigestConsistencyTest.kt, which
exercises many more access patterns.

This creates a directory of .kt files, okio/src/hashFunctions/kotlin/ that
is included by both nonJvmMain and jvmTest. I would have preferred to create
a sourceSet for hashFunctions, but that was causing Kotlin Multiplatform
builds to fail.

This doesn't restore the hash functions to the benchmark, though that should
be possible; either by depending on the test artifact or by creating a JVM
only artifact that is depended-upon by both test and JMH.


JakeWharton:
I think this wants to join forces with #783 so that `HashFunction` is erased to an alias of `MessageDigest` on the JVM.

JakeWharton:
Actually maybe you've accomplished that since the interface only lives in hashFunctions/ now.

JakeWharton (approved):

PR #780, opened 2020-10-04
Make CipherSink and CipherSource into public types

swankjesse:
This is a hedge against needing to add new public APIs on these
types. We needed it for InflaterSource so we could have a method
that would consume bytes without necessarily returning any.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020 Square, Inc.
+ * Copyright (C) 2020 Square, Inc. and others.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,25 +13,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-@file:JvmMultifileClass
-@file:JvmName("Okio")
-
 package okio
 
 import java.io.IOException
 import javax.crypto.Cipher
 
-private class CipherSink(
+class CipherSink(
   private val sink: BufferedSink,
-  private val cipher: Cipher
+  val cipher: Cipher
 ) : Sink {
-
   private val blockSize = cipher.blockSize
   private var closed = false
 
   init {
-    // Require block cipher, and check for unsupported (too large) block size (should never happen with standard algorithms)
+    // Require block cipher, and check for unsupported (too large) block size (should never happen
+    // with standard algorithms)
     require(blockSize > 0) { "Block cipher required $cipher" }
     require(blockSize <= Segment.SIZE) { "Cipher block size $blockSize too large $cipher" }
   }
@@ -53,7 +49,7 @@ private class CipherSink(
     val size = minOf(remaining, head.limit - head.pos).toInt()
     val buffer = sink.buffer
 
-    // For block cipher, output size cannot exceed input size in update
+    // For block cipher, output size cannot exceed input size in update.
     val s = buffer.writableSegment(size)
 
     val ciphered = cipher.update(head.data, head.pos, size, s.data, s.limit)
@@ -61,8 +57,8 @@ private class CipherSink(
     s.limit += ciphered
     buffer.size += ciphered
 
+    // We allocated a tail segment, but didn't end up needing it. Recycle!
     if (s.pos == s.limit) {
-      // We allocated a tail segment, but didn't end up needing it. Recycle!
       buffer.head = s.pop()
       SegmentPool.recycle(s)
     }
@@ -70,19 +66,17 @@ private class CipherSink(
     // Mark those bytes as read.
     source.size -= size
     head.pos += size
-
     if (head.pos == head.limit) {
       source.head = head.pop()
       SegmentPool.recycle(head)
     }
+
     return size
   }
 
-  override fun flush() =
-    sink.flush()
+  override fun flush() = sink.flush()
 
-  override fun timeout() =
-    sink.timeout()
+  override fun timeout() = sink.timeout()
 
   @Throws(IOException::class)
   override fun close() {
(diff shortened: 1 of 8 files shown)


okio/src/jvmMain/kotlin/okio/JvmOkio.kt
```diff
@@ -189,6 +190,20 @@ fun Path.sink(vararg options: OpenOption): Sink =
 fun Path.source(vararg options: OpenOption): Source =
     Files.newInputStream(this, *options).source()
 
+/**
+ * Returns a sink that processes data using this [Cipher] while writing to [sink].
+ *
+ * @throws IllegalArgumentException if this isn't a block cipher.
+ */
+fun Cipher.sink(sink: BufferedSink): CipherSink = CipherSink(sink, this)
```
bnorm, line 198: Say we have the use case of writing compressed and encrypted data to a socket:

```java
BufferedSink sink = Okio.buffer(new GzipSink(new CipherSink(Okio.buffer(Okio.sink(socket)), cipher)));
// or
BufferedSink sink = Okio.buffer(new GzipSink(Okio.sink(cipher, Okio.buffer(Okio.sink(socket)))));
```

```kotlin
val sink = cipher.sink(socket.sink().buffer()).gzip().buffer()
```

**My 2 cents:**

I would submit the Java examples are easier to read than the Kotlin example which mixes fluent and non-fluent style functions. Consider the following Kotlin alternative.

```kotlin
val sink = socket.sink().cipher(cipher).gzip().buffer()
```

I think this is easier to read as it only uses fluent style functions. I also think this leads to easier discoverability as the `cipher` function is an extension of a Sink. But again, my 2 cents.

swankjesse: Sure. What should we change?

MartinDevi: I also tend to prefer the readability of the alternative (which was I had initially suggested actually) where the extension method is on the sink/source rather than on the cipher instance. 

```kotlin
fun BufferedSink.cipher(cipher: Cipher) CipherSink = CipherSink(this, cipher)
```

In terms of discoverability, I do see why you could argue that having the extension methods on the `Cipher` class is better. It's tied to the same domain, and avoids potentially polluting the suggestions when looking through the list of methods on sink/source. 

I guess I'd prefer having code that's easy to read than that's easy to write 😄

swankjesse: Ahhh! I tried by swapping parametre & receiver, creating these methods:

```
fun BufferedSink.cipher(cipher: Cipher): CipherSink
fun BufferedSource.cipher(cipher: Cipher): CipherSource
```

Unfortunately with this, calling `Buffer.cipher()` is ambiguous because both methods match.

MartinDevi: Oh yeah I hadn't considered that, it would have to be something like `cipherSink()`/`cipherSource()`, which is admittedly a bit verbose...

swankjesse:
Updated the PR with cipherSink() / cipherSource. It reads okay.

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
bnorm: The `gzip()` extension suffers from the same plight; you have to cast a Buffer before you can apply it. I've not found it greatly annoying in practice as I don't encounter it very often, but it is a rough edge to the gzip API, and `cipherSink()` and `cipherSource()` is an okay work-around for the cipher API.

I wish you can check the opmode of a Cipher after it has been initialized. Then you could consider the extension functions `Sink.enrypt()` and `Source.decrypt()` (similar to `inflate()`/`deflate()`).

bnorm (approved):

okio/src/jvmMain/kotlin/okio/CipherSink.kt
```diff
@@ -127,13 +121,3 @@ private class CipherSink(
...
 }
-
-/**
- * Returns a [Sink] that processes data using this [Cipher] while writing to
- * [sink].
- *
- * @throws IllegalArgumentException
- *  If this isn't a block cipher.
- */
-fun Cipher.sink(sink: BufferedSink): Sink =
```
JakeWharton, line 138: This API seems backwards anyway. You don't write `UTF_8.toBytes("hello")` you write `"hello".toBytes(UTF_8)`.

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
```diff
@@ -189,6 +190,20 @@ fun Path.sink(vararg options: OpenOption): Sink =
 fun Path.source(vararg options: OpenOption): Source =
     Files.newInputStream(this, *options).source()
 
+/**
+ * Returns a sink that uses [cipher] to encrypt or decrypt [this].
+ *
+ * @throws IllegalArgumentException if [cipher] isn't a block cipher.
+ */
+fun Sink.cipherSink(cipher: Cipher): CipherSink = CipherSink(this.buffer(), cipher)
```
JakeWharton, line 198: 👍 yeah this is what I'm talking about

okio/src/jvmTest/kotlin/okio/CipherAlgorithm.kt
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import javax.crypto.spec.IvParameterSpec
+import javax.crypto.spec.SecretKeySpec
+import kotlin.random.Random
+
+data class CipherAlgorithm(
```
JakeWharton, line 22: `data` modifier is super dangerous for public API. this type will be extremely hard to compatibly evolve. so be sure.

swankjesse: Yep. Test only!

PR #784, opened 2020-10-09
Extensions for hashing source & sink

MartinDevi:
Hello 👋 me again with more crypto stuff 😅

This MR adds similar extensions to `cipherSource` and `cipherSink` for the hashing source and sink. 

More importantly though, it exposes the `Mac` and `MessageDigest` classes directly to be able to create the hashing source or sink directly with a pre-initialized instance, rather than using one of the predefined ones. This allows more extensibility to use with other algorithms that the platform may support, but it's also more flexible to perform operations on the mac or digest instances themselves. In a way, it's also consistent with the new API for using cipher source and sink. 

My use-case is a cipher spec where the `Mac` instance needs to be initialized with a an initialization vector, before being computed for all the data. This isn't possible with the current API. Granted, it's a bit of a unorthodox spec, but I'm sure there are other cases where this kind of flexibility could be needed.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/HashingSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/HashingSink.kt
@@ -41,9 +41,19 @@ class HashingSink : ForwardingSink {
   private val messageDigest: MessageDigest?
   private val mac: Mac?
 
+  internal constructor(sink: Sink, digest: MessageDigest) : super(sink) {
+    this.messageDigest = digest
+    this.mac = null
+  }
+
   internal constructor(sink: Sink, algorithm: String) : super(sink) {
     this.messageDigest = MessageDigest.getInstance(algorithm)
     this.mac = null
+  }
+
+  internal constructor(sink: Sink, mac: Mac) : super(sink) {
+    this.mac = mac
+    this.messageDigest = null
   }
 
   internal constructor(sink: Sink, key: ByteString, algorithm: String) : super(sink) {
--- a/okio/src/jvmMain/kotlin/okio/HashingSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/HashingSink.kt
@@ -46,26 +46,23 @@ class HashingSink : ForwardingSink {
     this.mac = null
   }
 
-  internal constructor(sink: Sink, algorithm: String) : super(sink) {
-    this.messageDigest = MessageDigest.getInstance(algorithm)
-    this.mac = null
-  }
+  internal constructor(sink: Sink, algorithm: String) : this(sink, MessageDigest.getInstance(algorithm))
 
   internal constructor(sink: Sink, mac: Mac) : super(sink) {
     this.mac = mac
     this.messageDigest = null
   }
 
-  internal constructor(sink: Sink, key: ByteString, algorithm: String) : super(sink) {
+  internal constructor(sink: Sink, key: ByteString, algorithm: String) : this(
+    sink,
     try {
-      this.mac = Mac.getInstance(algorithm).apply {
+      Mac.getInstance(algorithm).apply {
         init(SecretKeySpec(key.toByteArray(), algorithm))
       }
-      this.messageDigest = null
     } catch (e: InvalidKeyException) {
       throw IllegalArgumentException(e)
     }
-  }
+  )
 
   @Throws(IOException::class)
   override fun write(source: Buffer, byteCount: Long) {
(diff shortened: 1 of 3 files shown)


okio/src/jvmMain/kotlin/okio/HashingSink.kt
```diff
@@ -41,11 +41,21 @@ class HashingSink : ForwardingSink {
   private val messageDigest: MessageDigest?
   private val mac: Mac?
 
+  internal constructor(sink: Sink, digest: MessageDigest) : super(sink) {
+    this.messageDigest = digest
+    this.mac = null
+  }
+
   internal constructor(sink: Sink, algorithm: String) : super(sink) {
     this.messageDigest = MessageDigest.getInstance(algorithm)
```
swankjesse, line 50: change this constructor to call through the new one?

okio/src/jvmMain/kotlin/okio/HashingSink.kt
```diff
@@ -41,11 +41,21 @@ class HashingSink : ForwardingSink {
...
     this.mac = null
   }
 
+  internal constructor(sink: Sink, mac: Mac) : super(sink) {
+    this.mac = mac
+    this.messageDigest = null
+  }
+
   internal constructor(sink: Sink, key: ByteString, algorithm: String) : super(sink) {
```
swankjesse, line 59: Change this constructor to call through the new one?

okio/src/jvmMain/kotlin/okio/HashingSource.kt
```diff
@@ -42,11 +42,21 @@ class HashingSource : ForwardingSource {
   private val messageDigest: MessageDigest?
   private val mac: Mac?
 
+  internal constructor(source: Source, digest: MessageDigest) : super(source) {
+    this.messageDigest = digest
+    this.mac = null
+  }
+
   internal constructor(source: Source, algorithm: String) : super(source) {
```
swankjesse, line 50: same

okio/src/jvmMain/kotlin/okio/HashingSource.kt
```diff
@@ -42,11 +42,21 @@ class HashingSource : ForwardingSource {
...
     messageDigest = MessageDigest.getInstance(algorithm)
     mac = null
   }
 
+  internal constructor(source: Source, mac: Mac) : super(source) {
+    this.mac = mac
+    this.messageDigest = null
+  }
+
   internal constructor(source: Source, key: ByteString, algorithm: String) : super(source) {
```
swankjesse, line 60: same

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
```diff
@@ -204,6 +206,26 @@ fun Sink.cipherSink(cipher: Cipher): CipherSink = CipherSink(this.buffer(), ciph
...
+
+/**
+ * Returns a source that uses [mac] to hash [this].
+ */
+fun Source.macSource(mac: Mac): HashingSource = HashingSource(this, mac)
+
+/**
+ * Returns a sink that uses [digest] to hash [this].
+ */
+fun Sink.digestSink(digest: MessageDigest): HashingSink = HashingSink(this, digest)
```
swankjesse, line 222: rename to `hashingSink` ? I don’t think `digest` is as guessable as hashingSink for autocomplete!

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
```diff
@@ -204,6 +206,26 @@ fun Sink.cipherSink(cipher: Cipher): CipherSink = CipherSink(this.buffer(), ciph
...
 
+/**
+ * Returns a sink that uses [mac] to hash [this].
+ */
+fun Sink.macSink(mac: Mac): HashingSink = HashingSink(this, mac)
+
+/**
+ * Returns a source that uses [mac] to hash [this].
+ */
+fun Source.macSource(mac: Mac): HashingSource = HashingSource(this, mac)
```
swankjesse, line 217: rename to `hashingSink` ? I don’t think `macSink` is as guessable as hashingSink for autocomplete!

swankjesse (commented):
Works for me.

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
MartinDevi: Good point, 100% agree.

okio/src/jvmMain/kotlin/okio/HashingSink.kt
MartinDevi: Yup, you're right, makes sense

MartinDevi:
Could this be merged, so that I can use it from the snapshot build ?

swankjesse (approved):

MartinDevi:
Thanks!

PR #785, opened 2020-10-13
Fix README code snippets formatting

gasches:


Egorand:
The file definitely looks broken right now - thanks for fixing! - however the fix breaks the tabbed formatting for Java vs Kotlin samples in MkDocs (see https://square.github.io/okio/#read-a-text-file-line-by-line-javakotlin). I believe the best solution would be to move the full version of README into the docs folder and have the top-level `README.md` link to https://square.github.io/okio, similar to what [KotlinPoet](https://github.com/square/kotlinpoet) does.

Egorand:
Solved with #799, so I'm gonna close this PR. Thanks for contributing though!

PR #787, opened 2020-10-14
Path API for multiplatform

swankjesse:
This is a bit more experimental than I would have preferred.
Like Java's File and Path types, this uses a single type to
represent either a relative or absolute path. It attempts to
simplify '..' and '.' operators and remove unnecessary empty
paths.

### DIFF
--- a/okio-files/src/commonMain/kotlin/okio/Filesystem.kt
+++ b/okio-files/src/commonMain/kotlin/okio/Filesystem.kt
@@ -1,0 +1,25 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+abstract class Filesystem {
+  /**
+   * The current process's working directory. This is the result of the `getcwd` command on POSIX
+   * and the `user.dir` System property in Java. This is an absolute path that all relative paths
+   * are resolved against when using this filesystem.
+   */
+  abstract val cwd: Path
+}
--- a/okio-files/src/commonMain/kotlin/okio/Path.kt
+++ b/okio-files/src/commonMain/kotlin/okio/Path.kt
@@ -1,0 +1,152 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import okio.ByteString.Companion.EMPTY
+import okio.ByteString.Companion.encodeUtf8
+
+/**
+ * A hierarchical address on a filesystem. A path is an identifier only; a [Filesystem] is required
+ * to access the file that a path refers to, if any.
+ *
+ * Paths may be absolute or relative:
+ *
+ * * **Absolute paths** are prefixed with `/` and identify a location independent of any working
+ *   directory.
+ * * **Relative paths** are not prefixed with `/`. On their own, relative paths do not identify a
+ *   location on a filesystem; they must be resolved against an absolute path first. When a relative
+ *   path is used to access a [Filesystem], it is resolved against [Filesystem.cwd].
+ *
+ * After the optional leading `/`, the rest of the path is a sequence of segments separated by `/`
+ * characters. Segments satisfy these rules:
+ *
+ *  * Segments are always non-empty.
+ *  * Segments may not be `.`.
+ *  * Segments with the value `..` only occur in relative paths, where these must precede all other
+ *    segments.
+ *
+ * The only path that ends with `/` is the filesystem root, `/`. The empty path `""` is a relative
+ * path that resolves to whichever path it is resolved against.
+ *
+ * Sample Paths
+ * ------------
+ *
+ * | Path                   | Type       | Parent              |
+ * | :--------------------- | :--------- | :------------------ |
+ * | `/Users/jessewilson`   | Absolute   | `/Users`            |
+ * | `/Users`               | Absolute   | `/`                 |
+ * | `/`                    | Absolute   | null                |
+ * | `src/main/java`        | Relative   | `src/main`          |
+ * | `src/main`             | Relative   | `src`               |
+ * | `src`                  | Relative   | (empty)             |
+ * | (empty)                | Relative   | null                |
+ * | `../../src/main/java`  | Relative   | ` ../../src/main`   |
+ * | `../../src/main`       | Relative   | ` ../../src`        |
+ * | `../../src`            | Relative   | ` ../..`            |
+ * | `../..`                | Relative   | null                |
+ */
+class Path private constructor(
+  private val bytes: ByteString
+) {
+  val isAbsolute: Boolean
+    get() = bytes.startsWith(SLASH)
+
+  val isRelative: Boolean
+    get() = !isAbsolute
+
+  /**
+   * Returns the path immediately enclosing this path. This returns null if this is either the
+   * filesystem root (`/`), the empty path, or a traversal up from the empty path.
+   */
+  val parent: Path?
+    get() {
+      if (bytes == EMPTY || bytes == SLASH || bytes.endsWith(SLASH_DOT_DOT) || bytes == DOT_DOT) {
+        return null // Terminal path.
+      }
+      return when (val lastSlash = bytes.lastIndexOf(SLASH)) {
+        -1 -> Path(EMPTY) // Parent is the current working directory.
+        0 -> Path(bytes.substring(endIndex = 1)) // Parent is the filesystem root '/'.
+        else -> Path(bytes.substring(endIndex = lastSlash))
+      }
+    }
+
+  /** Returns a path that resolves [child] relative to this path. */
+  operator fun div(child: String): Path {
+    val buffer = Buffer()
+    buffer.write(bytes)
+    if (buffer.size > 0) {
+      buffer.write(SLASH)
+    }
+    buffer.writeUtf8(child)
+    return buffer.toPath()
+  }
+
+  override fun equals(other: Any?) = other is Path && other.bytes == bytes
+
+  override fun hashCode() = bytes.hashCode()
+
+  override fun toString() = bytes.utf8()
+
+  companion object {
+    private val SLASH = "/".encodeUtf8()
+    private val DOT = ".".encodeUtf8()
+    private val DOT_DOT = "..".encodeUtf8()
+    private val SLASH_DOT_DOT = "/..".encodeUtf8()
+
+    fun String.toPath(): Path = Buffer().writeUtf8(this).toPath()
+
+    private fun Buffer.toPath(): Path {
+      val absolute = !exhausted() && get(0) == '/'.toByte()
+      if (absolute) {
+        readByte()
+      }
+
+      val canonicalParts = mutableListOf<ByteString>()
+      while (!exhausted()) {
+        val limit = indexOf(SLASH)
+
+        val part: ByteString
+        if (limit == -1L) {
+          part = readByteString()
+        } else {
+          part = readByteString(limit)
+          readByte()
+        }
+
+        if (part == DOT_DOT) {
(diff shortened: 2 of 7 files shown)


swankjesse:
This is very similar to creating a `java.io.File()` and calling `normalize()` on it:

 * This PR says the parent of relative path `../..` is null; java.io.File says it’s `..`.
 * This PR says the parent of relative path `"src"` is `""`, java.io.File says it’s null.
 * `java.io.File("", "home")` turns two relative paths into an absolute path, `/home`.
 * `java.io.File("/../../")` retains all the dot-dots; this PR discards leading dot-dots on absolute paths.

swankjesse:
And the differences with a normalized `java.nio.file.Path` are just the first two from above. Java’s path and this PR agree on the 3rd and 4th bullet.

okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
```diff
@@ -0,0 +1,177 @@
...
+    assertNull("../..".toPath().parent)
+  }
+
+  @Test
+  fun `absolute path traversal with div operator`() {
+    val root = "/".toPath()
+    assertEquals("/home".toPath(), root / "home")
+    assertEquals("/home/jesse".toPath(), root / "home" / "jesse")
+    assertEquals("/home".toPath(), root / "home" / "jesse" / "..")
+    assertEquals("/home/jake".toPath(), root / "home" / "jesse" / ".." / "jake")
```
swankjesse, line 70: do I go to programming prison for this?

cketti: yes :wink:

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+abstract class Filesystem {
```
yschimke, line 18: Is a filesytem strictly the view you would have outside of the process? e.g. a POSIX filesystem?  Is it assumed to be a singleton? Or is there some idea of virtual filesystems, mocking for testing, or federated views within an application?

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,145 @@
...
+ * | `/Users`               | Absolute   | `/`                 |
+ * | `/`                    | Absolute   | null                |
+ * | `src/main/java`        | Relative   | `src/main`          |
+ * | `src/main`             | Relative   | `src`               |
+ * | `src`                  | Relative   | (empty)             |
+ * | (empty)                | Relative   | null                |
+ * | `../../src/main/java`  | Relative   | ` ../../src/main`   |
+ * | `../../src/main`       | Relative   | ` ../../src`        |
+ * | `../../src`            | Relative   | ` ../..`            |
+ * | `../..`                | Relative   | null                |
```
yschimke, line 52: Why is parent not relative to cwd? Because of the when accessed? This implies that a relative path is a persistent concept that can be reapplied to various cwd values with different results, correct?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -0,0 +1,25 @@
...
+  /**
+   * The current process' working directory. This is the result of the `getcwd` command on POSIX
+   * and the `user.dir` System property in Java. This is an absolute path that all relative paths
+   * are resolved against for by this filesystem.
+   */
+  abstract val cwd: Path
```
yschimke, line 24: Does this reflect changes via chdir?

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,145 @@
...
+ */
+class Path private constructor(
+  private val bytes: ByteString
```
yschimke, line 55: Any rules around double slash in path or trailing slash?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+abstract class Filesystem {
+  /**
+   * The current process' working directory. This is the result of the `getcwd` command on POSIX
+   * and the `user.dir` System property in Java. This is an absolute path that all relative paths
```
yschimke, line 21: You reference java filesystem rules elsewhere.  Are you willing to specify a reference implementation e.g. posix when things don't agree?  How will you decide default behaviour?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: I’m thinking similar to OkHttp’s `Dns.SYSTEM`. There’s an obvious one, but if you wanna use DI, it’s very easy.

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: A relative path can be resolved against an absolute path, or perhaps another relative path!

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: Ooooh I think it should. I don’t do anything off JVM so that’s stuff I’ll need to learn.

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: I wanted to be explicit that all path segments were non-empty, and do not have a trailing slash. Lemme make sure that’s called out.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: Lemme call out POSIX as source of truth. I have to add Windows things in a follow-up, which is when things will get trickier.

okio-files/src/commonMain/kotlin/okio/Path.kt
yschimke, line 55: What is the encoding of path? US-ASCII? Single byte os dependant? potentially multibyte?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
yschimke: I like this.  But means you aren't trying to deal with federation within this API. Java file systems are weird when jar ends up in a File, nothing works but eh abstraction of File is carrying data.

java.io.FileNotFoundException: file:/path/to/jarfile/bot.jar!/config/netclient.p (No such file or directory)

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,145 @@
...
+        0 -> Path(bytes.substring(endIndex = 1)) // Parent is the filesystem root '/'.
+        else -> Path(bytes.substring(endIndex = lastSlash))
+      }
+    }
+
+  /** Returns a path that resolves [child] relative to this path. */
+  operator fun div(child: String): Path {
+    val buffer = Buffer()
+    buffer.write(bytes)
+    if (buffer.size > 0) {
```
yschimke, line 83: child == "" seems like a bug in the handling here.  trailing slash.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -0,0 +1,25 @@
...
+   * are resolved against for by this filesystem.
```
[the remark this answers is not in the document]
swankjesse: Fixed.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+abstract class Filesystem {
+  /**
+   * The current process' working directory. This is the result of the `getcwd` command on POSIX
```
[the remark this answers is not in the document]
swankjesse: Fixed. Praise Jesus’s good grammar.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: No, federation is too difficult. Instead I think of `FileSystem` as being almost like a `Collection` type, like `Map`. You can interrogate and mutate state using `Path` as keys.

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: Added docs.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: One thing that’s weird is I could make this a `var` and call `chdir` when you set it!

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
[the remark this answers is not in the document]
Egorand: Or is it Jesus's disciples' grammar?

okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
Egorand: 🚔 🚔 🚔

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,152 @@
...
+ * ------------
+ *
+ * | Path                   | Type       | Parent              |
+ * | :--------------------- | :--------- | :------------------ |
+ * | `/Users/jessewilson`   | Absolute   | `/Users`            |
+ * | `/Users`               | Absolute   | `/`                 |
+ * | `/`                    | Absolute   | null                |
+ * | `src/main/java`        | Relative   | `src/main`          |
+ * | `src/main`             | Relative   | `src`               |
+ * | `src`                  | Relative   | (empty)             |
```
Egorand, line 54: Why is the parent of a relative path empty and not null?

Egorand (approved):

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: Probably best is to pattern match.

```
mkdir -p ~/xxx/yyy/zzz

cd ~/
cd "xxx/yyy/zzz"
pwd
// prints ~/xxx/yyy/zzz

cd ~/
cd "xxx/yyy"
pwd
// prints ~/xxx/yyy

cd ~/
cd "xxx"
pwd
// prints ~/xxx

cd ~/
cd ""
pwd
// prints ~/
```

The empty relative resolves to whatever it was resolved against.

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+  /** Returns a path that resolves [child] relative to this path. */
+  operator fun div(child: String): Path {
+    val buffer = Buffer()
+    buffer.write(bytes)
+    if (buffer.size > 0) {
+      buffer.write(SLASH)
+    }
+    buffer.writeUtf8(child)
```
JakeWharton, line 93: How do I resolve children that are not valid UTF-8?

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,152 @@
...
+
+  override fun toString() = bytes.utf8()
+
+  companion object {
+    private val SLASH = "/".encodeUtf8()
+    private val DOT = ".".encodeUtf8()
+    private val DOT_DOT = "..".encodeUtf8()
+    private val SLASH_DOT_DOT = "/..".encodeUtf8()
+
+    fun String.toPath(): Path = Buffer().writeUtf8(this).toPath()
```
JakeWharton, line 109: How do I create a path that is not valid UTF-8?

okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
eygraber: Just happened to stumble across this. I've done this in several projects that run on both Linux and Windows. It's amazing how humans have the ability to both love and detest the same thing at the same time :smile:

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: The internal representation as a `ByteString` is my acknowledging that paths aren’t strings. Gotta add more APIs to support byte strings in more places!

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: API forthcoming. I wanna cover as much happy path as I can, especially into native code, before handling the long tail of gotchas with these things.

okio-files/src/commonMain/kotlin/okio/Path.kt
JakeWharton: Yeah i was happy to see that. If you haven't, https://fasterthanli.me/articles/i-want-off-mr-golangs-wild-ride is a fun read.

PR #791, opened 2020-11-01
JVM implementation of Filesystem

swankjesse:


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -22,4 +22,12 @@ abstract class Filesystem {
    * are resolved against when using this filesystem.
    */
   abstract val cwd: Path
+
+  companion object {
+    /**
+     * The current process's host filesystem. Use this instance directly, or dependency inject a
+     * [Filesystem] to make code testable.
+     */
+    val SYSTEM: Filesystem = PLATFORM_FILESYSTEM
```
Egorand, line 31: Not a huge fan of `Filesystem.SYSTEM` at a potential call site (feels repetitive), but can't think of a really good alternative. Referencing the platform specific file system through just `Filesystem` would be nice, but not sure how to achieve it. I wanna look into the coroutines lib for inspiration on modeling platform-specific singletons (e.g. `Dispatchers.Main`) and come back with better suggestions!

okio-files/src/commonMain/kotlin/okio/Platform.kt
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+internal expect val PLATFORM_FILESYSTEM: Filesystem
```
Egorand, line 18: Hm, I wonder if `public expect val Filesystem: Filesystem` will compile?

okio-files/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
```diff
@@ -0,0 +1,26 @@
...
+ * limitations under the License.
+ */
+package okio
+
+import okio.Path.Companion.toPath
+
+object JvmSystemFilesystem : Filesystem() {
+  override val cwd: Path
+    get() {
+      val userDir = System.getProperty("user.dir") ?: error("user.dir system property missing?!")
```
Egorand, line 23: [Coding conventions](https://kotlinlang.org/docs/reference/coding-conventions.html#functions-vs-properties) state that:

- Properties shouldn't throw
- Properties should be expected to return the same result if the object state hasn't changed

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
JakeWharton: You're supposed to call through the abstraction. Referencing an implementation at a callsite is supposed to look bad because it's an anti-pattern.

Egorand: Makes sense

okio-files/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
swankjesse: The throwing case is extremely unlikely;  it only happens if the System properties are edited to remove this built-in property. But I wanted it to crash with a nice message rather than an NullPointerException in that rare case.

swankjesse: As for the changing value, again it's unlikely to change. But if it does it’s because something in the same process changed it. Think of this property as having a `private set` that manipulate it.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: I'm pattern matching `Dns.SYSTEM` from OkHttp which seemed like a good name for something provided by the OS.

JakeWharton: Also TimeSouce.SYSTEM from kotlinx.datetime!

Egorand: I'm sold!

Egorand (approved):

PR #792, opened 2020-11-01
New API, Path.name

swankjesse:


Egorand (approved):

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -67,6 +71,18 @@ class Path private constructor(
...
+  val nameBytes: ByteString
+    get() {
+      val lastSlash = bytes.lastIndexOf(SLASH)
+      return when {
+        lastSlash != -1 -> bytes.substring(lastSlash + 1)
+        else -> bytes
+      }
+    }
+
+  val name: String
```
JakeWharton, line 83: Needs docs and tests for behavior for invalid UTF-8 paths. Seems like it will return an invalid value as written, but it probably should throw.

JakeWharton (approved):

okio-files/src/commonMain/kotlin/okio/Path.kt
swankjesse: Agreed. I'm gonna need a whole bunch of PRs to test filenames with other charsets, segment delimiters, etc.

But I'd like to go wide and get something that's useful for ASCII before diving into the details. Otherwise I fear I'll be paralyzed by the underlying difficulty of the long tail of challenges here!

PR #793, opened 2020-11-01
Initial native implementation of Filesystem

swankjesse:
I'm working in IntelliJ and it's been happier to track a single
target platform than several. I'll expand this to Linux and iOS
in a follow-up PR.

This satisfies egorand's recommendation of making cwd() a function
rather than a val. Annoyingly the native implementation can fail
in ways far more plausible than the JVM implementation.

### DIFF
--- a/okio-files/src/macosX64Main/kotlin/okio/PosixSystemFilesystem.kt
+++ b/okio-files/src/macosX64Main/kotlin/okio/PosixSystemFilesystem.kt
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


okio-files/src/macosX64Main/kotlin/okio/PosixSystemFilesystem.kt
```diff
@@ -0,0 +1,47 @@
...
+import platform.posix.EACCES
+import platform.posix.ENOENT
+import platform.posix.ENOMEM
+import platform.posix.PATH_MAX
+import platform.posix.errno
+import platform.posix.free
+import platform.posix.getcwd
+
+object PosixSystemFilesystem : Filesystem() {
+  override fun cwd(): Path {
```
Egorand, line 30: This is fun

Egorand (approved):

PR #794, opened 2020-11-02
Filesystem directory listing

swankjesse:


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -28,6 +28,16 @@ abstract class Filesystem {
   @Throws(IOException::class)
   abstract fun cwd(): Path
 
+  /**
+   * Returns the children of the directory identified by [path], or null if [path] is a not a
```
Egorand, line 32: nit: either rename the parameter to `path` or change the doc to `dir`

okio-files/src/jvmMain/kotlin/okio/Jvm.kt
```diff
@@ -0,0 +1,21 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import java.io.File
+
+val Path.file: File
```
Egorand, line 20: nit: `internal`?

Egorand (approved):

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: good catch. Fixed.

okio-files/src/jvmMain/kotlin/okio/Jvm.kt
swankjesse: Fixed. Might promote this to public later.

JakeWharton: It would need to be `fun Path.toFile()` if it were public because this isn't really a property of a `Path`.

okio-files/src/macosX64Main/kotlin/okio/PosixSystemFilesystem.kt
```diff
@@ -36,12 +49,60 @@ object PosixSystemFilesystem : Filesystem() {
...
+    var dirToString = dir.toString()
+
+    // We use "" for cwd; this expects ".".
+    if (dirToString.isEmpty()) dirToString = "."
+
+    val opendir: CPointer<DIR>? = opendir(dirToString)
+
+    if (opendir == null) {
+      when (errno) {
+        EACCES -> throw IOException("EACCES: cannot ls $dir")
```
JakeWharton, line 71:
```suggestion
        EACCES -> throw IOException("EACCES: cannot list $dir")
```

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -28,6 +28,16 @@ abstract class Filesystem {
...
 
+  /**
+   * Returns the children of the directory identified by [dir], or null if [dir] is a not a
+   * directory.
+   *
+   * @throws IOException if [dir] cannot be read, such as if the current process doesn't have
+   *     access to [dir], or if there's a loop of symbolic links, or if any name is too long.
+   */
+  @Throws(IOException::class)
+  abstract fun list(dir: Path): List<Path>?
```
JakeWharton, line 39: Nullable list 😬

okio-files/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
```diff
@@ -23,4 +23,9 @@ object JvmSystemFilesystem : Filesystem() {
       ?: throw IOException("user.dir system property missing?!")
     return userDir.toPath()
   }
+
+  override fun list(dir: Path): List<Path>? {
+    val files = dir.file.list() ?: return null
+    return files.map { dir / it }
```
JakeWharton, line 29: nit: The local here isn't needed. Could be `dir.file.list()?.map { .. }`

JakeWharton (approved):

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: My feeling was it’s more ergonomic than a try/catch if the filesystem changes out from under you. Thoughts?

JakeWharton: Oh I thought null was for when it was a file, but it's also for when it's not available? If the directory disappears I would expect it to have the same behavior as when a file disappears and you ask for its bytes.

Even for the case when it's a file I probably also want it to throw.

swankjesse: I can do that. Perhaps with `listOrNul()` later if it becomes too annoying. (Follow-up ’cause I missed this when I merged.)

swankjesse: https://github.com/square/okio/pull/795

PR #795, opened 2020-11-02
Throw on all Filesystem.list() failures.

swankjesse:


okio-files/src/macosX64Main/kotlin/okio/PosixSystemFilesystem.kt
```diff
@@ -71,8 +71,8 @@ internal object PosixSystemFilesystem : Filesystem() {
         EACCES -> throw IOException("EACCES: cannot list $dir")
         ELOOP -> throw IOException("ELOOP: symbolic link loop resolving $dir")
         ENAMETOOLONG -> throw IOException("ENAMETOOLONG: path name too long in $dir")
-        ENOENT -> return null // A component in dir doesn't exist.
-        ENOTDIR -> return null // `dir` is not a directory.
+        ENOENT -> throw IOException("ENOENT: no such directory $dir")
+        ENOTDIR -> throw IOException("ENOTDIR: not a directory $dir")
```
JakeWharton, line 75: Wow I wonder what R8 does to this control flow! Ideally it does the equivalent of rewriting the function to
```kotlin
throw IOException(when (whevs) {
  ENOSTUFF -> "Stuff"
} + dir)
```

JakeWharton (approved):

PR #796, opened 2020-11-03
Filesystem.source()

swankjesse:


JakeWharton (approved):

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -45,4 +46,26 @@ class FileSystemTest {
     } catch (expected: IOException) {
     }
   }
+
+  @Test
+  fun `file source no such directory`() {
+    try {
```
Egorand, line 52: nit: `assertThrows`?

okio-files/src/macosX64Main/kotlin/okio/FileSource.kt
```diff
@@ -0,0 +1,66 @@
...
+import platform.posix.feof
+import platform.posix.ferror
+import platform.posix.fread
+import platform.posix.free
+
+/** Reads the bytes of a file as a source. */
+internal class FileSource (
+  private val file: CPointer<FILE>
+) : Source {
+  private var segmentSize = 8192L
```
Egorand, line 34: nit: companion object or top-level `const`

Egorand (approved):

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
swankjesse: We don’t have that in Okio?

okio-files/src/macosX64Main/kotlin/okio/FileSource.kt
swankjesse: Fixed!

PR #800, opened 2020-11-03
Filesystem.sink(Path)

swankjesse:


swankjesse:
Weird build failure.

```
> Task :okio-files:assemble
e: /home/runner/work/okio/okio/okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt: (102, 7): Unresolved reference: fail

> Task :okio-files:compileTestKotlinJvm FAILED
```

```
/* 102 */ @Test
/* 103 */ fun `file sink no such directory`() {
/* 104 */   try {
/* 105 */     Filesystem.SYSTEM.sink("/tmp/ce70dc67c24823e695e616145ce38403/unlikely-file".toPath())
/* 106 */     fail()
/* 107 */   } catch (expected: IOException) {
/* 108 */   }
/* 109 */ }
```

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -68,4 +70,50 @@ class FileSystemTest {
...
+
+    buffer.writeUtf8(" world!")
+    sink.write(buffer, buffer.size)
+    sink.close()
+    assertEquals("hello, world!", path.readUtf8())
+  }
+
+  @Test
+  fun `file sink no such directory`() {
+    try {
```
JakeWharton, line 104: assertFailsWith

JakeWharton (approved):

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
swankjesse: Is this in multiplatform? I couldn't find it

JakeWharton: Yep! https://github.com/square/okio/pull/798/

okio-files/src/macosX64Main/kotlin/okio/FileSink.kt
```diff
@@ -0,0 +1,75 @@
...
+import platform.posix.fflush
+import platform.posix.free
+import platform.posix.fwrite
+
+/** Writes bytes to a file as a sink. */
+internal class FileSink(
+  private val file: CPointer<FILE>
+) : Sink {
+  private var nativeBuffer = nativeHeap.allocArray<ByteVar>(segmentSize)
+  private var closed = false
```
Egorand, line 34: General question: should we be concerned about concurrency at this level, i.e. should these be volatile, atomic, or should the methods use locks etc.?

JakeWharton: Sources and Sinks are not thread-safe, like an InputStream or OutputStream. You have to guard things above them.

swankjesse: Fun fact: InputStream and OutputStream use synchronization, but it's the bad ill-advised synchronization of StringBuffer and Vector of 1990s Java. 

Either way, sources and sinks and buffers are not safe for concurrent use.

PR #805, opened 2020-11-05
Filesystem.mkdir()

swankjesse:


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -59,6 +59,16 @@ abstract class Filesystem {
...
 
+  /**
+   * Creates a directory at the path identified by [dir].
+   *
+   * @throws IOException if [dir]'s parent does not exist, is not a directory, or cannot be written.
+   *     A directory cannot be created if the current process doesn't have access, if there's a
+   *     loop of symbolic links, or if any name is too long.
+   */
+  @Throws(IOException::class)
+  abstract fun mkdir(dir: Path)
```
JakeWharton, line 70: Skeptical of retaining this nomenclature. It may be better to go with `createDirectory` and a `createFile` (for touch).

That also leads into a nice `FileSystem.createDirectories` extension function which creates a hierarchy!

JakeWharton (approved):

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: Let's do that. Follow-up!

PR #807, opened 2020-11-07
Filesystem.copy() and delete()

swankjesse:


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -86,6 +86,29 @@ abstract class Filesystem {
...
+   * Copies all of the bytes from the file at [source] to the file at [target]. This does not copy
+   * file metadata like last modified time, permissions, or extended attributes.
+   *
+   * This function is not atomic; a failure may leave [target] in an inconsistent state. For
+   * example, [target] may be empty or contain only a prefix of [source].
+   *
+   * @throws IOException if [source] cannot be read or if [target] cannot be written.
+   */
+  @Throws(IOException::class)
+  abstract fun copy(source: Path, target: Path)
```
JakeWharton, line 99: This may be able to be moved to an extension function whose implementation lives in common code.

It needs an expect/actual for `use`, but the actual implementation being source/sink based means it doesn't need to live directly on a `Path` because it's not a primitive operation (similar to `createDirectories` example I mentioned in other PR).

okio-files/src/macosX64Main/kotlin/okio/cinterop.kt
```diff
@@ -57,6 +57,22 @@ internal fun Buffer.read(
   }
 }
 
+internal inline fun <S : Source, T> S.use(block: (S) -> T): T {
+  try {
+    return block(this)
+  } finally {
+    close()
```
JakeWharton, line 64: Any idea what happens if this throws when `block` also threw?

JakeWharton (approved):

okio-files/src/macosX64Main/kotlin/okio/cinterop.kt
swankjesse: It incorrectly clobbers the thrown exception. Lemme fix up to do the correct thing and throw the first exception...

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: Probably I should just do commonCopy() like we do in Buffer.

JakeWharton: We only do that in Buffer because we have to retain binary compatibility. If we were writing it from scratch, 90% of its methods would be extensions, like this could be.

swankjesse: I want to keep these polymorphic for optionality. We might be able to get the Posix one do an off-CPU copy for example.

PR #808, opened 2020-11-08
Get tests passing under Linux

swankjesse:
This just adjusts some of the POSIX.1 APIs I'm using for macOS to
work on Linux also.

Note that IntelliJ is not very happy to work in the posixMain
directory; it cannot find the libraries it needs.


JakeWharton:
I think the @kpgalligan special is to only enable a single native source set in the IDE. Not necessary for this to land, though.

okio-files/build.gradle
```diff
@@ -32,10 +33,11 @@ kotlin {
         implementation deps.kotlin.test.jdk
       }
     }
-    macosX64Main {
+    posixMain {
```
JakeWharton, line 36: Ah, I see. Yeah with no context this isn't seen as a native source set but another common source set (which you are only pulling into native source sets). I think we may need to switch to HMPP for this to work properly.

JakeWharton (approved):

PR #809, opened 2020-11-08
Mingw64 support

### LINKED ISSUE #743: MingwX64 support ?
What would be needed to get mingwX64 support? adding the target? updating CI and publication? Anything else?

martinbonnin:
Support for windows platform, mostly changes to `Pipe` timeouts to handle the coarse timer resolution on the Windows JVM and some tests relaxed. See #743 for more details (closes #743)
**Edit**: oops, I didn't see the `okio-files` module. I'll need to make this work on windows too, I'll revisit this in the coming days
**Edit2**: the `Pipe` timeouts changes got their own pull request: https://github.com/square/okio/pull/811. So this pull request is mostly about `okio-files` now.


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -109,6 +109,8 @@ abstract class Filesystem {
   @Throws(IOException::class)
   abstract fun delete(path: Path)
 
+  abstract val separator: String
```
swankjesse, line 112: Difficult to get the name here. There's a path separator (`:` or `;`) and a directory separator (`\` or `/`).

I was halfway considering calling the val `slash` ?

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -29,12 +29,16 @@ import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
 import kotlin.test.assertTrue
 
+expect fun tmpDirectory(): String
```
swankjesse, line 32: Should we put this in `Filesystem` ?

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -46,45 +50,42 @@ class FileSystemTest {
...
+      Filesystem.SYSTEM.source("$tmpDirectory/unlikely-directory/ce70dc67c24823e695e616145ce38403".toPath())
     }
   }
 
   @Test
   fun `file source`() {
     val source = Filesystem.SYSTEM.source("gradle.properties".toPath())
     val buffer = Buffer()
-    assertEquals(47L, source.read(buffer, 100L))
+    assertTrue(source.read(buffer, 100L) <= 49L) // either 47 on posix or 49 with \r\n line feeds on windows
```
swankjesse, line 68: For this line feed stuff I'd prefer we do a replaceAll to get rid of them before asserting.

okio/src/jvmMain/kotlin/okio/Pipe.kt
```diff
@@ -56,21 +56,20 @@ class Pipe(internal val maxBufferSize: Long) {
         if (canceled) throw IOException("canceled")
 
         while (byteCount > 0) {
+          timeout.waitUntil(buffer) {
```
swankjesse, line 59: Can you move the Pipe and Timeout changes to a different PR? You can use assumeFalse(windows) if there's tests that fail.

okio/src/jvmTest/java/okio/WaitUntilNotifiedTest.java
```diff
@@ -56,7 +56,6 @@
     double start = now();
     try {
       timeout.waitUntilNotified(this);
-      fail();
     } catch (InterruptedIOException expected) {
```
swankjesse, line 59: I think we might want to increase the tolerance here; I definitely don't want to get rid of this fail() call.

swankjesse (changes requested):
Neat.

Great minds think alike. I just added a native Windows file system.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
martinbonnin: I went with the same naming as Java that has `separator` and `pathSeparator`. The reasoning was that people coming from Java shouldn't be disoriented. Maybe that doesn't work too well for native/js but yea coming up with something that works everywhere is hard.

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
martinbonnin: Done. There's a nice rabbit hole of naming/API designs questions though...
* `tmpDirectory` vs `tempDirectory`
* should it return a `Path` instead of String?
* maybe expose the `prefix` paramater?

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
martinbonnin: Yup, much better and it'll explicit that it's line feed stuff. Should we get rid on the asserts on the number or `read` bytes? I think it's never going to happen but technically, the contract of `read` is that it can return less than the requested number of bytes anyways so the test could be misleading?

okio/src/jvmMain/kotlin/okio/Pipe.kt
martinbonnin: PR for `Pipe` and `Timeout`: https://github.com/square/okio/pull/811/files

> You can use assumeFalse(windows) if there's tests that fail.

👍  Detecting that code is running on windows sounds like another rabbit hole but I'll try to do something. Given that these test are mostly JVM, `System.getProperty("os.name")` might work

okio/src/jvmTest/java/okio/WaitUntilNotifiedTest.java
martinbonnin: I'm not sure what you mean by "tolerance" ? We could keep the `fail()` for non-Windows hosts but on windows the early notification isn't caught by `waitUntilNotified()` so not sure how to test that.

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -101,92 +99,93 @@ class FileSystemTest {
...
+    val target = "$tmpDirectory/FileSystemTest-atomicMove-${randomToken()}".toPath()
     Filesystem.SYSTEM.createDirectory(target)
     assertFailsWith<IOException> {
       Filesystem.SYSTEM.atomicMove(source, target)
     }
   }
 
   @Test
+  @Ignore // somehow the behaviour is different on windows
   fun `atomicMove source is directory and target is file`() {
```
martinbonnin, line 185: There's something with `atomicMove` behaving differently depending the platforms, we might have to write platform dependant tests? Or maybe filesystem dependant tests? 

cf https://github.com/square/okio/pull/810/files#diff-5c23c164810a5a043102f66733c86a74ebcb9935847579369f0adea38ba6f2f2R168 too

okio/src/jvmTest/java/okio/WaitUntilNotifiedTest.java
swankjesse: Let's figure that out in the other PR. For this you can just do assumeNotWindows before the test proceeds so we get a green build.

swankjesse (commented):
Getting close. Thanks for driving this.

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
martinbonnin: Sounds good. Let's keep `<= 49L` then or did you have something else in mind?

swankjesse: Works for me.

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
martinbonnin: Moved to `Path.directorySeparator` using a platform `expect/actual`.

okio-files/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -138,6 +138,8 @@ class Path private constructor(
 
     fun String.toPath(): Path = Buffer().writeUtf8(this).toPath()
 
+    val directorySepartor = PLATFORM_SEPARATOR
```
PromanSEW, line 141: Typo: directorySepar**a**tor

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -109,6 +109,12 @@ abstract class Filesystem {
   @Throws(IOException::class)
   abstract fun delete(path: Path)
 
+  /**
+   * Returns a writable temporary directory on the current file system.
+   * This is the java.io.tmpdir system property on the JVM platform and the TMPDIR environment variable on the POSIX platform
+   */
+  abstract fun temporaryDirectory(): Path
```
JakeWharton, line 116: This doesn't feel like it belongs here since it's not a property intrinsic to a file system. What's the temporary directory for a read-only zip file mounted as a filesystem? A temp dir is a property intrinsic to an operating system or platform.

JakeWharton: Seems like it's only used by tests? Maybe just an actual/expect in there to start with?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -120,6 +120,9 @@ abstract class Filesystem {
      * The current process's host filesystem. Use this instance directly, or dependency inject a
      * [Filesystem] to make code testable.
      */
-    val SYSTEM: Filesystem = PLATFORM_FILESYSTEM
+    val SYSTEM: Filesystem
+    get() {
+      return PLATFORM_FILESYSTEM
+    }
```
martinbonnin: There's something fishy here... Not using the getter syntax triggers a NullPointerException in CI. Haven't been able to reproduce locally

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: I wanna use this to write multiplatform tests that use files. For example, tests of DiskLruCache.

A zip file is not an exemplar file system. For example, it doesn't support random access writing.

swankjesse: (and if you are doing a Zip filesystem, you can just return Filesystem.SYSTEM.temporaryDirectory().

JakeWharton: That argument doesn't sway me. I think you've only radicalized me further. Temporary directories aren't an intrinsic property of a file system, they're a property of an operating system. At best it's a path resolvable by `FileSystem.SYSTEM` but it is not intrinsic to all file systems. It belongs elsewhere.

swankjesse: They are certainly confined to a file system, however. It would be an error to take the temporary directory of my operating system and use it to write to a `.zip` file.

I think what you really want is this:
```
  fun temporaryDirectory(): Path?
```

That way it’s optional, and we can document that `FileSystem.SYSTEM` has one.

JakeWharton: What about a Samba or NFS filesystem mounted over the network? Or exposing S3 as a filesystem either through HTTP calls or the AWS S3 SDK? A zip file just an easy example to litmus test the API against that should quickly allow us to identify what is something primitive to a filesystem and what is not.

In your DiskLruCache example the test should have no behavioral difference whether given the system FS or an in-memory FS. The test doesn't get to ask for a temporary directory, it's given an FS and a base Path. If the test runner (or parameterized factory or whatever) wants to use the system FS and the OS temporary directory it can, but for an in-memory FS the temporary path can be the root dir.

swankjesse: @martinbonnin wanna pull this back from the public API? @JakeWharton and I will figure out the API in a follow-up.

martinbonnin: Sure, let me try

martinbonnin: I just made `temporaryDirectory` internal. I think that's enough to remove it from the public API. Let me know if you want me to bury it a bit deeper

swankjesse: 👍

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -109,11 +109,20 @@ abstract class Filesystem {
   @Throws(IOException::class)
   abstract fun delete(path: Path)
 
+  /**
+   * Returns a writable temporary directory on the current file system.
+   * This is the java.io.tmpdir system property on the JVM platform and the TMPDIR environment variable on the POSIX platform
+   */
+  internal abstract fun temporaryDirectory(): Path
```
swankjesse, line 116: In a follow up I’ll move this elsewhere. Otherwise users can’t extend this class because they can’t implement this method.

swankjesse (approved):
Nice PR!

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
martinbonnin: Makes sense 👍 . Thanks!

PR #811, opened 2020-11-09
Introduce Timeout.waitUntil(monitor, condition: () -> Boolean)

martinbonnin:
Similar to `waitUntilNotified` except it accounts for spurious wake ups and bad timer resolutions. 

Use it in `Pipe` where the bad timer resolution on Windows could make the `Pipe` wait way longer than it should because the `wait` would return a few nanoseconds early very often and the read/write loop would continue looping, waiting several times the nominal Timeout value

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Pipe.kt
+++ b/okio/src/jvmMain/kotlin/okio/Pipe.kt
@@ -56,21 +56,23 @@ class Pipe(internal val maxBufferSize: Long) {
         if (canceled) throw IOException("canceled")
 
         while (byteCount > 0) {
+          timeout.waitUntil(buffer) {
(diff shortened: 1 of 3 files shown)


yschimke:
If we go ahead with https://github.com/square/okio/pull/1176, might be worth waiting on this.

MukjepScarlet:
Well it be better to use `BooleanSupplier` here? It returns primitive boolean.

martinbonnin:
I have lost all context on this and won't be able to dive into this for the foreseeable future. Closing it but please feel free to fork and/or reuse anything in there!

MukjepScarlet:
Pity... (°ﾛ°٥)

PR #812, opened 2020-11-12
Common Hmac HashingSource HashingSink

### LINKED ISSUE #782: Promote HashingSource/HashingSink to common Kotlin
Now that Okio has an internal multiplatform hashing api, Hashing Source/Sink should be available for all compile targets.

MartinDevi:
Closes #782


okio/src/hashFunctions/kotlin/okio/internal/Hmac.kt
```diff
@@ -0,0 +1,75 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio.internal
+
+import okio.ByteString
+import kotlin.experimental.xor
```
JakeWharton, line 19: We can't rely on this in the main source set. I think Jesse wrote an internal version somewhere...

JakeWharton (approved):
I'm tempted to land the test conversion to Kotlin first and then move it to common (unchanged) as part of this PR. Based on the commits it seems like they were (somewhat) done in the opposite order which reduces my confidence that they haven't changed in any meaningful way.

MartinDevi:
> I'm tempted to land the test conversion to Kotlin first and then move it to common (unchanged) as part of this PR. Based on the commits it seems like they were (somewhat) done in the opposite order which reduces my confidence that they haven't changed in any meaningful way.

Full tranparency, I did change the expected hash for the randomly generated bytes, because since the migration went from using `java.util.Random` to `kotlin.Random`, the generated value was different.

okio/src/hashFunctions/kotlin/okio/internal/Hmac.kt
MartinDevi: I found this which seems to be the idea:

https://github.com/square/okhttp/blob/19771365f25d986d125478a75c15d24d6e285e12/okhttp/src/main/kotlin/okhttp3/internal/ws/WebSocketProtocol.kt#L109

Which, from what I can tell, basically comes down to inlining `kotlin.experimental.xor`

```kotlin
public inline infix fun Byte.xor(other: Byte): Byte = (this.toInt() xor other.toInt()).toByte()
```

I guess we want to do that to avoid compatibility changes when those methods leave the experimental stage?

MartinDevi: Replaced with internal copy of method

okio/src/jvmMain/kotlin/okio/HashingSink.kt
```diff
@@ -37,7 +37,7 @@ import javax.crypto.spec.SecretKeySpec
  * ByteString hash = hashingSink.hash();
  * ```
  */
-class HashingSink : ForwardingSink {
+actual class HashingSink : ForwardingSink, Sink { // Need to explicitly declare sink pending fix for https://youtrack.jetbrains.com/issue/KT-20641
```
swankjesse, line 40: delete the docs from these classes? I don’t like having two copies that can get out of sync

okio/src/jvmMain/kotlin/okio/HashingSink.kt
```diff
@@ -93,7 +93,7 @@ class HashingSink : ForwardingSink {
    * internal state is cleared. This starts a new hash with zero bytes accepted.
    */
   @get:JvmName("hash")
-  val hash: ByteString
+  actual val hash: ByteString
```
swankjesse, line 96: Off topic: using a `val` for this was a mistake.

okio/src/hashFunctions/kotlin/okio/internal/Sha256.kt
```diff
@@ -219,6 +221,22 @@ internal class Sha256 : HashFunction {
   }
   /* ktlint-enable */
 
+  private fun reset() {
+    messageLength = 0L
+    unprocessed.fill(0)
+    unprocessedLimit = 0
+    words.fill(0)
+
+    h0 = 1779033703
```
swankjesse, line 230: declare these to all be `0L` up front and call `reset` from `init {}` ?

swankjesse: Actually as-is is fine, it saves a `fill(0)` call in the common case.

okio/src/jvmMain/kotlin/okio/HashingSource.kt
```diff
@@ -38,7 +38,7 @@ import javax.crypto.spec.SecretKeySpec
  * ByteString hash = hashingSource.hash();
  * ```
  */
-class HashingSource : ForwardingSource {
+actual class HashingSource : ForwardingSource, Source { // Need to explicitly declare source pending fix for https://youtrack.jetbrains.com/issue/KT-20641
```
swankjesse, line 41: thanks for the comment

swankjesse (approved):
Nice PR!

PR #813, opened 2020-11-14
Basic Filesystem support for Windows

swankjesse:
Note that tests for clobbered files don't work right.

Also replace Filesystem.baseDirectory() with canonicalize(). You can get the
base directory by canonicalizing `.`. But this function is a better fit for
Windows. On that platform there can be many working directories, one per volume.

    > c:\> c:\
    > c:\> cd C:\Windows
    > c:\Windows> a:\
    > a:\> cd DOOM
    > a:\DOOM> type c:notepad.exe
      ... prints c:\Windows\notepad.exe
    > a:\DOOM> type a:DOOM.wad
      ... prints a:\DOOM\DOOM.wad

With canonicalize() we can get the full path to volume-relative paths
like `c:notepad.exe`.

### DIFF
--- a/okio-files/build.gradle
+++ b/okio-files/build.gradle
@@ -7,6 +7,7 @@ kotlin {
   if (kmpNativeEnabled) {
     macosX64()
     linuxX64()
+    mingwX64()
   }
   sourceSets {
     commonMain {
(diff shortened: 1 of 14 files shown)


okio-files/src/mingwX64Main/kotlin/okio/posix_variant.kt
```diff
@@ -0,0 +1,59 @@
+/*
```
swankjesse, line 1: is this the right name for this file?

JakeWharton: seems fine. i'd probably call it posixVariant.kt. underscores are a bit weird.

JakeWharton (approved):

okio-files/src/mingwX64Main/kotlin/okio/posix_variant.kt
JakeWharton: or variantPosix.kt

Issue #624, opened 2019-06-24
Dead loop in Okio AsyncTimeout caused by a circle in LinkedList


yihuaqi:
We're using com.sqaureup.okio:okio:1.14.0 and com.squareup.okhttp3:okhttp:3.10.0.2.

During developing, we occasionally have the issue that all our network requests are not executing because RealBufferedSink cannot flush. We looked into the issue deeper, and found out that one of our thread is stuck at a dead loop in AsyncTimeout:
```
  private static synchronized void scheduleTimeout(
      AsyncTimeout node, long timeoutNanos, boolean hasDeadline) {
    // Start the watchdog thread and create the head node when the first timeout is scheduled.
    if (head == null) {
      head = new AsyncTimeout();
      new Watchdog().start();
    }

    long now = System.nanoTime();
    if (timeoutNanos != 0 && hasDeadline) {
      // Compute the earliest event; either timeout or deadline. Because nanoTime can wrap around,
      // Math.min() is undefined for absolute values, but meaningful for relative ones.
      node.timeoutAt = now + Math.min(timeoutNanos, node.deadlineNanoTime() - now);
    } else if (timeoutNanos != 0) {
      node.timeoutAt = now + timeoutNanos;
    } else if (hasDeadline) {
      node.timeoutAt = node.deadlineNanoTime();
    } else {
      throw new AssertionError();
    }

    // Insert the node in sorted order.
    long remainingNanos = node.remainingNanos(now);
    for (AsyncTimeout prev = head; true; prev = prev.next) {
      if (prev.next == null || remainingNanos < prev.next.remainingNanos(now)) {
        node.next = prev.next;
        prev.next = node;
        if (prev == head) {
          AsyncTimeout.class.notify(); // Wake up the watchdog when inserting at the front.
        }
        break;
      }
    }
  }
```
Then we set a break point in the for loop, and found out that prev.next is prev itself, meaning that there's a circle in LinkedList, which causes the dead loop.

We're not able to reproduce it consistently, but it happens more frequently when there're a lot of network requests running at the same time, thus we suspect it to be a rare race condition.

And this happens during development, so it's not the same issue as https://github.com/square/okio/issues/185 since no ProGuard is involved.

I'm not sure that if it is a known issue that has been fix in the later version. What we're going to do to work around this issue is that we will periodically check if there's a circle in that LinkedList and fix that by using reflection. If anyone has a better fix please suggest.

Thanks in advance.

swankjesse:
[...]

PR #818, opened 2020-11-17
Prevent calling scheduleTimeout twice on an AsyncTimeout

findepi:
Invoking `enter` concurrently on an `AsyncTimeout` instance may lead to
`scheduleTimeout` being invoked twice.  If `scheduleTimeout` is called
on an `AsyncTimeout` instance that is already scheduled, it could lead
to a cycle in the single linked list starting with `head`, which in turn
would lead to an infinite loop in `cancelScheduledTimeout`.

Fixes: https://github.com/square/okio/issues/624

cc @wendigo @losipiuk

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -36,30 +36,28 @@ import java.util.concurrent.TimeUnit
  * [timedOut] is asynchronous, and may be called after [exit].
  */
 open class AsyncTimeout : Timeout() {
-  /** True if this node is currently in the queue.  */
-  private var inQueue = false
+  /** True if this node has been put in the queue.  */
+  // Guarded by AsyncTimeout.class
+  private var scheduled = false
 
   /** The next node in the linked list.  */
+  // Guarded by AsyncTimeout.class
   private var next: AsyncTimeout? = null
 
   /** If scheduled, this is the time that the watchdog should time this out.  */
   private var timeoutAt = 0L
 
   fun enter() {
-    check(!inQueue) { "Unbalanced enter/exit" }
     val timeoutNanos = timeoutNanos()
     val hasDeadline = hasDeadline()
     if (timeoutNanos == 0L && !hasDeadline) {
       return // No timeout and no deadline? Don't bother with the queue.
     }
-    inQueue = true
     scheduleTimeout(this, timeoutNanos, hasDeadline)
   }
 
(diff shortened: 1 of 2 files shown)


swankjesse:
How does this happen? Doesn’t `enter()` / `exit()` defend against this?

findepi:
> How does this happen?

`enter` is not synchronized


However, I am not saying this is the _only_ explanation for the infinite loop in `cancelScheduledTimeout` due to a cycle.
This is _a_ potential explanation only.

swankjesse:
Lets move all access to `inQueue` into the synchronized `scheduleTimeout` function and `cancelScheduledTimeout` functions? That way we can avoid the O(N) loop!

findepi:
We have O(n) for insertion anyway.
Also, `scheduleTimeout` is static, so i can't check `inQueue` there. Do you mean to pass the object as instance (this) rather than as parameter? (i don't speak Kotlin, so i am calling these things wrong, sorry)

we could also make `enter` and `exit` synchronized as well, but I am not versed in the okio to appreciate the consequences

swankjesse:
I think for `scheduleTimeout` and `cancelScheduledTimeout` we’d just dereference `AsyncTimeout.inQueue`, which is the passed-in argument.

findepi:
@swankjesse i reworked based on your suggestion.
I also added a test which reproduced the problem pretty consistently on my machine.
The test does 10000 iterations, while on my machine the deadlock occurs between 100 and 200 iteration.
With the code fixed, it is not the longest test in `AsyncTimeoutTest`, e.g. taking half the time of `instancesAddedAtFront`.

i struggled with formatting the code in the appropriate way, and some guidance would be helpful.

swankjesse:
The test ignores an IllegalStateException... when you observed the infinite loop in the wild was it after ignoring such exceptions?

findepi:
i observed this in Presto's project testing infrastructure (https://github.com/prestosql/presto/issues/5422).
There we use https://github.com/testcontainers/testcontainers-java/, which uses https://github.com/docker-java/docker-java, which uses `okhttp`.

AFAICT there was no `IllegalStateException` propagated to our code.

findepi:
Is there anything i can improve in my PR?

findepi:
rebased

swankjesse:
Got it. I suspect there's some concurrency problems in there. This PR is a good idea so things fall more gracefully in the event of concurrency problems.

swankjesse:
Thanks! I merged this manually after reverting some of the naming changes. I also dropped the concurrency test.

PR #822, opened 2020-12-01
Fix Android cipher issue

MartinDevi:
The Android implementation of ciphers has a weird behaviour, which I would qualify as a bug. On updates to the ciphers, it throws a `ShortBufferException` if the output array is smaller than the value that it estimates through its internal call to `getOutputSizeForUpdate`, which is itself erroneous because it returns a value which is too high. 

Here's an example illustrating the issue. The error is thrown even though the requested size isn't actually needed.

```kotlin
@Test
fun testEncrypt() {
    val key = Random.nextBytes(32)
    val cipher = Cipher.getInstance("AES/CBC/PKCS7Padding").apply {
        init(Cipher.ENCRYPT_MODE, SecretKeySpec(key, "AES"))
    }

    val input = Random.nextBytes(32)
    val output = ByteArray(16)
    var outputOffset = 0
    outputOffset += cipher.update(input, 0, 1, output, outputOffset) 
    // Here, outputOffset == 0, which is normal since no complete block has been processed
    outputOffset += cipher.update(input, 1, 16, output, outputOffset) 
    // throws `javax.crypto.ShortBufferException: output buffer too small during update: 16 < 32`
    // However if the output array is of length 32, outputOffset == 16, which is normal since a single complete block has been processed
}
```

This fix therefore removes from okio's `CipherSource` and `CipherSink` code the assumption that the output size is smaller or equal to the input size. The input size is shrunk until the output size is enough to work within a segment. Since the step used during the shrink is the block size, it should only take one step for that constraint to be met.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/CipherSink.kt
+++ b/okio/src/jvmMain/kotlin/okio/CipherSink.kt
@@ -26,10 +26,8 @@ class CipherSink(
   private var closed = false
 
   init {
-    // Require block cipher, and check for unsupported (too large) block size (should never happen
-    // with standard algorithms)
+    // Require block cipher
     require(blockSize > 0) { "Block cipher required $cipher" }
-    require(blockSize <= Segment.SIZE) { "Cipher block size $blockSize too large $cipher" }
   }
 
   @Throws(IOException::class)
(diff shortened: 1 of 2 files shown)


MartinDevi:
Scratch the previous example, it's actually event simpler.

```kotlin
@Test
fun testEncrypt() {
    val key = Random.nextBytes(32)
    val cipher = Cipher.getInstance("AES/CBC/PKCS7Padding").apply {
        init(Cipher.ENCRYPT_MODE, SecretKeySpec(key, "AES"))
    }

    val input = Random.nextBytes(16)
    val output = ByteArray(16)
    cipher.update(input, 0, 16, output, 0) 
}
```

The above fails with `javax.crypto.ShortBufferException: output buffer too small during update: 16 < 32`. But if the `output` array is of size 32, it only writes 16 bytes to the output, as it should. It mistakenly throws an error saying it needs more space, even though it doesn't.

JakeWharton:
Can you add that test case to the PR but through the public API of CipherSink and CipherSource?

MartinDevi:
> Can you add that test case to the PR but through the public API of CipherSink and CipherSource?

The current tests are actually already covering this case, they fail on Android. Specifically `CipherSinkTest.encryptLarge` and `CipherSinkTest.decryptLarge` fail. Since this is an Android-only issue though, they pass in the CI.

swankjesse (approved):

swankjesse:
Thanks for fixing this.

PR #827, opened 2020-12-06
Add an API for file metadata

swankjesse:


okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+   * supports this feature. This is typically available on Windows NTFS filesystems and not
+   * available on UNIX or Windows FAT filesystems.
+   */
+  val createdAtMillis: Long?,
+
+  /**
+   * Returns the system time of the host computer when this file was most recently written.
+   *
+   * Note that the accuracy of the returned time may be much more coarse than its precision. In
+   * particular, this value is is expressed with millisecond precision but may be accessed at
```
Egorand, line 77:
```suggestion
   * particular, this value is expressed with millisecond precision but may be accessed at
```

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+   * second- or day-accuracy only.
+   */
+  val lastModifiedAtMillis: Long?,
+
+  /**
+   * Returns the system time of the host computer when this file was most recently read or written.
+   *
+   * Note that the accuracy of the returned time may be much more coarse than its precision. In
+   * particular, this value is is expressed with millisecond precision but may be accessed at
```
Egorand, line 86:
```suggestion
   * particular, this value is expressed with millisecond precision but may be accessed at
```

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+
+  /** True if the path refers to a directory that contains 0 or more child paths. */
+  val isDirectory: Boolean,
+
+  /**
+   * Returns the number of bytes readable from this file. The amount of storage resources consumed
+   * by this file may be larger (due to block size overhead, redundant copies for RAID, etc.), or
+   * smaller (due to filesystem compression, shared inodes, etc).
+   */
+  val size: Long?,
```
Egorand, line 64: In which cases will this return null? Will it always be null if `isDirectory` is true?

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+  val createdAtMillis: Long?,
+
+  /**
+   * Returns the system time of the host computer when this file was most recently written.
+   *
+   * Note that the accuracy of the returned time may be much more coarse than its precision. In
+   * particular, this value is is expressed with millisecond precision but may be accessed at
+   * second- or day-accuracy only.
+   */
+  val lastModifiedAtMillis: Long?,
```
Egorand, line 80: Same as with `size`, maybe worth being more explicit about the cases when this is expected to be null. Also applies to `lastAccessedAtMillis`.

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+   * particular, this value is is expressed with millisecond precision but may be accessed at
+   * second- or day-accuracy only.
+   */
+  val lastAccessedAtMillis: Long?
+) {
+  override fun equals(other: Any?) = other is FileMetadata && toString() == other.toString()
+
+  override fun hashCode() = toString().hashCode()
+
+  override fun toString(): String {
```
Egorand, line 95: nit: is it worth caching this string to amortize the performance of `toString`, `equals` and `hashCode`?

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
```diff
@@ -0,0 +1,105 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+/**
+ * Description of a file or another object referenced by a path.
```
Egorand, line 19: I really like this comment

okio-files/src/commonMain/kotlin/okio/time.kt
```diff
@@ -0,0 +1,53 @@
...
+
+import kotlinx.datetime.Instant
+
+fun FileMetadata(
+  isRegularFile: Boolean = false,
+  isDirectory: Boolean = false,
+  size: Long? = null,
+  createdAt: Instant? = null,
+  lastModifiedAt: Instant? = null,
+  lastAccessedAt: Instant? = null
```
Egorand, line 26:
```suggestion
  lastAccessedAt: Instant? = null,
```

okio-files/src/commonMain/kotlin/okio/time.kt
```diff
@@ -0,0 +1,53 @@
...
+): FileMetadata {
+  return FileMetadata(
+    isRegularFile = isRegularFile,
+    isDirectory = isDirectory,
+    size = size,
+    createdAtMillis = createdAt?.toEpochMilliseconds(),
+    lastModifiedAtMillis = lastModifiedAt?.toEpochMilliseconds(),
+    lastAccessedAtMillis = lastAccessedAt?.toEpochMilliseconds())
```
Egorand, line 34:
```suggestion
    lastAccessedAtMillis = lastAccessedAt?.toEpochMilliseconds()),
```

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
```diff
@@ -286,6 +295,52 @@ abstract class FileSystemTest(
...
+    val minTime = clock.now().floorSeconds()
+    val path = base / "file-metadata"
+    path.writeUtf8("hello, world!")
+    val maxTime = clock.now().plus(1.seconds).floorSeconds()
+
+    val metadata = filesystem.metadata(path)
+    assertTrue(metadata.isRegularFile)
+    assertFalse(metadata.isDirectory)
+    assertEquals(13, metadata.size)
+    assertTrue { (metadata.createdAt ?: minTime) in minTime..maxTime }
```
Egorand, line 311: Do we expect these to ever be null under test?

okio-files/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
```diff
@@ -26,6 +29,36 @@ object JvmSystemFilesystem : Filesystem() {
...
+
+  /**
+   * Returns this time as a epoch millis. If this is 0L this returns null, because epoch time 0L is
+   * a special value that indicates the requested time was not available.
+   */
+  private fun FileTime.zeroToNull(): Long? {
+    return when (val millis = toMillis()) {
+      0L -> null
+      else -> millis
+    }
```
Egorand, line 59:
```suggestion
    return toMillis().takeIf { it != 0L }
```

Egorand (approved):

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
Egorand: We might get more compile-time safety (at the cost of more complex abstractions) if we model it this way:
```kotlin
sealed class FileMetadata {
  class RegularFileMetadata(val size: Long) : FileMetadata() // size always available, isRegularFile is redundant
  class DirectoryMetadata() : FileMetadata() // size is never available, isDirectory is redundant
}
```

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
swankjesse: Fixed!

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
swankjesse: Fixed!

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
swankjesse: Unfortunately the true type hierarchy isn’t static enough to model this way. For example, on my Mac I was getting a length of 64 bytes for the size of an empty directory!

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
swankjesse: I’d love to be more explicit, but the guarantees offered by the underlying systems are extremely weak and filesystem dependent. One trivial example: a file on a readonly media like a CD-ROM doesn’t have a mechanism to track last access time.

https://docs.microsoft.com/en-us/windows/win32/sysinfo/file-times

> Not all file systems can record creation and last access times, and not all file systems record them in the same manner. For example, the resolution of create time on FAT is 10 milliseconds, while write time has a resolution of 2 seconds and access time has a resolution of 1 day, so it is really the access date. The NTFS file system delays updates to the last access time for a file by up to 1 hour after the last access.

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
swankjesse: I was contemplating it. I think it’s probably quite unlikely to ever call `equals()` or `hashCode()` on these values.

okio-files/src/commonMain/kotlin/okio/time.kt
swankjesse: No trailing commas until we update spotless.

okio-files/src/commonMain/kotlin/okio/time.kt
swankjesse: Same.

okio-files/src/commonTest/kotlin/okio/files/FileSystemTest.kt
swankjesse: Mostly depends on the operating system / file system combo. I’d like the tests to reflect the true dynamic range of the values, even if our test coverage rarely runs on say, FAT filesystems.

okio-files/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
swankjesse: Neat.

okio-files/src/commonMain/kotlin/okio/FileMetadata.kt
Egorand: Interesting! But makes sense

okio-files/src/commonMain/kotlin/okio/time.kt
Egorand: I'll look into it

PR #828, opened 2020-12-06
Dokka 1.4.20 & generate docs in HTML

Egorand:


swankjesse:
Is HTML better? If we’re gonna do it for Okio we should do it for everything.

I like GFM for mkdocs, especially how it interacts with search.

Egorand:
I like how Dokka 1.4 renders docs in HTML, especially for multiplatform projects, IMO it's better than what MkDocs does - check out [KotlinPoet's docs](https://square.github.io/kotlinpoet/1.x/kotlinpoet/com.squareup.kotlinpoet/index.html). Don't feel strongly though, happy to revert.

swankjesse:
You're right that is nice. We aught to swap out the awkward DOCS thing though.

Egorand:
> We aught to swap out the awkward DOCS thing though.

Egorand:
> We aught to swap out the awkward DOCS thing though.

swankjesse:
It looks like the project’s brand, but it’s the tool’s brand.

PR #830, opened 2020-12-07
Metadata for Kotlin/Native

swankjesse:


Egorand (approved):

okio-files/build.gradle
```diff
@@ -44,7 +44,18 @@ kotlin {
     unixMain {
       dependsOn posixMain
     }
-    configure([macosX64Main, linuxX64Main]) {
+    appleMain {
```
Egorand, line 47: I wonder if Kotlin has hierarchical source sets for non-iOS native platforms

swankjesse: We end up using the full hierarchy here. 
Common, Native (-JVM), Unix (-Windows), Linux (-Apple)

PR #831, opened 2020-12-07
Adds notes to SegmentPool

adriancole:
Since this type's creation, I had it on my TODO to study the impl, so
that I could love it for good reason. I've added some notes which might
help others love it faster.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
@@ -120,7 +120,7 @@ internal actual object SegmentPool {
   }
 
   private fun firstRef(): AtomicReference<Segment?> {
-    // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread
+    // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread.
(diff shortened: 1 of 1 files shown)


okio/src/jvmMain/kotlin/okio/SegmentPool.kt
```diff
@@ -112,12 +113,14 @@ internal actual object SegmentPool {
...
-    // If we raced another operation: Don't recycle this segment.
+    // If we lost a race with another operation, don't recycle this segment.
+    if (!firstRef.compareAndSet(first, segment)) {
+      segment.next = null // Don't leak a reference in the pool either!
+    }
   }
 
   private fun firstRef(): AtomicReference<Segment?> {
-    // Get a value in [0..HASH_BUCKET_COUNT).
+    // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread
```
adriancole, line 123:
```suggestion
    // Get a value in [0..HASH_BUCKET_COUNT) based on the current thread.
```

swankjesse (approved):

swankjesse:
Great!

PR #833, opened 2020-12-09
Windows Paths

swankjesse:
The big challenge in this design is how to offer different behavior
for Windows vs. UNIX paths without a lot of baggage in the API.
Developers shouldn't have to choose a PathScheme or something!
So instead we just use the first slash character (`/` or `\`) as
a hint.

### DIFF
--- a/okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
+++ b/okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
@@ -16,6 +16,7 @@
 package okio.files
 
 import okio.Path.Companion.toPath
+import kotlin.test.Ignore
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertNull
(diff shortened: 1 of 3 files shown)


okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
```diff
@@ -198,4 +199,68 @@ class PathTest {
...
+  @Test
+  fun `windows absolute path`() {
+    assertEquals("\\Program Files", "\\Program Files".toPath().toString())
+    assertEquals("\\Program Files\\Photoshop".toPath(), "\\Program Files".toPath() / "Photoshop")
+  }
+
+  /** If a volume is present in the relative path... */
+  @Test
+  fun `windows volume relative path`() {
+    assertEquals("c:notepad.exe".toPath("\\"), "A:\\DOOM".toPath() / "c:notepad.exe")
```
Egorand, line 245: Did DOOM use to fit on a floppy drive?

Egorand (approved):
Damn file systems are hard

okio-files/src/commonTest/kotlin/okio/files/PathTest.kt
swankjesse: That’s how I played it! https://www.ebay.com/b/Doom-PC-1993-Video-Games/139973/bn_43712974

Egorand: Amazing! First PC I had came with a CD-ROM

PR #834, opened 2020-12-10
Extract FakeFileSystem into a new okio-testing module

swankjesse:
I want FakeFileSystem to depend on experimental APIs in kotlinx.datetime.
And I want to fold okio-files into the main okio module, where a dependency
on kotlinx.datetime would be too much.

### DIFF
--- a/okio-testing/build.gradle
+++ b/okio-testing/build.gradle
@@ -1,0 +1,22 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+
+kotlin {
+  jvm {
+    withJava()
+  }
+  if (kmpNativeEnabled) {
+    macosX64()
+    linuxX64()
+    mingwX64()
+  }
+  sourceSets {
+    commonMain {
+      dependencies {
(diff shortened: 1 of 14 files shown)


Egorand (approved):

yogurtearl:
@swankjesse any reason to not move `FakeFilesystemTest` over to `okio-testing` as well?

swankjesse:
@yogurtearl it shares code with the real Filesystem test, and it's hard to make the dependencies line up.

okio-testing/build.gradle
```diff
@@ -0,0 +1,22 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+
+kotlin {
+  jvm {
+    withJava()
```
JakeWharton, line 5: What Java code is in this module?

swankjesse: None. I’m attempting to configure this as close as possible to the main `okio` artifact so I can later merge the two without much friction.

PR #838, opened 2020-12-15
Filesystem.appendingSink

swankjesse:


okio-files/src/commonTest/kotlin/okio/files/AbstractFilesystemTest.kt
```diff
@@ -114,6 +114,31 @@ abstract class AbstractFilesystemTest(
...
+  @Test
+  fun `appending sink appends to existing file`() {
+    val path = base / "appending-sink-appends-to-existing-file"
+    path.writeUtf8("hello, world!\n")
+    val sink = filesystem.appendingSink(path)
+    val buffer = Buffer().writeUtf8("this is added later!")
+    sink.write(buffer, buffer.size)
+    sink.close()
+    assertTrue(path in filesystem.list(base))
+    assertEquals(0, buffer.size)
```
JakeWharton, line 126: Skeptical of these two assertions being relevant. The first is a side-effect of `writeUtf8` which should be covered elsewhere and in the second you're testing Okio internal side-effects of `write` and not `appendingSink` itself.

The following assertion covers the behavior of `appendingSink` and I think it's all that's needed.

JakeWharton (approved):

PR #839, opened 2020-12-15
Use custom wordmark in dokkaHtml

ajalt:
The latest version of Dokka allows you to add custom CSS and assets to the generated html, including the wordmark that is added to he top left of every page.

This commit replaces the default Dokka wordmark with `logo-square.png` which is already present in the repo. You'll probably want to use an different image that shows up better on a light background.

Screenshot_2020-12-15 okio

### DIFF
--- a/gradle/gradle-mvn-mpp-push.gradle
+++ b/gradle/gradle-mvn-mpp-push.gradle
@@ -11,11 +11,11 @@ def dokkaConfiguration = {
       skipDeprecated.set(true)
       jdkVersion.set(8)
       perPackageOption {
-        prefix.set("com.squareup.okio")
+        matchingRegex.set("com\\.squareup.okio.*")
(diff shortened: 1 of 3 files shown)


gradle/gradle-mvn-mpp-push.gradle
```diff
@@ -11,11 +11,11 @@ def dokkaConfiguration = {
       skipDeprecated.set(true)
       jdkVersion.set(8)
       perPackageOption {
-        prefix.set("com.squareup.okio")
+        matchingRegex.set("com\\.squareup.okio.*")
```
ajalt, line 14: The `prefix` property was removed in the latest version of dokka

swankjesse (approved):

swankjesse:
Rad!

Issue #814, opened 2020-11-16
okio-async module for Kotlin coroutines based asyncio


kevincianfarini:
I'm aware that the okio team has made attempts at integrating coroutines but have ultimately dropped the issue in hopes that Loom would solve for this. Seeing that okio is becoming increasingly multiplatform, Loom would only solve part of the problem. 

I'm proposing that this issue tracks progress on a multiplatform okio artifact, `okio-async`, which implements non-blocking and asynchronous I/O for the following platforms. 

* JVM/Android (via java.nio)
* Linux (via ~~epoll~~ io_uring)
* macOS/iOS/watchOS (via kqueue)
* mingw (via IOCTL)
* node.js (not sure if this makes sense)

There has been prior work done on this. In particular, I think that [okiox](https://github.com/bnorm/okiox/tree/master/coroutines) by @bnorm will be a good starting resource.

The following is a tracked list of functionalities which should be made available. 

# Tracking

- [ ] An `AsyncBuffer` which serves the same purpose as `Buffer`, however it implements `AsyncSource` as well as `AsyncSink`.  
- [ ] Multiplexed TCP Socket I/O
    - [ ] JVM
    - [ ] Linux
    - [ ] macOS, iOS, watchOS
    - [ ] Windows
    - [ ] node.js
- [ ] Multiplexed Server TCP Socket I/O
    - [ ] JVM
    - [ ] Linux
    - [ ] macOS, iOS, watchOS
    - [ ] Windows
    - [ ] node.js
- [ ] Multiplexed UDP Socket I/O
    - [ ] JVM
    - [ ] Linux
    - [ ] macOS, iOS, watchOS
    - [ ] Windows
    - [ ] node.js
- [ ] Asyncronous File I/O ([see why this can't be non-blocking here](https://www.remlab.net/op/nonblock.shtml))
    - [ ] JVM
    - [ ] Linux
    - [ ] macOS, iOS, watchOS
    - [ ] Windows
    - [ ] node.js

kevincianfarini:
I'm interested to hear thoughts on this approach. My rationale for this approach is that those using multiplatform kotlin to interact with okio should probably also reach for coroutines for async capabilities. Therefore it makes sense to break this out into it's own module. 

However, I do forsee a bit of code duplication. Particularly, things like `HashingSink/HashingSource` must have async counterparts. In reality, they do a lot of the same work. I don't see a very easy way to bridge to two in order to de-duplicate code which would be easy. 

[...]

PR #841, opened 2020-12-17
NodeJS filesystem

swankjesse:


okio-files/src/jsMain/kotlin/okio/FileSink.kt
```diff
@@ -0,0 +1,51 @@
...
+) : Sink {
+  private var closed = false
+
+  override fun write(source: Buffer, byteCount: Long) {
+    require(byteCount >= 0L) { "byteCount < 0: $byteCount" }
+    require(source.size >= byteCount) { "source.size=${source.size} < byteCount=$byteCount" }
+    check(!closed) { "closed" }
+
+    val data = Uint8Array(byteCount.toInt())
+    data.set(source.readByteArray().toTypedArray())
```
swankjesse, line 32: Working across Kotlin ByteArrays and JavaScript Uint8Arrays is pretty awful. They’re almost identical in capability, but I couldn’t find efficient ways to move data between them.

okio-files/src/jsMain/kotlin/okio/FileSource.kt
```diff
@@ -0,0 +1,55 @@
...
+      fd = fd,
+      buffer = data,
+      length = byteCount,
+      offset = 0,
+      position = null
+    ).toInt()
+
+    if (readByteCount == 0) return -1L
+
+    for (i in 0 until readByteCount) {
```
swankjesse, line 41: Another byte-at-a-time copy because of gaps between ByteArray and UInt8Array.

Almost makes me want to petition Kotlin people to make ByteArray ‘actual’ to UInt8Array, if that’s a thing.

okio-files/src/jsMain/kotlin/okio/FileSink.kt
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import fs.writeSync
```
JakeWharton, line 18: This means you can't use it in any critical path on a server which is unfortunate.

okio-files/src/jsMain/kotlin/okio/NodeJsSystemFilesystem.kt
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * Use [Node.js APIs][node_fs] to implement the Okio filesystem interface.
+ *
+ * This class needs to make calls to some fs APIs that have multiple competing overloads. To
+ * unambiguously select an overload this passes `undefined` as the target type to some functions.
+ *
+ * [node_fs]: https://nodejs.org/dist/latest-v14.x/docs/api/fs.html
+ */
+object NodeJsSystemFilesystem : Filesystem() {
```
JakeWharton, line 34: This should probably have the word "Sync" in it so callers are aware that it does not use non-blocking APIs.

JakeWharton (approved):

okio-files/src/jsMain/kotlin/okio/FileSink.kt
swankjesse: Agreed!

swankjesse: We can discuss our suspending strategy here. I don’t think this PR precludes that.
https://github.com/square/okio/issues/814

okio-files/src/jsMain/kotlin/okio/NodeJsSystemFilesystem.kt
swankjesse: Oh, I should have made this `internal`, users won’t ever interact directly with this. But we should document `Filesystem` as synchronous / blocking.

okio-files/src/jsMain/kotlin/okio/FileSink.kt
Egorand: > This means you can't use it in any critical path on a server

Interesting, why so? Do we want something async instead?

okio-files/src/jsMain/kotlin/okio/FileSink.kt
Egorand: Maybe worth filing a YouTrack issue?

Egorand (approved):

okio-files/src/jsMain/kotlin/okio/FileSink.kt
JakeWharton: Node runs on a single thread just like browser JS. It's an event loop. Blocking I/O blocks the entire server. So it's fine for command-line utilities running on node but not a proper server.

Egorand: Gotcha, makes sense

JakeWharton: > Do we want something async instead?

The entire design is built around being synchronous. To wrap the forthcoming filesystem API in the browser and to support Node's regular async file APIs you basically need a parallel design.

swankjesse: In particular, making this async makes it more difficult to use it in OkHttp’s cache, which is my initial target use case. OkHttp’s cache is blocking.

okio-files/src/jsMain/kotlin/okio/FileSink.kt
swankjesse: Probably. I need to do some research to file the right issue so that it’s actionable and compatible.

yschimke: Out of my depth here, but anything to learn from kotlinx-io?

https://github.com/Kotlin/kotlinx-io/blob/master/core/jsMain/src/kotlinx/io/buffer/Buffer.kt
https://kotlinlang.org/api/latest/jvm/stdlib/org.khronos.webgl/-data-view/

PR #842, opened 2020-12-17
Fix `okio-multiplatform` publication

martinbonnin:
Use `okio-files-multiplatform` for the okio-files multiplatform artifact so that it does not overwrite the regular `okio-multiplatform` artifact

Many thanks to @PaulWoitaschek for flagging this!


Egorand (approved):
Great catch, thanks!

swankjesse (approved):

PR #843, opened 2020-12-18
JVM interop functions and documentation

swankjesse:


okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -15,6 +15,74 @@
...
+ *    another filesystem implementation.
+ *
+ * For improved capability and testability, consider structuring your classes to dependency inject
+ * a `Filesystem` rather than using [SYSTEM] directly.
+ *
+ * Limited API
+ * -----------
+ *
+ * This interface is limited in which filesystem features it supports. Applications that need rich
+ * filesystem features should use another API, possibly alongside this API.
```
yschimke, line 44: Is it worth having a clean example of this escape hatch? i.e. how would you temporarily access the JDK Path or File objects when you want to opportunistically call those methods?

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -15,6 +15,74 @@
...
+ *
+ * It supports the path schemes of both Windows (like `C:\Users`) and UNIX (like `/home`). Note that
+ * path resolution rules differ by platform.
+ *
+ * Differences vs. Java IO APIs
+ * ----------------------------
+ *
+ * The `java.io.File` class is Java's original filesystem API. The `delete` and `renameTo` methods
+ * return false if the operation failed. The `list` method returns null if the file isn't a
+ * directory or could not be listed. This class always throws `IOExceptions` when operations don't
```
yschimke, line 75: For something like delete that might fail because it got deleted in the meantime, is it worth having some optionally includable safe operators?  Maybe extension methods .deleteSafely()? or deleteIfExists()?

swankjesse: Yeah, there's a bunch of convenience features that we can build on top of these primitives. createDirectories() !

okio-files/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: I don't really want to get into those examples because it gets ugly fast. The first thing you lose is testability!

JakeWharton (approved):

Egorand (approved):

PR #845, opened 2020-12-19
Fix some bugs in Windows path handling

swankjesse:
Also improve the structure of the path tests.


swankjesse:
Today I learned that you can navigate to `\\localhost\c$\Windows` in Windows Explorer!

JakeWharton (approved):

PR #848, opened 2020-12-21
Fold okio-files into the main okio artifact

swankjesse:


JakeWharton:
Is there going to be an experimental annotation? Making this immediately stable seems like a bad idea.

swankjesse:
I've been thinking of the API and JVM implementation here as the follow-on to the Filesystem interface in OkHttp.

But there's a lot that's immature:
 - JS
 - Native
 - Path API
 - Metadata API

Yeah, let's do a couple releases with an experimental annotation.

okio/build.gradle
```diff
@@ -66,45 +100,83 @@ kotlin {
       dependsOn nonJvmMain
       dependencies {
         api deps.kotlin.stdLib.js
+        implementation("org.jetbrains.kotlinx:kotlinx-nodejs:0.0.7")
```
JakeWharton, line 103: Does this block Okio's APIs from going stable? Perhaps it would be better to define external functions for the things in use and drop this dependency.

JakeWharton (approved):

okio/build.gradle
swankjesse: Yeah, lemme add that work to the queue.

PR #852, opened 2020-12-23
Filesystem.createDirectories and deleteRecursively

swankjesse:
I've implemented each of these without making recursive function calls. This
is very defensive; I'm attempting to prevent StackOverflowErrors for deep
paths.

### DIFF
--- a/okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
+++ b/okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
@@ -46,8 +46,25 @@ class FakeFilesystem(
   private val openPathsMutable = mutableListOf<Path>()
 
   /**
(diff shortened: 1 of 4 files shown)


JakeWharton (approved):
Why members over extensions for these ones?

swankjesse:
Same reason as copy: it opens the door to more efficient platform-specific implementations.

JakeWharton:
For copy it makes sense. For these, I don't think so. You can always make them polymorphic later which will take precedence over the extension (given a recompilation). I have a hard time seeing a filesystem offer a primitive version of these operations.

swankjesse:
I guess there's a discussion to have about whether to do polymorphic or not when we don't currently have a polymorphic use case.

I claim polymorphic is better for Java callers and for future optionality. What's the benefit of doing it as an extension?

JakeWharton:
Java callers convinces me

PR #853, opened 2020-12-26
Run tests on Android

swankjesse:
This is currently completely manual. In a follow-up it would be nice
to get this running in CI.

On a Pixel 3A emulator running Android 30 I observed test crashes in
the following tests:

    selectSpanningMultipleSegments
    readLargeNioBufferOnlyReadsOneSegment
    gzipSink
    gzipSource

I needed to rename BufferedSourceTest to BufferedSourceJavaTest, and similarly
for BufferedSinkJavaTest and ByteStringJavaTest. We were seeing simple name
collisions in these tests which I believe were previously preventing these tests
from being executed.


build.gradle
```diff
@@ -21,6 +21,13 @@ buildscript {
   ]
 
   ext.deps = [
+      'android': [
+        "gradlePlugin": "com.android.tools.build:gradle:4.0.2",
```
Egorand, line 25:
```suggestion
        "gradlePlugin": "com.android.tools.build:gradle:4.1.1",
```

okio/src/commonTest/kotlin/okio/AbstractBufferedSourceTest.kt
```diff
@@ -1013,6 +1013,7 @@ abstract class AbstractBufferedSourceTest internal constructor(
     assertTrue(source.exhausted())
   }
 
+  /** Note that this test crashes the VM on Android. */
```
Egorand, line 1016: Planning to address this or ignore the test in a followup?

android-test/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,6 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="okio.android.test">
+  <application
```
Egorand, line 3: Do you need an `android:networkSecurityConfig="@xml/network_security_config"` attribute here? Also, what's the reasoning behind allow clear text traffic here but forbidding in the `network_security_config`, is this supposed to set different rules for different API versions?

Egorand (approved):
Very nice! I wonder how feasible it would be to run these in a Firebase test lab. For now though, are you planning to add a CI workflow?

okio/src/commonTest/kotlin/okio/AbstractBufferedSourceTest.kt
swankjesse: Hoping to fix in follow-up!

build.gradle
swankjesse: Fix in follow-up!

android-test/src/main/AndroidManifest.xml
swankjesse: To be honest it’s just ’cause I copy-pasted OkHttp’s AndroidManifest.xml, and it seemed like keeping this would avoid problems around raw TCP sockets. This isn’t a regular application so opting out of security policy seemed natural.

swankjesse:
And yes, I’d like to run regression tests in CI. Only after they’re passing!

PR #861, opened 2020-12-30
Define our own externals for Node.js fs and os APIs

swankjesse:


okio/src/jsMain/kotlin/okio/fs.kt
```diff
@@ -0,0 +1,173 @@
...
+ * [kotlinx_nodejs]: https://github.com/Kotlin/kotlinx-nodejs
+ */
+@file:JsModule("fs")
+@file:JsNonModule
+package okio
+
+import org.khronos.webgl.Uint8Array
+import kotlin.js.Date
+
+internal external fun closeSync(fd: Number)
```
swankjesse, line 79: `internal external fun` ?
that means we can have fun EVERYWHERE

okio/src/jsMain/kotlin/okio/fs.kt
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This class declares the subset of Node.js filesystem APIs that we need in Okio.
```
Egorand, line 18: As long as all of these aren't API we may retire them later if Dukat or kotlinx-nodejs improve

Egorand (approved):

okio/src/jsMain/kotlin/okio/fs.kt
Egorand: This has to be a tweet

okio/src/jsMain/kotlin/okio/fs.kt
swankjesse: definitely

swankjesse: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50341/files !

PR #863, opened 2020-12-30
Combine Platform.kt and -Platform.kt

swankjesse:


okio/src/commonMain/kotlin/okio/-Platform.kt
```diff
@@ -16,6 +16,14 @@
 
 package okio
 
+@ExperimentalFilesystem
```
Egorand, line 19: Any value to having experimental annotations on internal members?

Egorand (approved):

okio/src/commonMain/kotlin/okio/-Platform.kt
swankjesse: Without it the code won't compile!

PR #865, opened 2020-12-31
Filesystem.list() returns files in sorted order

swankjesse:


yschimke:
Since this is expensive for large lists and not always required, what about making it the default for a nullable parameter? i.e. you can opt out of it when order is not important but lists may be large?

Seems a shame to add this cost for all calls.

swankjesse:
I claim the cost of sorting is dwarfed by the cost of the syscalls to get the data. I think if we ever find a situation where the cost is meaningful the fix will be to add an alternative to list() that returns a Sequence.

yschimke:
I'll trust your judgement.  One counter argument is that you are making a very low level API (file IO) in a widely used library with a strictly controlled API. The barrier to adding this later is high. But you convinced me on the Sequence argument.

Big Cloud thinking - I'm thinking S3 via the Filesystem abstraction.  We don't want to list all files, we do want a sequence since a find operation could terminate early.  And perhaps then you also want that sequence method to support an ordering parameter (name, metadata, or custom).

okio/src/commonMain/kotlin/okio/ForwardingFilesystem.kt
```diff
@@ -160,7 +160,7 @@ abstract class ForwardingFilesystem(
   override fun list(dir: Path): List<Path> {
     val dir = onPathParameter(dir, "list", "dir")
     val result = delegate.list(dir)
-    return result.map { onPathResult(it, "list") }
+    return result.map { onPathResult(it, "list") }.sorted()
```
yschimke, line 163: Shouldn't be needed here.

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -249,6 +249,12 @@ class Path private constructor(
     return buffer.toPath(directorySeparator = slash)
   }
 
+  override fun compareTo(other: Path): Int {
```
yschimke, line 252: This and equals/hashCode probably requires a comment somewhere on case sensitivity.  Do we have that already.

okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
```diff
@@ -133,7 +133,7 @@ class FakeFilesystem(
     val element = requireDirectory(canonicalPath)
 
     element.access(now = clock.now())
-    return elements.keys.filter { it.parent == canonicalPath }
+    return elements.keys.filter { it.parent == canonicalPath }.sorted()
```
yschimke, line 136: Might be worth ensuring that allPaths is also predictable for tests.

okio/src/commonMain/kotlin/okio/ForwardingFilesystem.kt
swankjesse: Consider for example a (bad) mapping function that mapped `foo.jpg` to `foo.jpeg`. That'll impact sort order.

I'll add a test that fails without it.

okio/src/commonMain/kotlin/okio/Path.kt
swankjesse: Lemme add it.

Our rule is very straightforward: only the OS can do this correctly.

On my Mac I have a case-sensitive volume mounted for whenever I need to do awkward git stuff like filter-branch. The paths `/Volumes/case-sensitive/foo.txt` and `/VOLUMES/case-sensitive/foo.txt` point to the same file, but `/Volumes/case-sensitive/FOO.txt` is a different file.

okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
swankjesse: ooooh good call.

okio/src/commonMain/kotlin/okio/Filesystem.kt
```diff
@@ -135,7 +135,7 @@ abstract class Filesystem {
   }
 
   /**
-   * Returns the children of the directory identified by [dir].
+   * Returns the children of the directory identified by [dir]. The returned list is sorted.
```
Egorand, line 138: Is it worth specifying that it's "sorted using natural ordering"?

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -87,6 +87,37 @@ import kotlin.jvm.JvmStatic
...
+ *  * **Hard links.** UNIX filesystems permit multiple paths to refer for same file. The paths may
+ *    be wildly different, like `/Users/jesse/bruce_wayne.vcard` and
+ *    `/Users/jesse/batman.vcard`, but changes via either path are reflected in both.
+ *  * **Symlinks.** Symlinks permit multiple paths and directories to refer to the same file. On
+ *     macOS `/tmp` is symlinked to `/private/tmp`, so `/tmp/notes.txt` and `/private/tmp/notes.txt`
+ *     resolve to the same file.
+ *
+ * To test whether two paths refer to the same file, try [Filesystem.canonicalize] first. This
+ * follows symlinks and looks up the preserved casing for case-insensitive case-preserved paths.
+ * **This method does not guarantee a unique result, however.** For example, each hard links to a
```
Egorand, line 116:
```suggestion
 * **This method does not guarantee a unique result, however.** For example, each hard link to a
```

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -249,6 +280,12 @@ class Path private constructor(
     return buffer.toPath(directorySeparator = slash)
   }
 
+  override fun compareTo(other: Path): Int {
+    val bytesResult = bytes.compareTo(other.bytes)
+    if (bytesResult != 0) return bytesResult
+    return slash.compareTo(other.slash)
```
Egorand, line 286: Comparing slashes is cheaper, maybe switch the order of operations?

okio/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
```diff
@@ -62,7 +62,7 @@ internal open class JvmSystemFilesystem : Filesystem() {
       if (!file.exists()) throw FileNotFoundException("no such file $dir")
       throw IOException("failed to list $dir")
     }
-    return entries.map { dir / it }
+    return entries.map { dir / it }.sorted()
```
Egorand, line 65: Should we sort in place instead, since `map {}.sorted()` now copies `entries` twice?

Egorand (approved):

okio/src/commonMain/kotlin/okio/Filesystem.kt
swankjesse: Fixed.

okio/src/commonMain/kotlin/okio/Path.kt
yschimke: But also more likely to almost always be true, so worth deferring until after you check bytes.

okio/src/commonMain/kotlin/okio/Path.kt
swankjesse: Fixed.

okio/src/commonMain/kotlin/okio/Path.kt
swankjesse: I wanna order ’em that way to get this order:

`a/b`
`a\b`
`c/d`
`c\d`

okio/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
swankjesse: sure!

swankjesse: Fixed.

Egorand: Same thing in FFS, but perhaps less of an issue

JakeWharton (approved):

okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
```diff
@@ -62,12 +62,12 @@ class FakeFilesystem(
...
   val allPaths: Set<Path>
     get() {
-      val result = mutableSetOf<Path>()
+      val result = mutableListOf<Path>()
       for (path in elements.keys) {
         if (path.isRoot) continue
         result += path
       }
-      return result
+      return result.sorted().toSet()
```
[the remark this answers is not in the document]
swankjesse: Fixed!

okio-testing/src/commonMain/kotlin/okio/FakeFilesystem.kt
```diff
@@ -133,7 +135,7 @@ class FakeFilesystem(
     val element = requireDirectory(canonicalPath)
 
     element.access(now = clock.now())
-    return elements.keys.filter { it.parent == canonicalPath }
+    return elements.keys.filter { it.parent == canonicalPath }.sorted()
```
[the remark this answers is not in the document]
swankjesse: Fixed!

okio/src/commonMain/kotlin/okio/ForwardingFilesystem.kt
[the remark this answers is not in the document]
swankjesse: Fixed.

okio/src/jvmMain/kotlin/okio/JvmSystemFilesystem.kt
```diff
@@ -62,7 +62,12 @@ internal open class JvmSystemFilesystem : Filesystem() {
       if (!file.exists()) throw FileNotFoundException("no such file $dir")
       throw IOException("failed to list $dir")
     }
-    return entries.map { dir / it }
+    val result = mutableListOf<Path>()
+    for (entry in entries) {
+      result += dir / entry
+    }
```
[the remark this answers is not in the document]
swankjesse: Nice! Fixed.

PR #866, opened 2021-01-03
Example test for S3 paths

yschimke:
For discussion, what should the expectations be?

### DIFF
--- a/okio/src/commonTest/kotlin/okio/S3PathTest.kt
+++ b/okio/src/commonTest/kotlin/okio/S3PathTest.kt
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


okio/src/commonTest/kotlin/okio/S3PathTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import kotlin.test.assertFalse
+import kotlin.test.assertNull
+import kotlin.test.assertTrue
+
+@ExperimentalFilesystem
+class S3PathTest {
+  @Test
+  fun httpsRoot() {
+    val path = "https://s3.amazonaws.com/bucket/key".toPath()
+    assertEquals("https://s3.amazonaws.com/bucket".toPath(), path.parent)
```
yschimke, line 31: breaking on normalisation to single slash "https:/s3.amazonaws.com/bucket"

JakeWharton:
These are URIs, not filesystem paths. A URI contains a unix-style path part, but the whole URI won't be able to be handled.

yschimke:
> These are URIs, not filesystem paths. A URI contains a unix-style path part, but the whole URI won't be able to be handled.

so /bucket/key, easy to split on "/" with a limit of 2.

PR #869, opened 2021-01-04
Split Path into an expect/actual class

swankjesse:
I want to move methods into Path to make use from Java nicer.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Path.kt
+++ b/okio/src/jvmMain/kotlin/okio/Path.kt
@@ -1,0 +1,85 @@
+/*
+ * Copyright (C) 2020 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


okio/src/jvmMain/kotlin/okio/Path.kt
```diff
@@ -0,0 +1,82 @@
...
+@ExperimentalFileSystem
+actual class Path internal actual constructor(
+  internal actual val slash: ByteString,
+  internal actual val bytes: ByteString
+) : Comparable<Path> {
+  actual val isAbsolute: Boolean
+    get() = commonIsAbsolute()
+
+  actual val isRelative: Boolean
+    get() = !isAbsolute
```
Egorand, line 38: My preference would be to have common methods for these as well (`isRelative`, `name`, `toString`), even though they're just one-liners still feels like code duplication.

Egorand (approved):

okio/src/jvmMain/kotlin/okio/Path.kt
swankjesse: Done!

PR #870, opened 2021-01-05
Add a multiplatform okio.Closeable interface

swankjesse:
Not having it made interacting with Okio particularly annoying
on non-JVM platforms.

Note that we use our own use {} function on all platforms and do
not delegate to Kotlin's use {} function on the JVM. I couldn't
find a way to call Kotlin's use function from our own use function
since they both extend Closeable.


Egorand (approved):
Dope!

okio/src/nonJvmMain/kotlin/okio/Sink.kt
```diff
@@ -15,12 +15,15 @@
  */
 package okio
 
-actual interface Sink {
+actual interface Sink : Closeable {
```
swankjesse, line 18: I ended up having to put `Sink` back so the JVM one could keep the `Flushable` interface.

PR #871, opened 2021-01-06
Syntactic sugar for writing files

swankjesse:


docs/index.md
```diff
@@ -973,6 +973,108 @@ fun decryptAesToByteString(file: File, key: ByteArray, iv: ByteArray): ByteStrin
...
+val path = "readme.md".toPath()
+fileSystem.write(path) {
+  writeUtf8(
+    """
+    |Hello, World
+    |------------
+    |
+    |This is a sample file.
+    |""".trimMargin()
+  )
```
swankjesse, line 1032: compare vs. Kotlin’s API:

```
val file = File("/hello.txt")
file.writeText(
    """
    |Hello, World
    |------------
    |
    |This is a sample file.
    |""".trimMargin()
)
```

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -164,6 +185,26 @@ abstract class FileSystem {
...
+    }
+  }
+
+  /**
+   * Creates a sink to write [file], executes [writerAction] to write it, and then closes the sink.
+   * This is a compact way to write a file.
+   *
+   * @param file the relative or absolute [path][Path] of the file to write.
+   */
+  inline fun <T> write(file: String, writerAction: BufferedSink.() -> T): T {
```
swankjesse, line 204: This is the first time we’re coercing a string to a file in our APIs. I’m not 100% sure this is a good idea.

okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -131,6 +131,36 @@ abstract class AbstractFileSystemTest(
     source.close()
   }
 
+  @Test
+  fun readPath() {
+    val path = base / "read-path"
+    val string = "hello, read with a Path"
+    path.writeUtf8(string)
+
+    val byteCount = fileSystem.read(path) {
```
yschimke, line 140: This is nice!

okio/src/commonMain/kotlin/okio/FileSystem.kt
yschimke: Path is a better abstraction of filenames than Strings.  But String is the expedient default, two downsides to supporting string here are

1) You weaken the consistency of Path as the filename abstraction. People will use it with confusion.
2) You are opting in to build APIs that are not responsible for the edge cases around conversion. It seems cleaner for the caller to deal with that (or choose not to) outside the API.

okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -131,6 +131,36 @@ abstract class AbstractFileSystemTest(
...
+    assertEquals("success", byteCount)
+  }
+
+  @Test
+  fun readString() {
+    val path = base / "read-string"
+    val string = "hello, read with a String"
+    path.writeUtf8(string)
+
+    val byteCount = fileSystem.read(path.toString()) {
```
yschimke, line 155: why byteCount? result? It's a String.

docs/index.md
```diff
@@ -973,6 +973,108 @@ fun decryptAesToByteString(file: File, key: ByteArray, iv: ByteArray): ByteStrin
...
+}
+```
+
+Read the prefix of `index.html` that precedes the first `<html>` substring:
+
+```
+val path = "index.html".toPath()
+val untilHtmlTag = fileSystem.read(path) {
+  val htmlTag = indexOf("<html>".encodeUtf8())
+  return@read if (htmlTag != -1L) readUtf8(htmlTag) else null
```
yschimke, line 1016: Why the `return@read` here?

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -164,6 +185,26 @@ abstract class FileSystem {
   @Throws(IOException::class)
   abstract fun sink(file: Path): Sink
 
+  /**
+   * Creates a sink to write [file], executes [writerAction] to write it, and then closes the sink.
+   * This is a compact way to write a file.
+   */
+  inline fun <T> write(file: Path, writerAction: BufferedSink.() -> T): T {
```
yschimke, line 192: I use append much less, but it seems like top level given incrementally built output like log files.  Should this be a flag?

docs/index.md
yschimke: This feels like the write balance here.  The Kotlin API can easily be added as extension APIs without exploding the surface area of the API!

okio/src/commonMain/kotlin/okio/FileSystem.kt
swankjesse: Lemme pull it out!

okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
swankjesse: whoops! Fixed. Just wanted to make sure the return value went through.

docs/index.md
swankjesse: I’m an implicit return hater. Removed.

okio/src/commonMain/kotlin/okio/FileSystem.kt
swankjesse: I claim that most of the time when you want to append, you want to open a Sink, and keep it open for a while, probably in a property of some managing object like `Appender`. The `write` function is most useful when you want to write and immediately close.

swankjesse: (We can add an `append` function later if necessary!)

oldergod (approved):
Super nice.

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -153,6 +153,16 @@ abstract class FileSystem {
   @Throws(IOException::class)
   abstract fun source(file: Path): Source
 
+  /**
+   * Creates a source to read [file], executes [readerAction] to read it, and then closes the
+   * source. This is a compact way to read the contents of a file.
+   */
+  inline fun <T> read(file: Path, readerAction: BufferedSource.() -> T): T {
```
JakeWharton, line 160: I don't think I like these names. Maybe `useSource` and `useSink`? `withSource` and `withSink`?

PR #880, opened 2021-01-09
Redesign FakeFileSystem's Windows limitations

swankjesse:
Instead of having a single windows-vs-UNIX option, introduce individual
options that describe specific behavior, and functions to select UNIX
and Windows' options as a combo.

This permits a follow-up to forbid reading while writing, and also to
forbid multiple concurrent writers.


yschimke (approved):

Egorand (approved):
Nice!

PR #884, opened 2021-01-11
Forbid dangerous operations in FakeFileSystem by default

swankjesse:


okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -218,8 +238,13 @@ class FakeFileSystem(
     if (element !is File) {
       throw IOException("not a file: $file")
     }
+    if (!allowReadsWhileWriting) {
+      openFileOrNull(canonicalPath, operation = WRITE)?.let {
```
yschimke, line 242: nit: open as a adjective (is open) instead of a verb (to open) threw me initially. I thought this was opening the file.

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -218,8 +238,13 @@ class FakeFileSystem(
...
+        throw IOException("file is already open for writing $file", it.backtrace)
+      }
+    }
 
-    val openFile = OpenFile(canonicalPath, Exception("file opened for reading here"))
+    val openFile = OpenFile(canonicalPath, READ, Exception("file opened for reading here"))
     openFiles += openFile
```
yschimke, line 248: Should this entire method be synchronized? I don't think this implementation is designed to be highly concurrent, and 100% predictability probably beats any chance at race conditions in tests?

yschimke:
I really like this.  Should allow fast testing but with really strict checks to ensure that code written by default is held to the highest standards.

Egorand (approved):

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
swankjesse: Renamed to findOpenFile.

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
swankjesse: Yeah, I gotta do that. For JVM it’s easy, for native? Unclear.

PR #885, opened 2021-01-12
Fix volume letter parsing in Windows Paths that contain forward slashes.

ajalt:
#882 introduced a regression in Windows Path parsing. Windows allows forward slashes in path names, replacing them with backslashes as part of normalization (Here it's mentioned in [the Win32 docs for the `lpFileName` parameter][1], and [here it's mentioned for .NET][2]). `C:/foo` and `C:/foo\bar` are both valid Windows paths.

This PR removes the requirement for a backslash to be present for volume letters to be parsed.

Note that this doesn't fix the case of UNC paths that use forward slashes. A test like the following still fails:

```kotlin
@Test  fun windowsUncAbsolutePathForwardSlashes() {
  val path = "//server/project/notes.txt".toPath()
  assertEquals("//server/project/notes.txt", path.toString())
  assertEquals("//server/project".toPath(), path.parent)
  assertNull(path.volumeLetter)
  assertEquals("notes.txt", path.name)
  assertTrue(path.isAbsolute)
  assertFalse(path.isRoot)
}
```

Since `//foo` is valid as both a UNC path and a POSIX path, I don't believe paths like that can be parsed unambiguously without the notion of a system file separator or similar.

[1]: https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#parameters
[2]: https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats#canonicalize-separators

### DIFF
--- a/okio/src/commonTest/kotlin/okio/PathTest.kt
+++ b/okio/src/commonTest/kotlin/okio/PathTest.kt
@@ -147,6 +147,39 @@ class PathTest {
   }
 
   @Test
+  fun windowsAbsolutePathWithVolumeLetterForwardSlashes() {
+    val path = "C:/Windows/notepad.exe".toPath()
+    assertEquals("C:/Windows/notepad.exe", path.toString())
+    assertEquals("C:/Windows".toPath(), path.parent)
+    assertEquals('C', path.volumeLetter)
+    assertEquals("notepad.exe", path.name)
+    assertTrue(path.isAbsolute)
+    assertFalse(path.isRoot)
+  }
+
+  @Test
+  fun windowsAbsolutePathWithVolumeLetterMixedSlashesBackslashFirst() {
+    val path = "C:\\Windows/notepad.exe".toPath()
+    assertEquals("C:\\Windows\\notepad.exe", path.toString())
+    assertEquals("C:\\Windows".toPath(), path.parent)
+    assertEquals('C', path.volumeLetter)
+    assertEquals("notepad.exe", path.name)
+    assertTrue(path.isAbsolute)
+    assertFalse(path.isRoot)
+  }
+
+  @Test
+  fun windowsAbsolutePathWithVolumeLetterMixedSlashesSlashFirst() {
+    val path = "C:/Windows\\notepad.exe".toPath()
+    assertEquals("C:/Windows/notepad.exe", path.toString())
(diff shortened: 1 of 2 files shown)


okio/src/commonTest/kotlin/okio/PathTest.kt
```diff
@@ -146,6 +146,39 @@ class PathTest {
...
+    assertEquals('C', path.volumeLetter)
+    assertEquals("notepad.exe", path.name)
+    assertTrue(path.isAbsolute)
+    assertFalse(path.isRoot)
+  }
+
+  @Test
+  fun windowsAbsolutePathWithVolumeLetterMixedSlashesSlashFirst() {
+    val path = "C:/Windows\\notepad.exe".toPath()
+    assertEquals("C:/Windows/notepad.exe", path.toString())
```
ajalt, line 174: Paths separators are currently normalized based on the first slash in the path. This isn't necessarily wrong, but it is different from how JVM and most other systems behave.

swankjesse:
It was my (misguided?) intention to use the path character to imply which rules the caller wanted to follow.

Consider:

```
cd /tmp
mkdir -p C:/Windows
cd C:/Windows
```

On a UNIX system, `C:` is not special, and so `/tmp/C:/Windows` is a valid path.

I was thinking we’d look at the slash character to determine whether `C:` is special.

ajalt:
If the library uses the presence of a forward slash to dictate parsing behavior, I think that would make some common use cases more difficult. For example, if I wanted to take a path from user input and call `volumeLetter` on it, I would have to manually check the current operating system and replace forward slashes with backslashes on windows before calling `toPath`. If someone doesn't know about that behavior and doesn't pre-normalize the path strings, it would be a source of bugs.

swankjesse:
Good point!

I guess it comes to this decision:
 - should Path behave the same regardless of the host OS, or
 - should Path behave the same as the host OS

swankjesse:
We discussed offline! Thank you!

Decisions:
 * path resolution rules to follow the host OS, not the slash
 * make `FakeFileSystem` follow whatever OS its emulating (“emulateWindows()”)
 * add an API to support this: `"C:".toPath().resolve("\\tmp", windows=true)`. FakeFileSystem will use this for the above. Anybody else could use if they’re trying to be portable and don’t care what the host OS is.

PR #890, opened 2021-02-09
Bump the minimum JDK requirement to Java 8

swankjesse:


swankjesse:
I want to introduce a default method!

swankjesse:
https://developer.android.com/studio/write/java8-support

okio-fakefilesystem/build.gradle
```diff
@@ -44,6 +44,13 @@ kotlin {
   }
 }
 
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
Egorand, line 47: nit: move this block to the top-level build.gradle under the `subprojects` block?

okio/src/commonMain/kotlin/okio/-Platform.kt
```diff
@@ -47,3 +47,5 @@ expect interface Closeable {
   @Throws(IOException::class)
   fun close()
 }
+
+expect annotation class JvmDefault()
```
Egorand, line 51: Not clear what this does, also should it be `internal`?

Egorand (approved):
Hype! Will this be 3.0?

okio/src/commonMain/kotlin/okio/-Platform.kt
swankjesse: Yes, definitely

okio-fakefilesystem/build.gradle
swankjesse: Good idea.

okio/src/commonMain/kotlin/okio/-Platform.kt
swankjesse: Ugh, that’s the right fix but we can’t have Okio’s `@JvmDefault` be internal because it’s typealiased to a public type.

swankjesse: Moved to `okio.internal` which is less satisfying but should do the trick.

Egorand: That makes sense. Ideally it should just be present in `stdlib-common`.

build.gradle
```diff
@@ -117,4 +115,17 @@ subprojects {
       endWithNewline()
     }
   }
+
+  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+    kotlinOptions {
+      jvmTarget = "1.8"
+      freeCompilerArgs += "-Xjvm-default=enable"
```
JakeWharton, line 122: How about "all" instead

build.gradle
```diff
@@ -34,8 +34,6 @@ buildscript {
           'stdLib': [
```
JakeWharton, line 34: The stdlib is added automatically everywhere so none of these should be needed, realistically.

PR #891, opened 2021-02-09
Cursor API and File implementation

swankjesse:
The implementation is easy here; FileInputStream.channel does all of the
work.

### DIFF
--- a/okio/jvm/japicmp/build.gradle
+++ b/okio/jvm/japicmp/build.gradle
@@ -51,6 +51,7 @@ task japicmp(type: JapicmpTask, dependsOn: 'jar') {
   methodExcludes = [
       'okio.ByteString#getByte(int)', // Became 'final' in 1.15.0.
       'okio.ByteString#size()', // Became 'final' in 1.15.0.
+      'okio.Source#cursor()', // Introduced in 3.0.0
   ]
 }
 check.dependsOn japicmp
--- a/okio/src/commonMain/kotlin/okio/Cursor.kt
+++ b/okio/src/commonMain/kotlin/okio/Cursor.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


swankjesse:
(stacked PR)

okio/src/jvmMain/kotlin/okio/-Platform.kt
```diff
@@ -52,3 +52,5 @@ actual typealias EOFException = java.io.EOFException
 actual typealias FileNotFoundException = java.io.FileNotFoundException
 
 actual typealias Closeable = java.io.Closeable
+
+actual typealias JvmDefault = kotlin.jvm.JvmDefault
```
Egorand, line 56: It's strange that they don't have it in the common stdlib artifact, unlike some other annotations (https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.jvm/). Should we file an issue (unless someone already did it)?

Egorand (approved):

okio/src/jvmMain/kotlin/okio/-Platform.kt
swankjesse: Yep.

JakeWharton: The better approach is to switch to `-Xjvm-default=all` so that you don't need to annotate anything.

swankjesse: Oh yes, that's what I want.

armay:
Hi @swankjesse! Happy to see this feature finally arrived! I have a question. Is random access in version 3.0 going to be supported only by sources? I think the case of a sink is similar and it is worth implementing because you can treat a random access file both as a sink and as a source.
Before this PR, I had already been using my own home-made JVM-only implementation of random access source and sink, rather close to yours, in a pet project, and it seems to me like this approach works both ways.
Please, excuse me if random access sink support is already announced somewhere.

swankjesse:
@armay yep! Got a favorite use case for random access writes?

armay:
Great! Thanks for the quick reply! I'll think of a practical example and share it in this thread. My current use case is a bit abstract)

PR #893, opened 2021-02-10
New zipfilesystem module and FixedLengthSource class

swankjesse:
The FixedLengthSource will be necessary with the ZIP implementation. When
reading an uncompressed slice we need to put a bound on the end of the
source. When reading a compressed slice we expect it to exhaust exactly
as expected.

### DIFF
--- a/okio-zipfilesystem/build.gradle
+++ b/okio-zipfilesystem/build.gradle
@@ -1,0 +1,33 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
+
+kotlin {
+  jvm {
+    withJava()
+  }
+  sourceSets {
+    commonMain {
+      dependencies {
+        api deps.kotlin.stdLib.common
+        api deps.kotlin.time
+        api project(":okio")
+      }
+    }
+    jvmTest {
+      dependencies {
+        implementation deps.test.junit
+        implementation deps.test.assertj
+        implementation deps.kotlin.test.jdk
+      }
+    }
+  }
+}
+
+// modify these lines for MANIFEST.MF properties or for specific bnd instructions
+project.ext.bndManifest = '''
+    Export-Package: okio.zipfilesystem
+    Automatic-Module-Name: okio.zipfilesystem
+    Bundle-SymbolicName: com.squareup.okio.zipfilesystem
+    '''
+
+apply from: "$rootDir/okio/jvm/jvm.gradle"
+apply from: "$rootDir/gradle/gradle-mvn-mpp-push.gradle"
--- a/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/FixedLengthSource.kt
+++ b/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/FixedLengthSource.kt
@@ -1,0 +1,77 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio.zipfilesystem
+
+import okio.Buffer
+import okio.ForwardingSource
+import okio.IOException
+import okio.Source
+
+/**
+ * A source that returns [size] bytes of [delegate].
+ *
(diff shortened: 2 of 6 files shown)


okio-zipfilesystem/src/jvmTest/kotlin/okio/zipfilesystem/FixedLengthSourceTest.kt
```diff
@@ -0,0 +1,132 @@
...
+ */
+package okio.zipfilesystem
+
+import okio.Buffer
+import okio.IOException
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.Test
+import kotlin.test.fail
+
+internal class FixedLengthSourceTest {
```
bnorm, line 24: Thinking about a test case along the lines of:
1. Source of length 16, wrap with non-truncated fixed size of 10
2. Read 10 (`bytesReceived` now equal to fixed size)
3. Read 10 and expect error and no data

All the existing tests look to put `bytesReceived` over the size and never equal. I don't think there is a bug or anything, but it might be nice to have a test for this boundary condition.

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/FixedLengthSource.kt
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * A source that returns [size] bytes of [delegate].
+ *
+ * This throws an [IOException] if the delegate returns fewer than [size] bytes.
+ *
+ * If [truncate] is true, this truncates to [size] bytes. Otherwise this requires that [delegate]
+ * will return exactly [size] bytes, and will throw an [IOException] if it doesn't.
+ */
+internal class FixedLengthSource(
```
JakeWharton, line 31: Can this ship in the core behind an extension on `Source` so that it can be used in OkHttp and places like length-delimited proto streams with Wire?

swankjesse: Yes!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/FixedLengthSource.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    if (result != -1L) bytesReceived += result
+
+    // Throw an exception if we received too few bytes or too many.
+    if ((bytesReceived < size && result == -1L) || bytesReceived > size) {
+      if (result > 0L && bytesReceived > size) {
+        // If we received bytes beyond the limit, don't return them to the caller.
+        sink.truncateToSize(sink.size - (bytesReceived - size))
+      }
+      throw IOException("expected $size bytes but got $bytesReceived")
```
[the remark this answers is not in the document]
swankjesse: I don’t think so? The numbers will tell the story

okio-zipfilesystem/src/jvmTest/kotlin/okio/zipfilesystem/FixedLengthSourceTest.kt
swankjesse: Good call! Added this fencepost test.

okio-zipfilesystem/build.gradle
```diff
@@ -0,0 +1,33 @@
...
+      dependencies {
+        implementation deps.test.junit
+        implementation deps.test.assertj
+        implementation deps.kotlin.test.jdk
+      }
+    }
+  }
+}
+
+// modify these lines for MANIFEST.MF properties or for specific bnd instructions
```
oldergod, line 25: Sentence

oldergod (approved):

okio-zipfilesystem/build.gradle
swankjesse: Can do. Follow-up!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/FixedLengthSource.kt
```diff
@@ -0,0 +1,77 @@
...
+
+  override fun read(sink: Buffer, byteCount: Long): Long {
+    // Figure out how many bytes to attempt to read.
+    //
+    // If we're truncating, we never attempt to read more than what's remaining.
+    //
+    // Otherwise we expect the underlying source to be exactly the promised size. Read as much as
+    // possible and throw an exception if too many bytes are returned.
+    val toRead = when {
+      bytesReceived > size -> 0L // Already read more than the promised size.
```
yschimke, line 46: Why isn't this an error? Doesn't this state turn a presumed while loop in a caller into an infinite loop? e.g. If there are multiple concurrent threads calling this class?

okio-zipfilesystem/build.gradle
swankjesse: The follow-up got rid of this comment! https://github.com/square/okio/pull/921/files#

PR #895, opened 2021-02-12
Import ZIP code from AOSP and use it to build a ZipFileSystem

swankjesse:
This is a start.

### DIFF
--- a/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
+++ b/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
@@ -1,0 +1,136 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio.zipfilesystem
+
+import okio.ByteString
+import java.util.Calendar
+import java.util.GregorianCalendar
+
+/**
+ * An entry within a zip file.
+ *
+ * An entry has attributes such as its name (which is actually a path) and the uncompressed size
+ * of the corresponding data. An entry does not contain the data itself, but can be used as a key
+ * with [ZipFileSystem.source].
+ */
+internal class ZipEntry(
+  /**
+   * The name is actually a path and may contain `/` characters.
+   *
+   * *Security note:* Entry names can represent relative paths. `foo/../bar` or
+   * `../bar/baz`, for example. If the entry name is being used to construct a filename
+   * or as a path component, it must be validated or sanitized to ensure that files are not
+   * written outside of the intended destination directory.
+   *
+   * At most 0xffff bytes when encoded.
+   */
+  val name: String,
+
+  /**
+   * Returns the comment for this `ZipEntry`, or `null` if there is no comment.
+   * If we're reading a zip file using `ZipInputStream`, the comment is not available.
+   */
+  val comment: String,
+
+  /**
+   * Gets the checksum for this `ZipEntry`.
+   *
+   * Needs to be a long to distinguish -1 ("not set") from the 0xffffffff CRC32.
+   *
+   * @return the checksum, or -1 if the checksum has not been set.
+   *
+   * @throws IllegalArgumentException if `value` is < 0 or > 0xFFFFFFFFL.
+   */
+  val crc: Long,
+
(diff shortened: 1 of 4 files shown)


swankjesse:
**The code is not original.** I started with Apache-licensed code imported from AOSP and converted it to Kotlin, then removed things we don’t need for a read-only filesystem API.

Upstream:
https://android.googlesource.com/platform/libcore/+/android-6.0.1_r21/luni/src/main/java/java/util/zip/Zip64.java
https://android.googlesource.com/platform/libcore/+/android-6.0.1_r21/luni/src/main/java/java/util/zip/ZipEntry.java
https://android.googlesource.com/platform/libcore/+/android-6.0.1_r21/luni/src/main/java/java/util/zip/ZipFile.java

swankjesse:
There’s still lots of work to do on this.

Egorand (approved):

oldergod (commented):
🚀

yschimke:
Nice alternative to address issues like https://twitter.com/headius/status/1269077314677035008 https://bugs.openjdk.java.net/browse/JDK-8246714

"URLClassLoader/JAR protocol handler poisons the global JarFile cache under concurrent load"

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
```diff
@@ -0,0 +1,136 @@
...
+   * `../bar/baz`, for example. If the entry name is being used to construct a filename
+   * or as a path component, it must be validated or sanitized to ensure that files are not
+   * written outside of the intended destination directory.
+   *
+   * At most 0xffff bytes when encoded.
+   */
+  val name: String,
+
+  /**
+   * Returns the comment for this `ZipEntry`, or `null` if there is no comment.
```
yschimke, line 44: the val appears non-null?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
```diff
@@ -0,0 +1,484 @@
...
+  val entries = mutableMapOf<Path, ZipEntry>()
+  for (i in 0 until record.numEntries) {
+    val newEntry = source.readEntry(
+      isZip64 = zip64EocdRecordOffset != -1L
+    )
+    if (newEntry.localHeaderRelOffset >= record.centralDirOffset) {
+      throw IOException("Local file header offset is after central directory")
+    }
+    val entryName = newEntry.name
+    val path = "/".toPath().div(entryName)
```
yschimke, line 187: why .div() here?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipFileSystem.kt
```diff
@@ -0,0 +1,115 @@
...
+    return when (entry.compressionMethod) {
+      ZipEntry.STORED -> {
+        FixedLengthSource(source, entry.size, truncate = true)
+      }
+      else -> {
+        val inflaterSource = InflaterSource(
+          FixedLengthSource(source, entry.compressedSize, truncate = true),
+          Inflater(true)
+        )
+        FixedLengthSource(inflaterSource, entry.size, truncate = false)
```
yschimke, line 98: I understand about truncated the compressed size so you don't overrun, but why truncate = false here?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipFileSystem.kt
```diff
@@ -0,0 +1,115 @@
...
+    return entries.keys.mapNotNull { if (it.parent == canonicalDir) it else null }
+  }
+
+  @Throws(IOException::class)
+  override fun source(path: Path): Source {
+    // Make sure this ZipEntry is in this Zip file.  We run it through the name lookup.
+    val canonicalPath = canonicalize(path)
+    val entry = entries[canonicalPath] ?: throw FileNotFoundException("no such file $path")
+    val source = fileSystem.source(zipPath).buffer()
+    val cursor = source.cursor()!!
```
yschimke, line 62: I assume a nicer error to come as a follow up.

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipFileSystem.kt
```diff
@@ -0,0 +1,115 @@
...
+    source.skip((fileNameLength + extraFieldLength).toLong())
+
+    return when (entry.compressionMethod) {
+      ZipEntry.STORED -> {
+        FixedLengthSource(source, entry.size, truncate = true)
+      }
+      else -> {
+        val inflaterSource = InflaterSource(
+          FixedLengthSource(source, entry.compressedSize, truncate = true),
+          Inflater(true)
```
yschimke, line 96: For future - what do some typical workloads look like for processing ZipFiles? Do you plan to later reset and reuse Inflater instances, or is it cheap enough to create as long as you end() them?

yschimke: Thinking about this, what is particularly nice about this file system is that it makes the lifecycle of the ZipFileSystem first class.  Potentially allowing for some closable, or cacheable/poolable, or singleton filesystems.  Compared to the implicit native filesystem or crazy world of JarURLConnection implicit caching etc.

yschimke:
@swankjesse Can I assume the KMM tweaks for this is are in flight? This is going to be great.

PR #897, opened 2021-02-14
Base read only filesystem

yschimke:
Split out a read only filesystem, as two I was interested in looking at would be read only.


yschimke:
undesired because of type system indicating a runtime capability (read only status).

PR #898, opened 2021-02-15
Better tests for zip files

swankjesse:
I've decided that I can't use Java's built-in zip APIs to create complete-enough
zip files. These APIs target a subset of the zip spec and can't create data rich
enough for a good test suite.

In follow-up it might be useful to commit these zip files rather than creating
them on-demand. That approach should also make it possible to test Windows-originated
zip files on UNIX and vice-versa.

### DIFF
--- a/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
+++ b/okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
@@ -17,6 +17,8 @@
 package okio.zipfilesystem
 
 import okio.ByteString
+import okio.ExperimentalFileSystem
+import okio.Path
 import java.util.Calendar
 import java.util.GregorianCalendar
 
@@ -27,6 +29,7 @@ import java.util.GregorianCalendar
  * of the corresponding data. An entry does not contain the data itself, but can be used as a key
  * with [ZipFileSystem.source].
  */
+@ExperimentalFileSystem
 internal class ZipEntry(
   /**
    * The name is actually a path and may contain `/` characters.
@@ -38,13 +41,20 @@ internal class ZipEntry(
    *
    * At most 0xffff bytes when encoded.
    */
-  val name: String,
+  val canonicalPath: Path,
+
+  /**
+   * Determine whether or not this `ZipEntry` is a directory.
+   *
+   * @return `true` when this `ZipEntry` is a directory, `false` otherwise.
+   */
+  val isDirectory: Boolean = false,
 
   /**
    * Returns the comment for this `ZipEntry`, or `null` if there is no comment.
    * If we're reading a zip file using `ZipInputStream`, the comment is not available.
    */
-  val comment: String,
+  val comment: String = "",
 
   /**
    * Gets the checksum for this `ZipEntry`.
@@ -55,7 +65,7 @@ internal class ZipEntry(
    *
    * @throws IllegalArgumentException if `value` is < 0 or > 0xFFFFFFFFL.
    */
-  val crc: Long,
+  val crc: Long = -1L,
 
   /**
    * Gets the compressed size of this `ZipEntry`.
@@ -63,7 +73,7 @@ internal class ZipEntry(
    * @return the compressed size, or -1 if the compressed size has not been
    * set.
    */
-  val compressedSize: Long,
+  val compressedSize: Long = -1L,
 
   /**
    * Gets the uncompressed size of this `ZipEntry`.
@@ -71,7 +81,7 @@ internal class ZipEntry(
    * @return the uncompressed size, or `-1` if the size has not been
    * set.
    */
-  val size: Long,
+  val size: Long = -1L,
 
   /**
    * Gets the compression method for this `ZipEntry`.
@@ -79,11 +89,11 @@ internal class ZipEntry(
    * @return the compression method, either `DEFLATED`, `STORED`
    * or -1 if the compression method has not been set.
    */
-  val compressionMethod: Int,
+  val compressionMethod: Int = -1,
 
-  val time: Int,
+  val time: Int = -1,
 
-  val modDate: Int,
+  val modDate: Int = -1,
 
   /**
    * Gets the extra information for this `ZipEntry`.
@@ -91,10 +101,12 @@ internal class ZipEntry(
    * @return a byte array containing the extra information, or `null` if
    * there is none.
    */
-  val extra: ByteString,
+  val extra: ByteString = ByteString.EMPTY,
 
-  val localHeaderRelOffset: Long
+  val localHeaderRelOffset: Long = -1L
 ) {
+  val children = mutableListOf<Path>()
+
   /**
    * Gets the last modification time of this `ZipEntry`.
    *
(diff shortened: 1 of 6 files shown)


okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
```diff
@@ -175,22 +175,60 @@ fun open(zipPath: Path, fileSystem: FileSystem): ZipFileSystem {
...
+  return ZipFileSystem(zipPath, fileSystem, index, comment)
+}
+
+/**
+ * Returns a map containing all of [entries], plus parent entries required so that all entries
+ * (other than the file system root `/`) have a parent.
+ */
+@ExperimentalFileSystem
+private fun buildIndex(entries: List<ZipEntry>): Map<Path, ZipEntry> {
+  val result = mutableMapOf<Path, ZipEntry>()
```
pyricau, line 200: Use LinkedHashMap and provide a capacity of `entries.size()` ?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
```diff
@@ -175,22 +175,60 @@ fun open(zipPath: Path, fileSystem: FileSystem): ZipFileSystem {
...
+
+  // Iterate in sorted order so each path is preceded by its parent.
+  for (entry in entries.sortedBy { it.canonicalPath }) {
```
pyricau, line 203: Sort based on the number of slashes instead? depth will guarantee the order you went and this likely reduces the amount to sort?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
```diff
@@ -38,13 +41,20 @@ internal class ZipEntry(
...
+   * @return `true` when this `ZipEntry` is a directory, `false` otherwise.
+   */
+  val isDirectory: Boolean = false,
 
   /**
    * Returns the comment for this `ZipEntry`, or `null` if there is no comment.
    * If we're reading a zip file using `ZipInputStream`, the comment is not available.
    */
-  val comment: String,
+  val comment: String = "",
```
yschimke, line 57: still non nullable while comment indicates null is no comment.

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
```diff
@@ -55,46 +65,48 @@ internal class ZipEntry(
    *
    * @throws IllegalArgumentException if `value` is < 0 or > 0xFFFFFFFFL.
```
yschimke, line 66: Is this still correct?

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
```diff
@@ -55,46 +65,48 @@ internal class ZipEntry(
...
    */
-  val crc: Long,
+  val crc: Long = -1L,
 
   /**
    * Gets the compressed size of this `ZipEntry`.
    *
    * @return the compressed size, or -1 if the compressed size has not been
```
yschimke, line 73: nit: inconsistent use of -1 vs `-1`

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
```diff
@@ -55,46 +65,48 @@ internal class ZipEntry(
...
+  val modDate: Int = -1,
 
   /**
    * Gets the extra information for this `ZipEntry`.
    *
    * @return a byte array containing the extra information, or `null` if
    * there is none.
    */
-  val extra: ByteString,
+  val extra: ByteString = ByteString.EMPTY,
```
yschimke, line 104: null again...

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
```diff
@@ -175,22 +175,60 @@ fun open(zipPath: Path, fileSystem: FileSystem): ZipFileSystem {
...
+    if (result.put(entry.canonicalPath, entry) != null) {
+      throw IOException("Duplicate path: ${entry.canonicalPath}")
```
yschimke, line 205: There is a suggestion the base format allows this but java makes it fail

https://stackoverflow.com/a/3113584/1542667

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
swankjesse: On the JVM mutableMapOf() will give me a LinkedHashMap. Pre-sizing it is probably not worth the extra code; this isn’t inner-loop stuff.

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
swankjesse: I don’t think the additional performance is worth the additional code.

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
swankjesse: Yeah, none of these comments make sense; they’re all still inherited from AOSP. I was hoping to go through all of them in a follow-up!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
swankjesse: Probably not!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
swankjesse: I think -1 is nicer in words, even if it’s -1L in code.

swankjesse: Ahh, I see the backticks. Lemme rewrite all these awful comments in a follow up!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/ZipEntry.kt
swankjesse: I’m (foolishly?) preserving as much as possible of the original structure of this code in this PR, in hopes of getting all the tests in place before I do any real cleanup. The code is also pretty clumsy!

okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
swankjesse: Agreed. It’s useful validation to reject it.

yschimke: Postels law seems relevent when consumung externally produced files?

swankjesse: Not in this case. The detection is preventative!

https://androidstudygroup.slack.com/archives/D01M90K55LK/p1613431367008700

swankjesse: Not in this case. The detection is preventative!

https://nakedsecurity.sophos.com/2013/07/10/anatomy-of-a-security-hole-googles-android-master-key-debacle-explained/

oldergod (approved):

PR #900, opened 2021-02-17
Support cursors in FakeFileSystem sources

swankjesse:


oldergod (approved):

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -437,17 +438,51 @@ class FakeFileSystem(
...
+      return size - buffer.size
+    }
+
+    override fun size(): Long {
+      check(!closed) { "closed" }
+      return size
+    }
+
+    override fun seek(position: Long) {
+      check(!closed) { "closed" }
```
yschimke, line 471: The thing I love here about FakeFileSystem is that the default behaviour is the most strict (allowReadsWhileWriting) which will flush out issues in testing, while being more lenient in production.  

Reviewing this code, I'm less certain whether I should worry about thread concurrency.

Should by default a FakeFileSource be locked to a single thread by default, simplifying reviewing? Potentially throwing if threads swap, or if we detect concurrency?

yschimke:
Any idea why it's failing on 2x JDK 8 CI builds?

```
okio.SystemFileSystemTest[jvm] > sourceCursorWhenClosed[jvm] FAILED
    java.lang.AssertionError at Assert.java:88
```

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
swankjesse: Unfortunately that'll break producer consumer models where one thread picks files to read and another thread reads em.

But we could add an opt-in for callers that know they don't do that?

strict concurrency()

swankjesse:
I think the exception we're checking for is different on JDK 8.

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
yschimke: My difficulty of the review is how to review whether this code is concurrency safe. Maybe on first read it makes the thread sticky is strictConcurrency is set?  Also in that case which thread creates the source instances?

swankjesse: ZipFileSystem is intended to be immutable once created. Individual sources are not safe for concurrent use.

yschimke (approved):

Issue #877, opened 2021-01-09
okio-testing module with FileSystem support


yschimke:
Feature Request.

For integration with test libraries, cleanup after tests run.  Possibly a FileSystem that can read from classpath, although that might live in Okio itself.

yschimke:
Feature request for classpath access in IDE or builds.

Port https://github.com/cashapp/misk/blob/master/misk-core/src/main/kotlin/misk/resources/ClasspathResourceLoaderBackend.kt
to a readonly file system.

hiding assumptions that your build is src/test/resources or module/src/test/resources, or classpath:/ or class.getResource("xyz")

yschimke:
Thinking more about this, there is actually a subtle difference here we should maintain between

- Runtime resource class path, potentially loading resources out of a binary, even automatic hooks for Graal to be aware of resources
- Build/Test time access to test resources, usually available from the filesystem

yschimke:
I saw mostly read only, because for testing a common pattern would be diffs and rebasing of tests including screenshots etc.

PR #901, opened 2021-02-20
Resource File System

yschimke:
Classpath resources either from project structure or classpath Jar files.


yschimke:
Other benefits

- Paths should remain stable even if the behaviour, api and implementation differ in other systems
- Alternative class path based file systems could exist e.g. Spring with better file lists etc
- Avoids awkward multiple APIs like class.getResource(), class.getResourceAsString(), class.classloader.xxx

yschimke:
Replacement for the complete and utter shed show that is resources, particularly in theory for KMM projects.

yschimke:
https://github.com/square/okio/issues/877

okio-resourcefilesystem/build.gradle
```diff
@@ -0,0 +1,33 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
```
JakeWharton, line 1: Probably can be .jvm, right?

yschimke: Flipping it around, how could I make this for non JVM usage? in builds or in binaries?

Also, any opinions of the utility of this PR? Am I over engineering?

yschimke:
TODO - hook this up to make sure that Java resources survive through the GraalVM nativeImage building. Could that be via detection in an Automatic Feature?

yschimke:
Some thing behind this

- Resource usage shouldn't be part of some implicit global but not always available set of resources.
- The current APIs are annoyingly different from Files, with multiple source paths, jars and test resources.
- The set of resources each library provides could be exposed to downstream libraries as e.g. 
MyFeature.OfflineResources -> ResourceFileSystem
WWW.Site -> ResourceFileSystem ("src/main/resources/site") / NativeFileSystem("public_html")

yschimke:
Question: Should a ResourceFileSystem have an optional prefix root? Left off by applied automatically?

I guess either it's visible in the Path instances, but used for relative resources.
or
It's invisible in the Path instances, and implicitly applied before the call to lookup a resource.

swankjesse:
I’d prefer it be rooted at `/`. It’s very easy to get `ForwardingFileSystem` to remap it if necessary.

yschimke:
> I’d prefer it be rooted at `/`. It’s very easy to get `ForwardingFileSystem` to remap it if necessary.

Ok. But should it have a concept of relative paths?

yschimke:
In this design its deliberate its not an object singleton for the entire process. You have a set of resources you are providing. That could be everything in the classpath.  But hopefully its constrained like a directory of web content shipped in the app.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,96 @@
...
+ */
+@ExperimentalFileSystem
+class ResourceFileSystem(val paths: List<Path>? = null) : ReadOnlyFilesystem() {
+  override fun canonicalize(path: Path): Path {
+    val cleaned = super.canonicalize(path)
+
+    if (paths != null) {
+      // TODO should this fail early on files that don't exist also
+      paths.forEach {
+        val pathBytes = path.toString()
```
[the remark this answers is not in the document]
swankjesse: I think you want `==` on the parent path.

Doing startsWith() won’t work for cases like this:
```
  val path = "/hot
  val cleaned = "/hotdog
```

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,96 @@
...
+    val stream: InputStream = this.javaClass.classLoader.getResourceAsStream(resourceName)
+      ?: throw FileNotFoundException("file not found: $file")
+
+    return stream.source()
+  }
+
+  /**
+   * Return the [SYSTEM] path for a file if it is available.
+   */
+  fun toSystemPath(path: Path): Path? {
```
swankjesse, line 85: Is this dangerous? This is the sort of API that works in tests and breaks in production.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,96 @@
...
+ * A file system exposing the traditional java style resource paths.
+ *
+ * Both metadata and file listings are best effort, and will work better
+ * for local project paths. The file system does not handle merging of
+ * multiple paths from difference resources like overlapping Jar files.
+ *
+ * TODO hook into GrallVM building such that resources remain available
+ */
+@ExperimentalFileSystem
+class ResourceFileSystem(val paths: List<Path>? = null) : ReadOnlyFilesystem() {
```
swankjesse, line 41: I don’t think this should be the public constructor

swankjesse (commented):
I like the feature but there’s a few behaviors I’d prefer to change...

I think we can borrow code from Misk to make `list()` work.
https://github.com/cashapp/misk/blob/master/misk-core/src/main/kotlin/misk/resources/ClasspathResourceLoaderBackend.kt

I don’t quite understand how `canonicalize()` is supposed to work here. I assume that it should be doing is just forcing a leading `/` on the path.

swankjesse:
I really like this! Perhaps the API should be a constant like `SYSTEM` on JVM platforms, and maybe more later?

```
abstract class FileSystem {
  ...
  companion object {
    ...
    /** Returns a file system that reads resources from the current process. */
    val RESOURCES: FileSystem = PLATFORM_FILE_SYSTEM
  }
}
```

swankjesse:
> In this design its deliberate its not an object singleton for the entire process. You have a set of resources you are providing. That could be everything in the classpath. But hopefully its constrained like a directory of web content shipped in the app.

Ahh, I think I misread the purpose of this PR. 

For _that_, I recommend something else. Use `ForwardingFileSystem` to mask another file system to a directory or set of directories that are permitted. You could mask `FileSystem.SYSTEM` to just the `/var/www/html` directory, or mask `FileSystem.RESOURCES` to say, `/public` directory if the web assets are in `src/main/resources/public`.

yschimke:
@swankjesse It would be good to discuss. I started trying to make a file system for class path resources.  Halfway through I decided that was a mistake to correct. And that a global resource filesystem is a worse design based on unpredictable base (graalvm builds, build/test behaviour, IDE).  Mostly what we want is for specific modules of code to define they bring along a set of resources and provide access to those at test and runtime.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
yschimke: +1 hence nullability here. It's basically to allow for optional functionality that is only easy before we bundle the resources in the build into a assembly.  Can remove.

yschimke:
> I like the feature but there’s a few behaviors I’d prefer to change...
> 
> I think we can borrow code from Misk to make `list()` work.
> https://github.com/cashapp/misk/blob/master/misk-core/src/main/kotlin/misk/resources/ClasspathResourceLoaderBackend.kt
> 
> I don’t quite understand how `canonicalize()` is supposed to work here. I assume that it should be doing is just forcing a leading `/` on the path.

Force "/" convention and cleanup paths.

yschimke:
Trying to look at what it would be like in practice (a start, non compiling) look at using this in OkHttp

https://github.com/square/okhttp/pull/6567/files

yschimke:
> I really like this! Perhaps the API should be a constant like `SYSTEM` on JVM platforms, and maybe more later?
> 
> ```
> abstract class FileSystem {
>   ...
>   companion object {
>     ...
>     /** Returns a file system that reads resources from the current process. */
>     val RESOURCES: FileSystem
>   }
> }
> ```

I'm adding this because you are right it makes sense.  I want to make the argument that the fact the Java has a single class path is an implementation detail we have all grown accustomed to.  Not how we should think about this with Okio FileSystem and multi platform code.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,117 @@
...
+
+  override fun metadataOrNull(path: Path): FileMetadata? {
+    val systemPath = toSystemPath(path) ?: throw IOException("metadata for $path not supported")
+
+    // TODO consider unwrapping jar urls
+    return SYSTEM.metadataOrNull(systemPath)
+  }
+
+  override fun source(file: Path): Source {
+    val resourceName = canonicalize(file).toString().substring(1)
```
swankjesse, line 89: Presumably substring(1) is to remove the leading `/`. Is that necessary?

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
swankjesse: Lemme try another approach: what’s the use case for this?

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,117 @@
...
+      }
+
+      throw IOException("Requested path $path is not within resource filesystem $paths")
+    }
+
+    return cleaned
+  }
+
+  override fun list(dir: Path): List<Path> {
+    val systemPath = toSystemPath(dir) ?: throw IOException("not listable")
```
swankjesse, line 76: We can list the contents of `.jar` files!

swankjesse (commented):
My ideal for this is a a public API consisting of a single new symbol; either like this:

```
object ResourceFileSystem : FileSystem {
  ...
}
```

or like this:

```
val RESOURCES_FILE_SYSTEM : FileSystem = ...
```

We should be able to implement file system filtering or composition in a complementary API.

yschimke:
> My ideal for this is a a public API consisting of a single new symbol; either like this:
> 
> ```
> object ResourceFileSystem : FileSystem {
>   ...
> }
> ```
> 
> or like this:
> 
> ```
> val RESOURCES_FILE_SYSTEM : FileSystem = ...
> ```
> 
> We should be able to implement file system filtering or composition in a complementary API.

OK, I'll post an update with this as a base global filesystem.  Try to pitch the modular resource approach separately.

I still think it's worth doing, essentially as you bring in extra dependencies, they provide APIs exposing the FileSystems with those attached resources rather than implicitly contributing to the global shared classpath.  It also ensures a more DI approach, code doesn't reach into the singleton ResourceFileSystem in the middle of a code block, but must have a clean API to access the resources or else get given the resources to use in prod or testing etc.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * A file system exposing something similar to the traditional java style resources.
+ *
+ * Both metadata and file listings are best effort, and will work better
+ * for local project paths. The file system does not handle merging of
+ * multiple paths from difference resources like overlapping Jar files.
+ *
+ * While a single `val everything = ResourceFileSystem()` close to the callsite would allow
+ * code requiring classpath resources to access the implicit process classpath, this should be
+ * avoided.
```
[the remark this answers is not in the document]
yschimke: That feels like a stretch, the experience to most developers is that the order is ill-defined and can frequently change when there are conflicts, and most developers are unaware of the totality of the class path resources (ignoring class files). We essentially provide a public copy of the Public Suffix Database in an opaque format to anyone who wants to read it, but we choose to make the related classes internal.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
yschimke: It's related to the shitty nuances of class loading.getResource vs class.getResource https://stackoverflow.com/questions/6608795/what-is-the-difference-between-class-getresource-and-classloader-getresource

I'll double check if I can leave it.

okio-resourcefilesystem/src/jvmMain/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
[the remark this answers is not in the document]
yschimke: > My ideal for this is a a public API consisting of a single new symbol; either like this:
> 
> ```
> object ResourceFileSystem : FileSystem {
>   ...
> }
> ```
> 
> or like this:
> 
> ```
> val RESOURCES_FILE_SYSTEM : FileSystem = ...
> ```
> 
> We should be able to implement file system filtering or composition in a complementary API.

This even worries me if someone has something like a container process (tomcat etc).  And two overlapping war files, the scope of ResourceFileSystem seems tied to how okio-resourcefilesystem is loaded, while by splitting up we could let each Jar provide it's partial view of class path resources just for what it knows it provides.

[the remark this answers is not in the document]
yschimke: > I don’t think I follow this advice. We don’t offer a similar warning for FileSystem.SYSTEM.

I think we should also encourage not implicitly assuming paths in code relate to SYSTEM is a good thing. Will make tests and even abstractions like Cache a lot better if we encourage passing in the FileSystem as a dependency

PR #903, opened 2021-02-21
Redo the zip.kt internals

swankjesse:
Use idiomatic Okio.

Remove unnecessary fields, features, comments.

Remove some unnecessary seeks.


okio-zipfilesystem/src/jvmMain/kotlin/okio/zipfilesystem/zip.kt
```diff
@@ -231,298 +175,211 @@ private fun buildIndex(entries: List<ZipEntry>): Map<Path, ZipEntry> {
...
-  for (element in name) {
-    if (element == '\u0000') throw IOException("Filename contains NUL byte")
+  skip(8) // disk number start (2) + internal file attributes (2) + external file attributes (4).
+  var offset = readIntLe().toLong() and 0xffffffffL
+  val name = readUtf8(nameSize.toLong())
+  if ('\u0000' in name) throw IOException("bad zip: filename contains 0x00")
+
+  val requiredZip64ExtraSize = (if (size == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) 8L else 0L) +
+    (if (compressedSize == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) 8L else 0L) +
+    (if (offset == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) 8L else 0L)
```
oldergod, line 216: Could we make that more readable? Something like that maybe
```suggestion
  val requiredZip64ExtraSize = run {
    var result = 0L
    if (size == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) result += 8
    if (compressedSize == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) result += 8
    if (offset == MAX_ZIP_ENTRY_AND_ARCHIVE_SIZE) result += 8
    return@run result
  }
```

oldergod (approved):

yschimke:
This bloodbath of a red PR is a good indication.  nice!

PR #907, opened 2021-03-06
Fix FakeFileSystem.list() to return relative paths

swankjesse:
We were previously returning absolute paths always, even when the
input directory was a relative path.

### DIFF
--- a/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
+++ b/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
@@ -228,6 +228,11 @@ class FakeFileSystem(
 
     element.access(now = clock.now())
     val paths = elements.keys.filterTo(mutableListOf()) { it.parent == canonicalPath }
+    if (dir.isRelative) {
+      for (i in paths.indices) {
+        paths[i] = dir / paths[i].name
+      }
+    }
     paths.sort()
     return paths
   }
--- a/okio/src/commonMain/kotlin/okio/FileSystem.kt
+++ b/okio/src/commonMain/kotlin/okio/FileSystem.kt
@@ -133,7 +133,8 @@ abstract class FileSystem {
 
   /**
    * Returns the children of the directory identified by [dir]. The returned list is sorted using
-   * natural ordering.
+   * natural ordering. If [dir] is a relative path, the returned elements will also be relative
+   * paths. If it is an absolute path, the returned elements will also be absolute paths.
    *
    * @throws IOException if [dir] does not exist, is not a directory, or cannot be read. A directory
    *     cannot be read if the current process doesn't have access to [dir], or if there's a loop of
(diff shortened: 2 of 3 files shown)


oldergod (approved):

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -228,6 +228,9 @@ class FakeFileSystem(
 
     element.access(now = clock.now())
     val paths = elements.keys.filterTo(mutableListOf()) { it.parent == canonicalPath }
+    for (i in paths.indices) {
```
yschimke, line 231: Since workingDirectory is known to be absolute, it seems only worthwhile doing this if dir.isRelative == true.

swankjesse: As a performance optimization? Or for different behavior?

yschimke: Performance optimisation, but the good kind, skipping work when you know it isn't needed :)

Working on ResourceFileSystem made me aware of how frequently we end up doing path comparisons in tight loops.  

Some thoughts, outside of this PR which is after all a Fake FileSystem.

1) They aren't particularly optimised since ultimately if length matches they become a front to back array comparison in ByteString. And given properties of a file system, the prefix is much more likely (not guaranteed) to match than suffix. What's the plan here? optimise each case like ZipFileSystem's fully populated index.
2) Should we use the cached hashCode of ByteString to avoid this work.  It might also minimise (poorly) string length oracle attacks.
3) We are getting into the territory of us implementing a lot of [accidentally quadratic](https://accidentallyquadratic.tumblr.com) library functions.

I still remember the terrible Java 1.0.2? string hashcode that skipped characters and was pathologically bad on structured filesystems (think a strict naming system for camera images).

swankjesse: Implemented the performance optimization.

Using the cached hash code in equals is potentially tricky. We won’t always have a hash code for one or both strings, so short-circuiting on it only works if we have it for both of ’em. (I wonder how often we’ll have a hash code for both strings?!)

As for accidentally quadratic things . . . I think the biggest danger we face with performance surprises is in Buffers and internal segments. I’ve tried to make it so segments are 50% full or better whenever possible; this is partially to avoid quadratic hazards.

yschimke: Yeah, most of this was me planting the thought. Since it's been bugging me. I've seen weird cases when Intellij freezes (10-15 seconds) turned out to be assumptions made in jetbrains library code that broke on FB or Google size codebase. 

On a large directory, with a very long prefix...

Nothing urgent.

PR #909, opened 2021-03-07
Implement cursoring on JS sources

swankjesse:


okio/src/jsMain/kotlin/okio/FileSource.kt
```diff
@@ -33,11 +34,13 @@ internal class FileSource(
       buffer = data,
       length = byteCount,
       offset = 0,
-      position = null
+      position = position_.toDouble()
```
yschimke, line 37: Why double here?

swankjesse: In JS it's declared to accept `Number` but I found it didn't work properly until I manually converted to JS’ only native number type.

oldergod (approved):

PR #913, opened 2021-03-31
Move NodeJsFileSystem into its own module

swankjesse:
This removes SYSTEM as a statically-available FileSystem instance
on the JS platform, and converts FileSystem into an expect class.


okio-nodefilesystem/build.gradle
```diff
@@ -0,0 +1,58 @@
+apply plugin: 'org.jetbrains.kotlin.multiplatform'
```
JakeWharton, line 1: Should probably be a `org.jetbrains.kotlin.js` module

okio-nodefilesystem/build.gradle
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+
+// modify these lines for MANIFEST.MF properties or for specific bnd instructions
+project.ext.bndManifest = '''
+    Export-Package: okio.nodefilesystem
+    Automatic-Module-Name: okio.nodefilesystem
+    Bundle-SymbolicName: com.squareup.okio.nodefilesystem
+    '''
```
JakeWharton, line 56: Can be deleted. No OSGi nonsense in the leftpad ecosystem.

JakeWharton (approved):

okio-nodefilesystem/build.gradle
swankjesse: NICE yes!

okio-nodefilesystem/build.gradle
swankjesse: Left as MPP for now, only because I don’t know the syntax.

okio-nodefilesystem/gradle.properties
```diff
@@ -0,0 +1,2 @@
+POM_ARTIFACT_ID=okio-nodefilesystem
+POM_NAME=Okio Fake Node.js System
```
yschimke, line 2: why Fake?

swankjesse: Ummm copy paste error! Good catch! Will fix.

okio-nodefilesystem/build.gradle
JakeWharton: Tracking at #914

PR #916, opened 2021-04-03
FileHandle, for reading and writing, streaming and random access

swankjesse:
I'm hoping to get something close to the underlying file descriptor or file
handle abstraction in POSIX.

I am unsatisfied with the design of okio.Cursor and how it will work with
random-access reading and writing. I'm hoping to make this the successor to
that class, and to completely delete that class from our public API.

I am planning a follow up PR on FileSystem, FileSystem.open(Path) that
returns a FileHandle. I will also make FileSystem.source() and
FileSystem.sink() all delegate to FileSystem.open().

### DIFF
--- a/okio/src/commonMain/kotlin/okio/FileHandle.kt
+++ b/okio/src/commonMain/kotlin/okio/FileHandle.kt
@@ -1,0 +1,219 @@
+/*
+ * Copyright (C) 2021 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+/**
+ * An open file for reading and writing; using either streaming and random access.
+ *
+ * Use [read] and [write] to perform one-off random-access reads and writes. Use [source], [sink],
+ * and [appendingSink] for streaming reads and writes.
+ *
+ * File handles must be closed when they are no longer needed. It is an error to read, write, or
+ * create streams after a file handle is closed. The operating system resources held by a file
+ * handle will be released once the file handle **and** all of its streams are closed.
+ *
+ * Although this class offers both reading and writing APIs, file handle instances may be
+ * read-only or write-only. For example, a handle to a file on a read-only file system will throw an
+ * exception if a write is attempted.
+ *
+ * File handles may be used by multiple threads concurrently. But the individual sources and sinks
+ * produced by a file handle are not safe for concurrent use.
+ */
+@ExperimentalFileSystem
+abstract class FileHandle : Closeable {
+  /**
+   * True once the file handle is closed. Resources should be released with [closeInternal] once
+   * this is true and [openStreamCount] is 0.
+   */
+  private var closed = false
+
+  /**
+   * Reference count of the number of open sources and sinks on this file handle. Resources should
+   * be released with [closeInternal] once this is 0 and [closed] is true.
+   */
+  private var openStreamCount = 0
+
+  /**
+   * Removes at least 1, and up to [byteCount] bytes from this and appends them to [sink]. Returns
+   * the number of bytes read, or -1 if this file is exhausted.
+   */
+  @Throws(IOException::class)
+  abstract fun read(offset: Long, sink: Buffer, byteCount: Long): Long
+
+  /**
+   * Returns the total number of bytes in the file. This will change if the file size changes.
+   */
+  @Throws(IOException::class)
+  abstract fun size(): Long
+
+  /** Removes [byteCount] bytes from [source] and writes them to this at [offset]. */
+  @Throws(IOException::class)
+  abstract fun write(offset: Long, source: Buffer, byteCount: Long)
+
+  /** Pushes all buffered bytes to their final destination. */
+  @Throws(IOException::class)
+  abstract fun flush()
+
+  /**
+   * Returns a source that reads from this starting at [offset]. The returned source must be closed
+   * when it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun source(offset: Long = 0L): Source {
+    synchronized(this) {
+      check(!closed) { "closed" }
+      openStreamCount++
+    }
+    return FileHandleSource(this, offset)
+  }
+
+  /**
+   * Returns the position of [source] in the file. The argument [source] must be either a source
+   * produced by this file handle, or a [BufferedSource] that directly wraps such a source. If the
+   * parameter is a [BufferedSource], it adjusts for buffered bytes.
+   */
+  @Throws(IOException::class)
+  fun position(source: Source): Long {
+    var source = source
+    var bufferSize = 0L
+
+    if (source is RealBufferedSource) {
+      bufferSize = source.buffer.size
+      source = source.source
+    }
+
+    require(source is FileHandleSource && source.fileHandle === this) {
+      "source was not created by this FileHandle"
+    }
+
+    return source.position - bufferSize
+  }
+
+  /**
+   * Returns a sink that writes to this starting at [offset]. The returned sink must be closed when
+   * it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun sink(offset: Long = 0L): Sink {
+    synchronized(this) {
+      check(!closed) { "closed" }
+      openStreamCount++
+    }
+    return FileHandleSink(this, offset)
+  }
+
+  /**
+   * Returns a sink that writes to this starting at the end. The returned sink must be closed when
+   * it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun appendingSink(): Sink {
+    return sink(size())
+  }
+
+  /**
+   * Returns the position of [sink] in the file. The argument [sink] must be either a sink produced
+   * by this file handle, or a [BufferedSink] that directly wraps such a sink. If the parameter is a
+   * [BufferedSink], it adjusts for buffered bytes.
+   */
+  @Throws(IOException::class)
+  fun position(sink: Sink): Long {
+    var sink = sink
+    var bufferSize = 0L
+
+    if (sink is RealBufferedSink) {
+      bufferSize = sink.buffer.size
+      sink = sink.sink
+    }
+
+    require(sink is FileHandleSink && sink.fileHandle === this) {
+      "sink was not created by this FileHandle"
+    }
+
+    return sink.position + bufferSize
+  }
+
+  @Throws(IOException::class)
+  override fun close() {
+    synchronized(this) {
+      if (closed) return@close
+      closed = true
+      if (openStreamCount != 0) return@close
+    }
+    closeInternal()
+  }
+
+  /**
+   * Subclasses should implement this to release resources held by this file handle. It is invoked
+   * once both the file handle is closed, and also all sources and sinks produced by it are also
+   * closed.
+   */
+  @Throws(IOException::class)
+  protected abstract fun closeInternal()
+
+  private class FileHandleSink(
+    val fileHandle: FileHandle,
+    var position: Long
+  ) : Sink {
+    var closed = false
+
+    override fun write(source: Buffer, byteCount: Long) {
+      fileHandle.write(position, source, byteCount)
+      position += byteCount
+    }
+
+    override fun flush() {
+      fileHandle.flush()
+    }
+
+    override fun timeout() = Timeout.NONE
+
+    override fun close() {
+      synchronized(fileHandle) {
+        if (closed) return@close
+        closed = true
+        fileHandle.openStreamCount--
+        if (fileHandle.openStreamCount != 0 || !fileHandle.closed) return@close
+      }
+      fileHandle.closeInternal()
+    }
+  }
+
+  private class FileHandleSource(
+    val fileHandle: FileHandle,
+    var position: Long
+  ) : Source {
+    var closed = false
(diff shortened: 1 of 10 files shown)


okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+   * be released with [closeInternal] once this is 0 and [closed] is true.
+   */
+  private var openStreamCount = 0
+
+  /**
+   * Removes at least 1, and up to [byteCount] bytes from this and appends them to [sink]. Returns
+   * the number of bytes read, or -1 if this file is exhausted.
+   */
+  @Throws(IOException::class)
+  abstract fun read(offset: Long = 0L, sink: Buffer, byteCount: Long): Long
```
yschimke, line 54: Is there any alternative to Buffer here as the target that would allow for optimised operations internally within the filesystem implementation?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+  private var closed = false
+
+  /**
+   * Reference count of the number of open sources and sinks on this file handle. Resources should
+   * be released with [closeInternal] once this is 0 and [closed] is true.
+   */
+  private var openStreamCount = 0
+
+  /**
+   * Removes at least 1, and up to [byteCount] bytes from this and appends them to [sink]. Returns
```
yschimke, line 50: I wonder if there is any annotation to use here that you can't ignore the return value?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+   * the number of bytes read, or -1 if this file is exhausted.
+   */
+  @Throws(IOException::class)
```
yschimke, line 53: What happens with partial and likely interrupted or failed reads?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    override fun flush() {
+      fileHandle.flush()
+    }
+
+    override fun timeout() = Timeout.NONE
+
+    override fun close() {
+      synchronized(this) {
```
yschimke, line 185: Is synchronizing on the sink here correct? Couldn't that race if two sinks or sources are closed at same time?

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: If we wanna do such optimizations, we can change buffer to support ’em!

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Yeah, it’s `CheckReturnValue`. This is the same as the signature of Source.read().

(Follow-up?)

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Yeah, we’re already potentially problematic here.

Tracking as: https://github.com/square/okio/issues/917

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Totally incorrect. Great catch.

swankjesse: Fixed to synchronize on `fileHandle`.

Great catch. Hooray for your careful code review.

okio/src/commonMain/kotlin/okio/FileHandle.kt
yschimke: I was wondering what a file to file direct copy would look like? Operations like https://docs.oracle.com/javase/7/docs/api/java/nio/channels/FileChannel.html#transferFrom(java.nio.channels.ReadableByteChannel,%20long,%20long)

yschimke: I wonder whether there is a good set of common operations to support? https://www.baeldung.com/java-filechannel

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+    return FileHandleSource(this, offset)
+  }
+
+  /**
+   * Returns the position of [source] in the file. The argument [source] must be either a source
+   * produced by this file handle, or a [BufferedSource] that directly wraps such a source. If the
+   * parameter is a [BufferedSource], it adjusts for buffered bytes.
+   */
+  @Throws(IOException::class)
+  fun position(source: Source): Long {
```
bnorm, line 89: What's the benefit of this API over exposing `FileSystemSource` as a public type with a public `position` property? This could get awkward if doing more than buffering the source/sink, for example, hashing to validate a checksum.

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+    var source = source
+    var bufferSize = 0L
+
+    if (source is RealBufferedSource) {
```
bnorm, line 93: There's no protection from double-buffering a source or sink. Should this be a `while`? Or should that protection be added in a separate PR?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+
+  /**
+   * Returns the total number of bytes in the file. This will change if the file size changes.
+   */
+  @Throws(IOException::class)
+  abstract fun size(): Long
+
+  /** Removes [byteCount] bytes from [source] and writes them to this at [offset]. */
+  @Throws(IOException::class)
+  abstract fun write(offset: Long = 0L, source: Buffer, byteCount: Long)
```
bnorm, line 64: Is it going to be common enough to repeatedly read from/write to the beginning of a file such that there is a benefit to having a default value for offset? Would it be better without to avoid any confusion between this and Source/Sink style reading/writing? It's probably pretty low risk though so this might be bikeshedding territory.

oldergod (approved):

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: That’s a good list. I think probably these:

- Reading
- Writing
- Current Position
- Get file size
- Set file size (ie. truncate)
- Force data
- Force metadata

And not these:
 - Memory mapping
 - Locks

These last two are exotic enough that omitting them shouldn’t be missed by most developers.

swankjesse: As for direct filesystem transfer, that’s a neat optimization and it’s possible our APIs don’t preclude them!

But it’s a difficult API to use because it’s easy to fall off the fast path, and there’s no way to make sure that you’re on the fast path.

I’d prefer to omit that.

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Yeah, you’re absolutely right. I definitely wrote that out and tried to see how it’d play out. The biggest drawback of it is how it interacts with `BufferedSource`.

If we introduce a new type then `BufferedSource` either won’t implement the interface, or other `BufferedSource` implementations will implement the interface but not be able to do anything reasonable in this method.

The use case I’m imagining for this is passing a BufferedSource to a method to read some data, then checking the new position.

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: I contemplated that too. I think this API is too fragile to be used dynamically like that. This API isn’t so flexible to take an arbitrary transformation of a source; just the source it created and a direct wrapping of it.

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Great point. I think you’re right; the default here doesn’t make sense. Removed!

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+
+  /**
+   * Subclasses should implement this to release resources held by this file handle. It is invoked
+   * once both the file handle is closed, and also all sources and sinks produced by it are also
+   * closed.
+   */
+  @Throws(IOException::class)
+  protected abstract fun closeInternal()
+
+  private class FileHandleSink(
```
yschimke, line 167: "inner class" skip the val constructor param?

okio/src/commonMain/kotlin/okio/FileHandle.kt
yschimke: It's already really brittle here, the type system is not protecting you here.  So it pretty much needs to be the source that you immediately called buffer on and never let out of your sight.

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: I will, if you can help me figure out how to make this work in the enclosing class:

```
    require(source is FileHandleSource && source.fileHandle === this) {
      "source was not created by this FileHandle"
    }
```

oldergod: It could work if you did something like this I think.
```kotlin
  inner class FileHandleSink(
    var position: Long
  ) : Sink {
    var fileHandle = this@FileHandle
```

But I don't care either way how you end up writing it. I like constructor param.

yschimke: [https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS40LjMwIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsImpzQ29kZSI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiY2xhc3MgQSB7XG4gICAgZnVuIGIoKSA9IEIoKVxuICAgIFxuICAgIGlubmVyIGNsYXNzIEIge1xuICAgICAgICBmdW4gcnVuKCkge1xuICAgICAgICAgICAgcHJpbnRsbih0aGlzKVxuICAgICAgICAgICAgcHJpbnRsbih0aGlzQEEpXG4gICAgICAgICAgICBwcmludGxuKHRoaXNAQilcbiAgICAgICAgfVxuICAgIH1cbn1cblxuZnVuIG1haW4oKSB7XG4gICAgQSgpLmIoKS5ydW4oKVxufSJ9](play.kotlinlang.org)

```
class A {
    fun b() = B()
    
    inner class B {
        fun run() {
            println(this)
            println(this@A)
            println(this@B)
        }
    }
}

fun main() {
    A().b().run()
}
```

swankjesse: How about this one?
https://pl.kotl.in/GiG6aSoGI

yschimke: Yeah, your solution seems cleaner.

okio/src/commonMain/kotlin/okio/FileHandle.kt
yschimke: Maybe add another case which is dropping into native APIs.  Make sure that you design for that case.  Here is how to bridge to java.nio, or node filesystem.

yschimke: To be clear, I'm not pushing for where the line is, just being clear around where it is and an image in your mind that it will generally work for typical client use cases, combined with the types of FileSystems you see implemented.

okio/src/commonMain/kotlin/okio/FileHandle.kt
yschimke: I'm now worried about this breaking with even a simple decorated file system, like logging.

swankjesse: When it implements `position` it'll need to delegate with the unwrapped instance.

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+  @Throws(IOException::class)
+  abstract fun flush()
+
+  /**
+   * Returns a source that reads from this starting at [offset]. The returned source must be closed
+   * when it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun source(offset: Long = 0L): Source {
+    synchronized(this) {
```
JakeWharton, line 76: what in the magic hell synchronized is this that it works (correctly) from common code? shouldn't we be using atomicfu?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+  /** Pushes all buffered bytes to their final destination. */
+  @Throws(IOException::class)
+  abstract fun flush()
+
+  /**
+   * Returns a source that reads from this starting at [offset]. The returned source must be closed
+   * when it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun source(offset: Long = 0L): Source {
```
JakeWharton, line 75: jvm overloads? just in case!

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -0,0 +1,219 @@
...
+
+    return source.position - bufferSize
+  }
+
+  /**
+   * Returns a sink that writes to this starting at [offset]. The returned sink must be closed when
+   * it is no longer needed.
+   */
+  @Throws(IOException::class)
+  fun sink(offset: Long = 0L): Sink {
```
JakeWharton, line 110: jvm overloads? just in case!

JakeWharton (commented):
👍

PR #925, opened 2021-04-24
Change FileHandle's lowest-abstraction to be a ByteArray

swankjesse:
Using a Buffer for low-level reads had two problems

 - Annoying for low-level OS implementations to implement
 - Prevent use by applications that want to read an entire file
   into a single massive array, bypassing our Buffers


okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -445,93 +462,90 @@ class FakeFileSystem(
...
       if (closed) return
       closed = true
-      file.data = buffer.snapshot()
-      file.access(now = clock.now(), modified = true)
+      file.access(now = clock.now(), modified = writeAccess)
       openFiles -= openFile
     }
 
-    override fun toString() = "sink(${openFile.canonicalPath})"
+    override fun toString() = "FileHandler(${openFile.canonicalPath})"
```
yschimke, line 548: Why not FakeFileHandle?

okio-resourcefilesystem/src/main/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
```diff
@@ -57,10 +57,21 @@ class ResourceFileSystem internal constructor(
...
+    write: Boolean
+  ): FileHandle {
     val (fileSystem, fileSystemPath) = toSystemPath(file)
       ?: throw FileNotFoundException("file not found: $file")
-    return fileSystem.open(fileSystemPath)
+    return fileSystem.open(
+      file = fileSystemPath,
+      read = read,
+      write = write,
+      createIfAbsent = createIfAbsent,
```
yschimke, line 71: How does this work?

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -145,7 +145,11 @@ expect abstract class FileSystem() {
    *     if there's a loop of symbolic links, or if any name is too long.
    */
   @Throws(IOException::class)
-  abstract fun open(file: Path): FileHandle
+  abstract fun open(
+    file: Path,
+    read: Boolean = false,
```
yschimke, line 150: Do you want @JvmOverloads for this?

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -35,49 +35,102 @@ package okio
...
+  fun read(
+    fileOffset: Long,
+    array: ByteArray,
+    arrayOffset: Int,
+    byteCount: Int
+  ): Int {
+    synchronized(this) {
+      check(!closed) { "closed" }
+    }
+    return protectedRead(fileOffset, array, arrayOffset, byteCount)
```
yschimke, line 64: Should you check arrayOffset and byteCount are valid for array? Or trust protectedRead to do that?

oldergod (approved):

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
swankjesse: It’s not a public type name, so it’s a bit weird to show up in toString(). Anyone who wants that information can get it from `.getClass()`.

okio-resourcefilesystem/src/main/kotlin/okio/resourcefilesystem/ResourceFileSystem.kt
swankjesse: It’s in StandardOpenOptions.
https://docs.oracle.com/javase/7/docs/api/java/nio/file/StandardOpenOption.html#CREATE

(I’ve since changed my mind on exposing this, not completely decided either way).

okio/src/commonMain/kotlin/okio/FileSystem.kt
swankjesse: After spending more time with this I think there’s basically two modes:

openReadOnly – file must already exist, it is an error to attempt to write
openReadWrite – file will be created if it doesn’t exist

Thoughts on exposing this rather than having one function with boolean options?

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Yeah, I think that’s protectedRead’s job.

swankjesse:
This PR is superseded by 
https://github.com/square/okio/pull/926

PR #927, opened 2021-04-27
Fold the zipfilesystem and resourcefilesystem into the main okio module

swankjesse:
The amount of extra code in the Okio module is minimal. This should
make it easier to do real-world work with these file systems.


swankjesse:
There’s an important follow-up: the zipfilesystem should be multiplatform!

samples/src/jvmTest/java/okio/samples/ChannelsTest.java
```diff
@@ -84,7 +83,7 @@
   }
 
   @Test public void testReadWriteFile() throws Exception {
-    Path path = temporaryFolder.newFile().toPath();
+    java.nio.file.Path path = temporaryFolder.newFile().toPath();
```
oldergod, line 86: explicit package on purpose? It's only a sample so whatever.
You could also alias import to `NioPath`

oldergod (approved):

samples/src/jvmTest/java/okio/samples/ChannelsTest.java
swankjesse: Tragically no import aliases in `.java` files! I wanted to avoid using `Path` here ’cause locally it looks like our `okio.Path` type.

oldergod: totally missed the elephant in the room.

JakeWharton (approved):

PR #929, opened 2021-05-03
Fix using okio from background thread on Kotlin/Native targets

varahash:
- Marks all top level properties with `@SharedImmutabble` annotation in order to avoid error `kotlin.native.IncorrectDereferenceException: Trying to access top level value not marked as @ThreadLocal or @SharedImmutable from non-main thread
` when calling okio from background thread.
- Configure additional test tasks for native targets to run code in the background ([the same approach](https://github.com/Kotlin/kotlinx.coroutines/blob/1efc9f16a352632e6b632b4daab075ac882e661a/kotlinx-coroutines-core/build.gradle#L87-L102) was taken by kotlin.coroutines project).

### DIFF
--- a/okio/build.gradle.kts
+++ b/okio/build.gradle.kts
@@ -1,4 +1,7 @@
 import aQute.bnd.gradle.BundleTaskConvention
+import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithTests
+import org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType
+import org.jetbrains.kotlin.gradle.plugin.mpp.TestExecutable
 import ru.vyarus.gradle.plugin.animalsniffer.AnimalSnifferExtension
 
 plugins {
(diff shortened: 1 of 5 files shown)


okio/build.gradle.kts
```diff
@@ -210,6 +213,20 @@ kotlin {
       }
     }
   }
+
+  targets.withType<KotlinNativeTargetWithTests<*>> {
+    binaries {
+      // Configure a separate test where code runs in background
+      test("background", setOf(NativeBuildType.DEBUG)) {
+        freeCompilerArgs += listOf("-e", "okio.mainBackground")
```
JakeWharton, line 221: I'm not sure a custom entrypoint is required and instead you can use the `-trw` flag with the regular one.

varahash: Indeed, thanks for the suggestion.

JakeWharton (approved):

okio/build.gradle.kts
```diff
@@ -210,6 +213,20 @@ kotlin {
       }
     }
   }
+
+  targets.withType<KotlinNativeTargetWithTests<*>> {
```
swankjesse, line 217: Neat

swankjesse (approved):

varahash:
@swankjesse Can you, please, merge this fix for alpha.6?

swankjesse:
Of course!

okio/build.gradle.kts
saket: What does `trw` stand for?

Johnett: -generate-worker-test-runner (-trw)﻿
Produce an application for running unit tests in a worker thread.

test run in worker thread.

Johnett: https://kotlinlang.org/docs/compiler-reference.html#generate-test-runner-tr

saket: TIL. Thank you!

PR #939, opened 2021-05-22
Filter .class files from the ResourceFileSystem

swankjesse:
The resource file system could drag a lot of data into memory.
Omitting .class files from the index should save space.

This prevents .class files from being opened directly. I think
that's okay.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/ResourceFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/ResourceFileSystem.kt
@@ -44,7 +44,7 @@ class ResourceFileSystem internal constructor(
 
   override fun list(dir: Path): List<Path> {
     val (fileSystem, fileSystemPath) = toSystemPath(dir) ?: return listOf()
-    return fileSystem.list(fileSystemPath).filterNot { it.name.endsWith(".class") }
+    return fileSystem.list(fileSystemPath)
   }
 
   override fun openReadOnly(file: Path): FileHandle {
@@ -105,7 +105,11 @@ class ResourceFileSystem internal constructor(
     val existing = jarCache[this]
     if (existing != null) return existing
 
-    val created = SYSTEM.openZip(this)
+    val created = okio.internal.openZip(
+      zipPath = this,
+      fileSystem = SYSTEM,
+      predicate = { zipEntry -> !zipEntry.canonicalPath.name.endsWith(".class", ignoreCase = true) }
+    )
 
     // Recover from a race if two threads open the same zip at the same time.
     val replaced = jarCache.putIfAbsent(this, created) ?: return created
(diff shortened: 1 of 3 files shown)


okio/src/jvmMain/kotlin/okio/ResourceFileSystem.kt
```diff
@@ -105,7 +105,9 @@ class ResourceFileSystem internal constructor(
     val existing = jarCache[this]
     if (existing != null) return existing
 
-    val created = SYSTEM.openZip(this)
+    val created = okio.internal.openZip(this, SYSTEM) { zipEntry ->
```
yschimke, line 108: I wonder whether this is clearer with the filter as a named Param. It's not obvious to mean that a filter is the defining block here. Your call.

okio/src/jvmMain/kotlin/okio/internal/zip.kt
```diff
@@ -55,10 +55,17 @@ private const val HEADER_ID_EXTENDED_TIMESTAMP = 0x5455
 /**
  * Opens the file at [zipPath] for use as a file system. This uses UTF-8 to comments and names in
  * the zip file.
+ *
+ * @param predicate a function that returns false for entries that should be omitted from the file
```
yschimke, line 59: TIOLI: a filter for entries...

yschimke (approved):

okio/src/jvmMain/kotlin/okio/ResourceFileSystem.kt
swankjesse: Done!

okio/src/jvmMain/kotlin/okio/internal/zip.kt
swankjesse: I always always need to check whether true means keep or true means discard. Saying exactly that is my hope at helping at least myself.

yschimke:
@swankjesse FWIW I actually hope that we head away from ResourceFileSystem.SYSTEM_RESOURCES, to a much more declarative resource mechanism.  Closer to apps having specific sets of resources they provide without overlap etc.  Think the public top level database for example. It should come from a specific module, and maybe you overide how its accessed in other environments e.g. platform specific within KMM.

swankjesse:
@yschimke what does the API for this look like? I actually think what we have here is pretty simple. Making it work on native is interesting!

yschimke:
I think that's why I believe there is another API. Current design is like java classpath. I think better design is declaring which resources you include and make them supported anywhere in KMM apps.

swankjesse:
Gotcha. My preference is to start with this JVM only thing cause I want to use it in a bunch of JVM only applications!

But also to leave the door open for multiplatform things too.

Issue #910, opened 2021-03-17
JS IR compiler support


andersio:
While JS IR backend is still in alpha, IR compiler support is valuable to prototype and test TypeScript integration. So it would be great if okio can enable `kotlin.js.compiler=both` at some point.

Having said that, when I try to convert square/okio 3.0.0-alpha1 to work with the JS IR compiler, I stumbled upon a compiler bug & filed it as [KT-45542](https://youtrack.jetbrains.com/issue/KT-45542). It is reproducible on both Kotlin 1.4.31 and 1.5.0-M1. Since square/okio uses default parameter values in its expect classes quite a lot, guess we will all have to wait then. :shipit:

JakeWharton:
Thanks! Feel free to fire over a PR when upstream is fixed as well!

andersio:
@JakeWharton Is it acceptable to workaround the issue by dropping default arguments?

e.g.
```diff
-  fun lastIndexOf(other: ByteString, fromIndex: Int = size): Int
+  fun lastIndexOf(other: ByteString, fromIndex: Int): Int
+  fun lastIndexOf(other: ByteString): Int
```

Not sure if this would change the ABI though.

swankjesse:
@andersio that is not a binary-compatible change.
 Too bad!

PR #940, opened 2021-05-22
Support Kotlin/JS IR and Legacy JS

swankjesse:
This introduces a very ugly workaround to a limitation in the IR compiler
where default parameter values cannot be functions on the invoked object.
https://youtrack.jetbrains.com/issue/KT-45542

It also works around a bug in IR where Long.toString() returns the wrong
value.
https://youtrack.jetbrains.com/issue/KT-39891

Closes: https://github.com/square/okio/issues/910

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,7 @@
 org.gradle.jvmargs='-Dfile.encoding=UTF-8'
 android.enableJetifier=true
 android.useAndroidX=true
+kotlin.js.compiler=both
 
 # Publishing SHA 256 and 512 hashses of maven-metadata is not supported by Sonatype and Nexus.
 # See https://github.com/gradle/gradle/issues/11308 and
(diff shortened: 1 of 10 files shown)


swankjesse:
FYI @andersio

okio-fakefilesystem/build.gradle.kts
```diff
@@ -10,7 +10,7 @@ kotlin {
     withJava()
   }
   if (kmpJsEnabled) {
-    js {
+    js(BOTH) {
```
JakeWharton, line 13: You can specify this once in the root `gradle.properties` as
```properties
kotlin.js.compiler=both
```
rather than having to remember to update all modules.

JakeWharton (approved):

okio-fakefilesystem/build.gradle.kts
swankjesse: Fixed!

PR #944, opened 2021-05-29
Convert gradle-mvn-mpp-push.gradle to KTS

swankjesse:


swankjesse:
Test failure is unrelated and I’m trying to understand more.
https://github.com/square/okio/pull/945

buildSrc/build.gradle.kts
```diff
@@ -1,9 +1,12 @@
-import org.gradle.kotlin.dsl.`kotlin-dsl`
```
oldergod, line 1: I _think_ you need this, don't you?

okio-fakefilesystem/build.gradle.kts
```diff
@@ -80,4 +80,4 @@ dependencies {
   signature(deps.animalSniffer.javaSignature)
 }
 
-apply(from = "$rootDir/gradle/gradle-mvn-mpp-push.gradle")
+apply(plugin = "okio-publishing")
```
oldergod, line 83: I had no idea you could declare a local plugin like that

oldergod (approved):

buildSrc/build.gradle.kts
swankjesse: CI says no!

PR #959, opened 2021-06-20
APIs to seek a FileHandle's source or sink

swankjesse:
The one fancy thing we're doing here is preserving the buffered data if
we're seeking forward on a source.

I don't think there's a compelling equivalent for preserving buffered
data when seeking backwards on a sink. In theory we could truncate the
buffer to the new seek position, but this breaks if the writer doesn't
overwrite the full range of what was truncated.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/FileHandle.kt
+++ b/okio/src/commonMain/kotlin/okio/FileHandle.kt
@@ -167,8 +167,42 @@ abstract class FileHandle(
     require(source is FileHandleSource && source.fileHandle === this) {
       "source was not created by this FileHandle"
     }
+    check(!source.closed) { "closed" }
 
     return source.position - bufferSize
+  }
+
+  /**
+   * Change the position of [source] in the file to [position]. The argument [source] must be either
+   * a source produced by this file handle, or a [BufferedSource] that directly wraps such a source.
+   * If the parameter is a [BufferedSource], it will skip or clear buffered bytes.
+   */
+  @Throws(IOException::class)
+  fun reposition(source: Source, position: Long) {
+    if (source is RealBufferedSource) {
+      val fileHandleSource = source.source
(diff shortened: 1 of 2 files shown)


okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -171,6 +171,37 @@ abstract class FileHandle(
     return source.position - bufferSize
   }
 
+  /**
+   * Change the position of [source] in the file to [position]. The argument [source] must be either
+   * a source produced by this file handle, or a [BufferedSource] that directly wraps such a source.
+   * If the parameter is a [BufferedSource], it will skip or clear buffered bytes.
+   */
+  @Throws(IOException::class)
+  fun position(source: Source, position: Long) {
```
swankjesse, line 180: should this be reposition? I’m tempted by that because it’s symmetric with size/resize

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -171,6 +171,37 @@ abstract class FileHandle(
...
+      } else {
+        // The new position doesn't share data with the current buffer.
+        source.buffer.clear()
+        fileHandleSource.position = position
+      }
+    } else {
+      require(source is FileHandleSource && source.fileHandle === this) {
+        "source was not created by this FileHandle"
+      }
+      source.position = position
```
yschimke, line 201: Should these check closed?

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: done.

okio/src/commonMain/kotlin/okio/FileHandle.kt
swankjesse: Good call. Done.

oldergod (approved):

Issue #947, opened 2021-05-31
Upgrade to kotlin 1.5.0 or 1.5.10


9468305:
current release depends on 1.4.20.

PR #960, opened 2021-06-24
Kotlin 1.5.20

anthonymonori:
Changelog:
- Updated versions.kotlin to 1.5.20
    - Changelog for 1.5.20: https://kotlinlang.org/docs/whatsnew1520.html
    - Changelog for 1.5.0: https://kotlinlang.org/docs/whatsnew15.html
    - Changelog for 1.4.30: https://kotlinlang.org/docs/whatsnew1430.html
- Updated kotlinx-datetime from 0.1.1 to 0.2.1 since Kotlin 1.5.0 depends on datetime > 0.2.0
    - Due to the changes to the experimental Duration inline value class
    - Changelog: https://github.com/Kotlin/kotlinx-datetime/blob/master/CHANGELOG.md#021
- Changed expected values in tests due to hardcoded values of some updates around Kotlin JS/IR and Legacy JS APIs

Closes: #947

### DIFF
--- a/okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
+++ b/okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
@@ -58,8 +58,8 @@ abstract class AbstractFileSystemTest(
     val cwdString = cwd.toString()
     assertTrue(cwdString) {
       cwdString.endsWith("okio${Path.DIRECTORY_SEPARATOR}okio") ||
-        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-jsLegacy-test") ||
-        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-jsIr-test") ||
+        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-js-legacy-test") ||
+        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-js-ir-test") ||
         cwdString.contains("/CoreSimulator/Devices/") || // iOS simulator.
         cwdString == "/" // Android emulator.
     }
--- a/buildSrc/src/main/kotlin/Dependencies.kt
+++ b/buildSrc/src/main/kotlin/Dependencies.kt
@@ -5,7 +5,7 @@ val kmpJsEnabled = System.getProperty("kjs", "true").toBoolean()
 val kmpNativeEnabled = System.getProperty("knative", "true").toBoolean()
 
 object versions {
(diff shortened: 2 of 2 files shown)


okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -58,8 +58,8 @@ abstract class AbstractFileSystemTest(
     val cwdString = cwd.toString()
     assertTrue(cwdString) {
       cwdString.endsWith("okio${Path.DIRECTORY_SEPARATOR}okio") ||
-        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-jsLegacy-test") ||
-        cwdString.endsWith("${Path.DIRECTORY_SEPARATOR}okio-parent-okio-jsIr-test") ||
```
anthonymonori, line 62: @swankjesse I've seen you changing these from `js-ir` to `jsIr` in a recent PR, but I couldn't figure out what these actually come from. It seems to have changed in some stdlib shipped with 1.5.0. Do advise if you think this is unexpected!

swankjesse:
Unfortunately I'd like for Okio to be viable in Gradle plugins, and Gradle doesn't want us using anything newer than 1.4.30, even if we're on the latest Gradle.
https://docs.gradle.org/7.0/release-notes.html

anthonymonori:
> Unfortunately I'd like for Okio to be viable in Gradle plugins, and Gradle doesn't want us using anything newer than 1.4.30, even if we're on the latest Gradle.
> https://docs.gradle.org/7.0/release-notes.html

I see, thanks for pointing that out! In that case we need to block this until they ship it (in 7.2 potentially): https://github.com/gradle/gradle/issues/17465

twyatt:
When trying the `anthonymonori:update-to-kotlin-1-5-20` branch, I'm unable to publish to maven local:

```
$ ./gradlew publishToMavenLocal

...

> Task :okio:publishKotlinMultiplatformPublicationToMavenLocal FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':okio:publishKotlinMultiplatformPublicationToMavenLocal'.
> Failed to publish publication 'kotlinMultiplatform' to repository 'mavenLocal'
   > Invalid publication 'kotlinMultiplatform': multiple artifacts with the identical extension and classifier ('jar', 'sources').
```

Whereas `master` @ ed6f833 (parent commit of `anthonymonori:update-to-kotlin-1-5-20`) publishes to local Maven successfully.

swankjesse (approved):

swankjesse:
Well I'm convinced. Shading is a viable workaround.

@anthonymonori thanks for the PR!

PR #961, opened 2021-06-30
Test case that demonstrates a problem with ResourceFileSystem

swankjesse:
Directory listing is broken when the resource is in a .jar file.

### DIFF
--- a/okio/src/jvmTest/java/okio/internal/ResourceFileSystemTest.kt
+++ b/okio/src/jvmTest/java/okio/internal/ResourceFileSystemTest.kt
@@ -21,15 +21,19 @@ import okio.FileSystem
 import okio.IOException
 import okio.Path
 import okio.Path.Companion.toPath
+import okio.ZipBuilder
 import okio.ZipFileSystem
+import okio.randomToken
 import org.assertj.core.api.Assertions.assertThat
 import org.junit.Test
+import java.net.URLClassLoader
 import kotlin.reflect.KClass
 import kotlin.test.fail
 
 @ExperimentalFileSystem
 class ResourceFileSystemTest {
   private val fileSystem = FileSystem.RESOURCES as ResourceFileSystem
+  private var base = FileSystem.SYSTEM_TEMPORARY_DIRECTORY / randomToken(16)
 
   @Test
   fun testResourceA() {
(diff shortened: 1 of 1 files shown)


swankjesse:
This fails because ClassLoader resources really don’t have good support for listing at the root level.

I think the fix is going to require some overhauling of ResourceFileSystem. Something like this:
1. Attempt to get all the URLs from the class loader and its parent class loaders (with [URLClassLoader.getURLs](https://docs.oracle.com/javase/7/docs/api/java/net/URLClassLoader.html#getURLs()))
2. Convert those URLs to instances of Pair<FileSystem, okio.Path> as we do today
3. Union those into a single file system

Ultimately this ends up not using ClassLoader.getResource() and instead implementing resource finding manually.

Upsides of fixing this:
 * Directories will work properly
 * Directories that are contributed by multiple classpath URLs will work properly

Drawbacks: 
 * It’ll only work with URLClassLoaders. I _think_ this should be fine?

yschimke:
I think overall we should probably consider ResourceFileSystem experimental until it gets a thorough thrashing.

Another option is to adopt a set of practices that we support for ResourceFileSystem, be opinionated about it's usage in a way that we are confident we can support (theoretically) across JDK, Android, KMP, Graal. I imagined for native images we may want to use the presence of a ResourceFileSystem subset e.g.

```
object OkHttpResources {
  val fileSystem = ResourceFileSystem(OkHttpClient.package.path)
}
```

to auto generate the resource listing that native image needs.

yschimke:
To frame it another way - what is a separate set of conventions we could apply to Okio ResourceFileSystem that would make it's usage predictably safe across all supported environments KMP (JS, Native), GraalVM, JDK, JDK shaded, Android.

yschimke:
Random half baked thought leadership aside, I do like the idea of supporting this on JDK platform at least

```
    val resourceFileSystem = ResourceFileSystem(
      classLoader = URLClassLoader(arrayOf(zipPath.toFile().toURI().toURL()), null)
    )
```

yschimke:
To be clear it isn't so much "/" that fails, it just seems quite awkward listing to find a resource that could exist in multiple directories, like META-INF. Your suggested fix sounds perfect to me on the JDK, it just exposes that we don't have a clean solution for all Okio supported platforms.

yschimke:
You want me to make a fix? Or you got this already?

swankjesse:
I'll prep a fix!

oldergod (approved):

PR #962, opened 2021-07-08
Change ResourceFileSystem to load roots eagerly

swankjesse:
Previously ResourceFileSystem delegated most operations to the
ClassLoader on-demand. Now it eagerly computes a set of base paths
and then unions them.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
@@ -27,36 +27,54 @@ import okio.Sink
 import okio.Source
 import java.io.File
 import java.io.IOException
-import java.util.concurrent.ConcurrentHashMap
+import java.net.URL
 
 /**
  * A file system exposing Java classpath resources. It is equivalent to the files returned by
- * [ClassLoader.getResource].
+ * [ClassLoader.getResource] but supports extra features like [metadataOrNull] and [list].
  *
- * This file system does not implement merging of multiple paths from difference resources like
- * overlapping `.jar` files.
+ * If `.jar` files overlap, this returns an arbitrary element. For overlapping directories it unions
+ * their contents.
  *
  * ResourceFileSystem excludes `.class` files.
+ *
+ * This file system is read-only.
  */
 @ExperimentalFileSystem
 internal class ResourceFileSystem internal constructor(
-  private val classLoader: ClassLoader
+  classLoader: ClassLoader
 ) : FileSystem() {
-  private var jarCache = ConcurrentHashMap<Path, FileSystem>()
+  private var roots: List<Pair<FileSystem, Path>> = classLoader.toClasspathRoots()
 
   override fun canonicalize(path: Path): Path {
-    return "/".toPath() / path
+    return ROOT / path
   }
 
   override fun list(dir: Path): List<Path> {
-    val (fileSystem, fileSystemPath) = toSystemPath(dir) ?: return listOf()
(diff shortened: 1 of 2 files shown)


oldergod (approved):

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
```diff
@@ -27,101 +27,77 @@ import okio.Sink
...
-    return fileSystem.list(fileSystemPath)
+    val relativePath = dir.toRelativePath()
+    val result = mutableSetOf<Path>()
+    var foundAny = false
+    for ((fileSystem, base) in roots) {
+      try {
+        result += fileSystem.list(base / relativePath)
+          .map { it.removeBase(base) }
+        foundAny = true
+      } catch (_: IOException) {
```
yschimke, line 62: I wonder whether any non-FileNotFoundException here should be retained and later thrown in the suppression list?

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
```diff
@@ -136,4 +112,56 @@ internal class ResourceFileSystem internal constructor(
...
+      // entry.
+      //
+      // Classpath entries that aren't directories and don't have a META-INF/MANIFEST.MF file will
+      // not be visible in this file system.
+      return getResources("").toList().mapNotNull { it.toFileRoot() } +
+        getResources("META-INF/MANIFEST.MF").toList().mapNotNull { it.toJarRoot() }
+    }
+
+    fun URL.toFileRoot(): Pair<FileSystem, Path>? {
+      if (protocol != "file") return null // Ignore unexpected URLs.
```
yschimke, line 146: I wonder what this does in Graal native images...?

yschimke: If it doesn't behave we should probably add specific support via nativeImage build time generation of file lists.  So not a blocker here.  That would be a fun feature to add, and potentially fix generally or with some specific API for that world.

yschimke:
Thanks for adding, I like the FileSystem model.  Think it's a great (sleeper) feature for okio.

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
swankjesse: Yep, great idea. It'll likely be different from this! Gotta look at how the public suffix list works in OkHttp?!

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
swankjesse: I think it might be confusing as it'll have the physical path, not the mapped path.

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
yschimke: We whitelist this https://github.com/square/okhttp/blob/master/okhttp/src/main/resources/META-INF/native-image/okhttp/okhttp/resource-config.json

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
yschimke: Seems arguably confusing if a file permission issue or data corruption is silently swallowed and results in empty.  But happy to see real results and use any bugs raised as a trigger to address this.

PR #964, opened 2021-07-12
The release process is now simpler

swankjesse:


JakeWharton (approved):
Same setup as cashapp that I can use on other square projects?

swankjesse:
Yes!

swankjesse:
i wanna hook up startship next so there’s no manual promotion necessary.

PR #965, opened 2021-07-13
Make ResourceFileSystem initialize lazily

swankjesse:
My recent PRs to load all roots makes behavior better, but it causes the
file system to be initialized too eagerly. This makes every FileSystem user
pay for initialization of the ResourceFileSystem even if they don't use it.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
@@ -42,9 +42,16 @@ import java.net.URL
  */
 @ExperimentalFileSystem
 internal class ResourceFileSystem internal constructor(
-  classLoader: ClassLoader
+  classLoader: ClassLoader,
+  indexEagerly: Boolean,
 ) : FileSystem() {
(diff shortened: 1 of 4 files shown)


okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
```diff
@@ -42,9 +42,16 @@ import java.net.URL
...
+  indexEagerly: Boolean,
 ) : FileSystem() {
-  private var roots: List<Pair<FileSystem, Path>> = classLoader.toClasspathRoots()
+  private val roots: List<Pair<FileSystem, Path>> by lazy { classLoader.toClasspathRoots() }
+
+  init {
+    if (indexEagerly) {
+      roots.size
+    }
+  }
```
oldergod, line 54: That's clever and weird at the same time...

oldergod (approved):

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
swankjesse: Yeah, I tried to find something more explicit and failed!

PR #968, opened 2021-07-18
Open files in binary mode on Windows

swankjesse:
Without this, files that contain 0x1a will be prematurely terminated.

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-160

### DIFF
--- a/okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
+++ b/okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
@@ -17,6 +17,7 @@ package okio
 
 import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
+import okio.ByteString.Companion.encodeUtf8
 import okio.ByteString.Companion.toByteString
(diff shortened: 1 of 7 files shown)


okio/src/commonTest/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -184,6 +185,69 @@ abstract class AbstractFileSystemTest(
...
   }
 
+  /**
+   * Write a file by concatenating three mechanisms, then read it in its entirety using three other
+   * mechanisms. This is attempting to defend against unwanted use of Windows text mode.
+   *
+   * https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/fopen-wfopen?view=msvc-160
+   */
+  @Test
+  fun fileSinkSpecialCharacters() {
```
oldergod, line 195: Nice test

oldergod (approved):

PR #969, opened 2021-07-23
ByteString.copyInto()

swankjesse:
I have a use-case for this in Wire and doing an intermediate copy is not
appropriate.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/ByteString.kt
+++ b/okio/src/commonMain/kotlin/okio/ByteString.kt
@@ -133,6 +133,14 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
    */
   fun rangeEquals(offset: Int, other: ByteArray, otherOffset: Int, byteCount: Int): Boolean
 
+  /**
+   * Copies bytes of this in `[offset..offset+byteCount]` to other in
+   * `[targetOffset..targetOffset+byteCount]`.
+   *
+   * @throws IndexOutOfBoundsException if either range is out of bounds.
+   */
+  fun copyInto(offset: Int = 0, target: ByteArray, targetOffset: Int = 0, byteCount: Int)
+
   fun startsWith(prefix: ByteString): Boolean
 
   fun startsWith(prefix: ByteArray): Boolean
--- a/okio/src/jvmMain/kotlin/okio/ByteString.kt
+++ b/okio/src/jvmMain/kotlin/okio/ByteString.kt
@@ -18,6 +18,7 @@ package okio
 import okio.internal.commonBase64
 import okio.internal.commonBase64Url
 import okio.internal.commonCompareTo
+import okio.internal.commonCopyInto
 import okio.internal.commonDecodeBase64
 import okio.internal.commonDecodeHex
 import okio.internal.commonEncodeUtf8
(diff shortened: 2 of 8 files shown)


okio/src/commonMain/kotlin/okio/ByteString.kt
```diff
@@ -133,6 +133,14 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
    */
   fun rangeEquals(offset: Int, other: ByteArray, otherOffset: Int, byteCount: Int): Boolean
 
+  /**
+   * Copies bytes of this in `[offset..offset+byteCount]` to other in
+   * `[otherOffset..otherOffset+byteCount]`.
+   *
+   * @throws IndexOutOfBoundsException if either range is out of bounds.
+   */
+  fun copyInto(offset: Int = 0, target: ByteArray, targetOffset: Int = 0, byteCount: Int)
```
Egorand, line 142: Any reason not to make `byteCount` equal to `size` by default?

okio/src/jvmMain/kotlin/okio/ByteString.kt
```diff
@@ -156,6 +157,13 @@ internal actual constructor(
     byteCount: Int
   ): Boolean = commonRangeEquals(offset, other, otherOffset, byteCount)
 
+  actual open fun copyInto(
```
Egorand, line 160: Need overloads?

okio/src/jvmMain/kotlin/okio/SegmentedByteString.kt
```diff
@@ -108,6 +109,13 @@ internal actual class SegmentedByteString internal actual constructor(
     byteCount: Int
   ): Boolean = commonRangeEquals(offset, other, otherOffset, byteCount)
 
+  override fun copyInto(
```
Egorand, line 112: Same - overloads?

Egorand (approved):

okio/src/commonMain/kotlin/okio/ByteString.kt
swankjesse: I think it’s too difficult because ... which size? It’s uncertain whether source or target is larger

okio/src/jvmMain/kotlin/okio/ByteString.kt
swankjesse: I think I’d prefer to omit them just because in Java you’ll be calling this with a bunch of int arguments and having lots of overloads with lots of int parameters feels somehow less ergonomic.

okio/src/jvmMain/kotlin/okio/SegmentedByteString.kt
swankjesse: Same as above.

okio/src/commonMain/kotlin/okio/ByteString.kt
Egorand: Is your Wire use case "copy all bytes in `this.data` into `target`", or something different?

swankjesse: it’s copying an entire ByteString into a specific range of a target ByteArray

okio/src/commonMain/kotlin/okio/ByteString.kt
```diff
@@ -133,6 +133,14 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
    */
   fun rangeEquals(offset: Int, other: ByteArray, otherOffset: Int, byteCount: Int): Boolean
 
+  /**
+   * Copies bytes of this in `[offset..offset+byteCount]` to other in
+   * `[otherOffset..otherOffset+byteCount]`.
```
cketti, line 138: Should this be `targetOffset` rather than `otherOffset`?

swankjesse: YEP! Good catch

swankjesse: Fixed!

PR #970, opened 2021-07-23
Remove our custom ArrayIndexOutOfBoundsException

swankjesse:
This is binary-compatible on the JVM where the type was
already aliased to the real thing.

On native this is not a backwards-compatible change.

### DIFF
--- a/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
+++ b/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
@@ -17,7 +17,6 @@ package okio.fakefilesystem
 
 import kotlinx.datetime.Clock
 import kotlinx.datetime.Instant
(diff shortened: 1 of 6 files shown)


swankjesse:
Hmmm . . . . I don’t quite understand this.

```
> Task :okio:compileKotlinMetadata FAILED
e: /Users/jwilson/Projects/okio/okio/src/commonMain/kotlin/okio/-Util.kt: (24, 11): Unresolved reference: ArrayIndexOutOfBoundsException
e: /Users/jwilson/Projects/okio/okio/src/commonMain/kotlin/okio/internal/-Buffer.kt: (1537, 11): Unresolved reference: ArrayIndexOutOfBoundsException
e: /Users/jwilson/Projects/okio/okio/src/commonMain/kotlin/okio/internal/-Utf8.kt: (28, 11): Unresolved reference: ArrayIndexOutOfBoundsException
```

JakeWharton:
It's not available in common or JS.

PR #980, opened 2021-08-18
Support Kotlin Hierarchical Project structure

sellmair:
This Pull Request will be updated soon with a pinned 1.6 dev build and later with 1.6-M1.


JavierSegoviaCordoba:
This will work with Kotlin 1.6 exclusively, or it will work with 1.5.30 too?

sellmair:
This PR now targets Kotlin 1.6.0 only, yes!

JakeWharton:
Time to rebase on 1.6.0 final?

sellmair:
I updated the branch to 1.6.10-RC and found one blocking issue in https://github.com/square/okio/blob/79656d80e812c36342add36580ca5784e5b8f637/okio/src/commonMain/kotlin/okio/Path.kt#L286

This companion object's extension function with default parameter will lead to tooling issues. 
I will update this PR by linking a YT issue for this situation, explaining further technical details!

sellmair:
So here is a little more details: 
Here are the issues I found after rebasing the branch and updadting to 1.6.10:

## [KT-50120](https://youtrack.jetbrains.com/issue/KT-50120)
The compiler is failing to find default parameter values when those are declared in an 'common' module / expect class. 
Usages in OKIO of `String.toPath()` will, therefore fail to compile metadata with a NO_VALUE_FOR_PARAMETER error.

## [KT-50156](https://youtrack.jetbrains.com/issue/KT-50156)
I noticed that analysis inside source sets like 'commonMain' will take very long. The IDE performance I experienced here is not acceptable and shall be investigated/fixed as soon as possible. 

____ 
Both issues are assigned to myself and I am targeting 1.6.20 for both. 
I will start working on the NO_VALUE_FOR_PARAMETER issue right away.

My proposal for OKIO: Let's wait for 1.6.20 for migration. I will keep you updated on the progress of both of those issues!

swankjesse:
@sellmair that's a great plan. Thanks for the update.

sellmair:
Update: 
Both issues are resolved and in the 1.6.20 release branch of Kotlin 🎉 
For testing purposes I bumbped the Kotlin version to 1.6.20-M1-6.

I am particularly proud about the IDE/Compiler Frontend performance improvements. 
Analyzing 'commonMain/Path.kt`:

Before: ~48s
After: ~1s

Whoever is interested in the changes for this performance ticket: 
https://github.com/JetBrains/kotlin/compare/832433082339...ba201a249ec6

Feel free to ask on public lang slack, if somebody would be interested in more depth on what was going on here.

https://user-images.githubusercontent.com/34319766/147241780-d30eb88f-ca5a-44cf-bdaf-f0a15fa38851.mp4

JakeWharton:
Wow that's quite an improvement!

Thanks for all the time you spent on this. Helping real-world projects successfully use the latest and greatest features is a huge boon to the community.

micHar:
@sellmair I understand that we are waiting for stable release of 1.6.20 and then okio will merge these changes. But had another question: this performance improvement is part of the plugin or IDE? More specifically: will Android Studio also benefit from this?

sellmair:
Exactly! I will test this with 1.6.20 milestone and rc versions and afterwards, we can hopefully merge this in 1.6.20. 
The performance improvement is inside the Kotlin Compiler and therefore any IDE plugin (Android Studio and IntelliJ) should benefit from that with any plugin that includes a 1.6.20 compiler!

yschimke:
Can we hold off on this, even though 1.6.20 is released, because of Compose compatibility?

https://developer.android.com/jetpack/androidx/releases/compose-kotlin

JakeWharton:
There's no reason to wait. You can use the 1.6.10 compiler in your project with the 1.6.20 stdlib pulled from a dependency.

sellmair:
Updated to 1.6.20 and fixed the spotless issue ✅

sellmair:
Needs a fix for non-macOS host metadata compilation 😭 
I will be on vacation, this week. Could start working on it, afterward.

sellmair:
The metadata compilation is suffering from https://youtrack.jetbrains.com/issue/KT-52050 which I am going to fix now. 
The reason is that during commonization of nativeMain, the CPointed supertype was lost in platform.posix.DIR. Should be working for the project now, by using type inference instead. The fix for the root cause is targeting in 1.7.0 (which comes after 1.6.21)

JakeWharton (approved):

PR #981, opened 2021-08-19
Make Flushable available in common and fully implement sink in common

FabianTerhorst:
After trying to extend ``Sink`` in a common project i noticed it isn't possible because it is a `expect interface`. 
In this PR i make ``Sink`` a ``interface`` by making ``Flushable`` available in the common project. Same as it was done with ``Closeable``.


Egorand (approved):
Nice, I like it!

JakeWharton:
Can you explain the problem more? Nothing should prevent you from implementing the `Sink` interface in common as it is today.

FabianTerhorst:
I fixed the problem that the inheritance didn't worked by adding:
```kotlin
tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().all {
    kotlinOptions {
        jvmTarget = "1.8"
        freeCompilerArgs += "-Xjvm-default=all"
    }
}
```
Seems like that was unrelated to ``expect interface``.

FabianTerhorst:
But i think the pull request is still valid since it reduces the code.

JakeWharton:
It reduces code at the expense of exposing an otherwise unused public API. We begrudgingly expose `Closeable` to support the `use` extension in common code. There's no general purpose usage of `Flushable`.

As a library I much prefer to keep the public API surface as small as possible even if it increases the maintenance burden. Much better to burden us than our users.

FabianTerhorst:
Ok, thats understandable. When https://youtrack.jetbrains.com/issue/KT-37316 is resolved it probably would be more acceptable to do this change, because there would be no exposed api.

swankjesse:
I wonder if we can we use this instead?
```
@Suppress("ACTUAL_WITHOUT_EXPECT")
```

swankjesse:
No action on this for now. May revisit.

PR #984, opened 2021-08-30
Update gradle 7.2

Goooler:


Goooler:
Maybe we should merge this to solve [failure on publish](https://github.com/square/okio/runs/3540132553).

swankjesse (approved):
thanks!

PR #993, opened 2021-09-11
Restore support for JS browser platforms

swankjesse:
When I introduced the Node FileSystem I broke support for running Okio with
JavaScript in the browser. This restores that break.

This change introduces a new okio-testing-support module so that the
AbstractFileSystemTest can be shared by both okio and okio-nodefilesystem
modules. I had to move NodeJsFileSystemTest to the okio-nodefilesystem
module because the core okio module no longer depends on NPM, not even in
testing.


okio-testing-support/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
+plugins {
+  kotlin("multiplatform")
+}
+
+kotlin {
+  jvm {
+    withJava()
```
JakeWharton, line 7: Didn't see any Java files in the move

JakeWharton (approved):

okio-testing-support/build.gradle.kts
swankjesse: Good catch!

okio-testing-support/src/jsMain/kotlin/okio/testingSupportJs.kt
```diff
@@ -13,12 +13,8 @@
...
-
-/**
- * See `fs.kt` for information on what this file does and how to keep it up-to-date.
- */
-@file:JsModule("os")
-@file:JsNonModule
 package okio
 
-internal external fun tmpdir(): String
+actual fun isBrowser(): Boolean {
```
swankjesse, line 18: FYI, this is new! Added to ignore out tests that fail in browsers but not NodeJS. (Also want to dig into that, but not as urgently. Fails look like timeouts and not bugs.)

okio/src/jsMain/kotlin/okio/-JsPlatform.kt
```diff
@@ -15,11 +15,38 @@
...
+
+private val os: dynamic
   get() {
-    // TODO(swankjesse): return path.path.sep instead, once it has @JsNonModule
-    return when (platform()) {
-      "win32" -> "\\"
-      else -> "/"
+    return try {
+      js("require('os')")
+    } catch (t: Throwable) {
```
swankjesse, line 37: FYI also needed to change this up from the approved commit.

PR #1000, opened 2021-09-20
Code cleanups

Goooler:


JakeWharton:
I find a PR like this hard to approve because it does too much. Can we get ones which are split by topic? For example, all of the updates to the JUnit assert methods seems fine and would be an easy merge. But mixed with some more controversial changes (like removing named locals and the charset stuff) means they're unlikely to land.

Goooler:
Ok, I'll split them.

Goooler:
Simplify assertions #1003.

Goooler:
The others #1004

PR #1004, opened 2021-09-21
Code cleanups

Goooler:


okio/jvm/jmh/src/jmh/java/com/squareup/okio/benchmarks/IndexOfElementBenchmark.java
```diff
@@ -40,7 +40,7 @@
 @BenchmarkMode(Mode.SampleTime)
 @OutputTimeUnit(TimeUnit.MICROSECONDS)
 public class IndexOfElementBenchmark {
-  ByteString byteString = ByteString.encodeUtf8("abcd");
+  final ByteString byteString = ByteString.encodeUtf8("abcd");
```
swankjesse, line 43: I don’t think `final` on locals in Java is a good tradeoff for legibility vs. safety

okio/src/commonMain/kotlin/okio/-Base64.kt
```diff
@@ -87,8 +87,7 @@ internal fun String.decodeBase64ToArray(): ByteArray? {
     }
   }
 
-  val lastWordChars = inCount % 4
-  when (lastWordChars) {
+  when (inCount % 4) {
```
swankjesse, line 90: This loses a useful name, lastWordChars

okio/src/commonMain/kotlin/okio/internal/-Buffer.kt
```diff
@@ -1297,7 +1296,7 @@ internal inline fun Buffer.commonIndexOf(bytes: ByteString, fromIndex: Long): Lo
     while (offset < resultLimit) {
       // Scan through the current segment.
       val data = s.data
-      val segmentLimit = okio.minOf(s.limit, s.pos + resultLimit - offset).toInt()
```
swankjesse, line 1300: This one was interesting; we had our own minOf for mixed parameter types. Is this no longer necessary?

okio/src/commonMain/kotlin/okio/internal/-SegmentedByteString.kt
```diff
@@ -115,10 +115,9 @@ internal inline fun SegmentedByteString.commonSubstring(beginIndex: Int, endInde
 
   val newSegments = segments.copyOfRange(beginSegment, endSegment + 1)
   val newDirectory = IntArray(newSegments.size * 2)
-  var index = 0
-  for (s in beginSegment..endSegment) {
+  for ((index, s) in (beginSegment..endSegment).withIndex()) {
     newDirectory[index] = minOf(directory[s] - beginIndex, subLen)
-    newDirectory[index++ + newSegments.size] = directory[s + segments.size]
+    newDirectory[index + newSegments.size] = directory[s + segments.size]
```
swankjesse, line 120: Good find here

okio/src/hashFunctions/kotlin/okio/internal/Md5.kt
```diff
@@ -95,8 +95,7 @@ internal class Md5 : HashFunction {
     var d = h3
 
     for (i in 0 until 16) {
-      val g = i
-      val f = ((b and c) or (b.inv() and d)) + a + localK[i] + words[g]
+      val f = ((b and c) or (b.inv() and d)) + a + localK[i] + words[i]
```
swankjesse, line 98: I think this is closer to the spec when defined longform

okio/src/jvmTest/java/okio/ByteStringJavaTest.java
```diff
@@ -92,7 +93,7 @@
         new Object[] { Factory.ONE_BYTE_PER_SEGMENT, "SegmentedByteString (one-at-a-time)" });
   }
 
-  @Parameter(0) public Factory factory;
+  @Parameter() public Factory factory;
```
swankjesse, line 96: I think this is more clear with the literal 0

swankjesse (changes requested):
Thanks, but I don’t think this is a net improvement. Most of the time when there’s style warnings we’re deliberately disobeying them because we disagree with the tools emitting those warnings.

okio/src/commonMain/kotlin/okio/internal/-Buffer.kt
Goooler: Just simplify the package name, it has been imported.

okio/jvm/jmh/src/jmh/java/com/squareup/okio/benchmarks/IndexOfElementBenchmark.java
Goooler: Reverted.

okio/src/commonMain/kotlin/okio/-Base64.kt
Goooler: Reverted.

okio/src/jvmTest/java/okio/ByteStringJavaTest.java
Goooler: Reverted.

PR #1007, opened 2021-09-24
Implement Path.relativeTo(path)

oldergod:


okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,82 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+    return ".".toPath()
+  }
+
+  if (lastCommonByteIndex == 0 && isAbsolute) {
+    throw IllegalArgumentException(
+      "Paths of different roots cannot be relative to each other: $this and $other")
+  }
+
+  val lastCommonSlashIndex = bytes.substring(0, lastCommonByteIndex).lastIndexOf(slash)
+  val firstDiffIndex = lastCommonSlashIndex + slash.size
```
oldergod, line 192: I don't like this part

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -194,6 +194,16 @@ expect class Path internal constructor(bytes: ByteString) : Comparable<Path> {
    */
   operator fun div(child: Path): Path
 
+  /**
+   * Returns this path relative to [other].
```
swankjesse, line 198: Gimme more docs....

This inverts the resolve operator, `/`. For any two paths `a` and `b` that have the same root, `(a / (b.relativeTo(a))) == b`

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -194,6 +194,16 @@ expect class Path internal constructor(bytes: ByteString) : Comparable<Path> {
...
+   *
+   * This will throw an exception if this path and [other] are not both [absolute paths][isAbsolute]
```
swankjesse, line 200: ‘This will throw an exception’ –> ‘@throws IllegalArgumentException’

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+@ExperimentalFileSystem
+@Suppress("NOTHING_TO_INLINE")
+internal inline fun Path.commonRelativeTo(other: Path): Path {
+  require(isAbsolute && other.isAbsolute || isRelative && other.isRelative) {
+    "Paths of different roots cannot be relative to each other: $this and $other"
+  }
+
+  val thisSlash = slash
+  val otherSlash = other.slash
+  require(thisSlash == otherSlash) {
```
swankjesse, line 167: I think this is too strict for null slashes. I think these should pass?

```
val a = "foo".toPath()
val b = "foo/bar".toPath()
assertEquals(a.relativeTo(b), "..".toPath())
assertEquals(b.relativeTo(a), "bar".toPath())
```

I also think we should forgive different slashes if they aren’t leading:

```
val a = "foo/red".toPath()
val b = "foo\\green".toPath()
assertEquals(a.relativeTo(b), "...\\red".toPath())
assertEquals(b.relativeTo(a), "../green".toPath())
```

okio/src/commonMain/kotlin/okio/Path.kt
swankjesse: Let’s also add that invariant to the test suite?! Might find bugs in this, or in resolve!

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+    "Paths of different platforms cannot be relative to each other: $this and $other"
+  }
+
+  val slash = thisSlash ?: otherSlash ?: Path.DIRECTORY_SEPARATOR.toSlash()
+
+  // We look at the path both have in common.
+  var lastCommonByteIndex = 0
+  while (lastCommonByteIndex < bytes.size &&
+    lastCommonByteIndex < other.bytes.size &&
+    (bytes[lastCommonByteIndex] and 0xff) == (other.bytes[lastCommonByteIndex] and 0xff)
```
swankjesse, line 177: I think the `and 0xff` bit isn’t strictly necessary here, because left and right sides will both be signed bytes.

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+  ) {
+    lastCommonByteIndex++
+  }
+
+  if (lastCommonByteIndex == bytes.size && bytes.size == other.bytes.size) {
+    // `this` and `other` are the same path.
```
swankjesse, line 183: I really like your implementation so far . . .

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+    return ".".toPath()
+  }
+
+  if (lastCommonByteIndex == 0 && isAbsolute) {
+    throw IllegalArgumentException(
```
swankjesse, line 188: Okay, so here’s a super annoying wrench. For UNC names the root is everything up to the 3rd slash, and I don’t think these can traverse to each other.

That is, I think we should throw on these paths:
```
val a = "\\\\server1\foo".toPath()
val b = "\\\\server2\bar".toPath()
```

Lemme independently verify this though.

okio/src/commonMain/kotlin/okio/internal/-Path.kt
swankjesse: I like this line. I might have been tempted to assume slash.size is 1!

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+      "Paths of different roots cannot be relative to each other: $this and $other"
+    )
+  }
+
+  val lastCommonSlashIndex = bytes.substring(0, lastCommonByteIndex).lastIndexOf(slash)
```
swankjesse, line 193: Save an allocation and do `bytes.lastIndexOf(slash, lastCommonByteIndex)`

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+  val firstDiffIndex = lastCommonSlashIndex + slash.size
+
+  val buffer = Buffer()
+  val stepsToCommonParent = other.bytes.substring(firstDiffIndex, other.bytes.size).count(slash)
```
swankjesse, line 197: Once we’re happy with the behavior we should go over this to avoid allocations.

swankjesse: Another weird test case this suggests:

```
val a = "../../../foo".toPath()
val b = "../../../../bar".toPath()
assertEquals(b.relativeTo(a), "../../bar".toPath())
assertFailsWith<IllegalArgumentException> { a.relativeTo(b) }

```

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+      buffer.write(slash)
+    }
+  }
+
+  buffer.write(bytes.substring(firstDiffIndex, bytes.size))
+  return buffer.toPath()
+}
+
+@ExperimentalFileSystem
+private val Path.slash: ByteString?
```
swankjesse, line 211: Nice

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+  get() {
+    return when {
+      bytes.indexOf(SLASH) != -1 -> SLASH
+      bytes.indexOf(BACKSLASH) != -1 -> BACKSLASH
+      else -> null
+    }
+  }
+
+/** Returns the number of [value]s present in this [ByteString]. */
+private fun ByteString.count(value: ByteString): Int {
```
swankjesse, line 221: Nice

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -159,6 +155,84 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+  var count = 0
+  var i = 0
```
swankjesse, line 223: rename `i` to `pos`

okio/src/commonTest/kotlin/okio/PathTest.kt
```diff
@@ -380,4 +381,131 @@ class PathTest {
...
   }
+
+  @Test
+  fun relativeToWindowsPaths() {
+    val a = "C:\\Windows\\notepad.exe".toPath()
+    val b = "C:\\".toPath()
+    assertEquals("..\\..".toPath(), b.relativeTo(a))
+    assertEquals("Windows\\notepad.exe".toPath(), a.relativeTo(b))
+
+    val c = "C:\\Windows\\".toPath()
```
swankjesse, line 392: this test is a bit impossible because `c` won’t have the trailing `/` after toPath(). (But keep it, I like it)

okio/src/commonTest/kotlin/okio/PathTest.kt
```diff
@@ -380,4 +381,131 @@ class PathTest {
...
+    assertEquals("../../resume.txt".toPath(), a.relativeTo(b))
+
+    val c = "documents/resume.txt".toPath()
+    val d = "downloads/2021/taxes.txt".toPath()
+    assertEquals("../../downloads/2021/taxes.txt".toPath(), d.relativeTo(c))
+    assertEquals("../../../documents/resume.txt".toPath(), c.relativeTo(d))
+  }
+
+  @Test
+  fun unixRelativeToWindows() {
```
swankjesse, line 493: Let’s support this. Getting the slash right isn’t that important!

swankjesse (changes requested):

okio/src/commonMain/kotlin/okio/internal/-Path.kt
swankjesse: Just checked. It’s consistent with java.io.File.

```
    checkRelative("""aa/bb""", """aa/bb/c""", """c""")
    checkRelative("""../../aa/bb""", """../../aa/bb/c""", """c""")
    checkRelative("""../../aa""", """../../../bb""", """../../bb""")
    checkRelative("""../../../aa""", """../../bb""", null)
```

okio/src/commonMain/kotlin/okio/internal/-Path.kt
swankjesse: On the JVM:
```
\\localhost\c$\Windows
okio exists: true
file exists: true

\\127.0.0.1\c$\Windows
okio exists: true
file exists: true

\\127.0.0.1\c$\Windows\..\Windows
okio exists: true
file exists: true

\\127.0.0.1\c$\..\c$\Windows
okio exists: false
file exists: false

\\localhost\..\127.0.0.1\c$\Windows
okio exists: java.nio.file.FileSystemException: \\localhost\..\127.0.0.1\c$\Windows: The network name cannot be found.
file exists: false

\\127.0.0.1\..\localhost\c$\Windows
okio exists: java.nio.file.FileSystemException: \\127.0.0.1\..\localhost\c$\Windows: The network name cannot be found.
file exists: false

\\127.0.0.1\c$\..\..\localhost\c$\Windows
okio exists: false
file exists: false
```

Okio native:
```
\\localhost\c$\Windows
okio exists: true

\\127.0.0.1\c$\Windows
okio exists: true

\\127.0.0.1\c$\Windows\..\Windows
okio exists: true

\\127.0.0.1\c$\..\c$\Windows
okio exists: false

\\localhost\..\127.0.0.1\c$\Windows
okio exists: false

\\127.0.0.1\..\localhost\c$\Windows
okio exists: false

\\127.0.0.1\c$\..\..\localhost\c$\Windows
okio exists: false
```

swankjesse: (In these I’ve manually removed support for dropping the `..` operator in `String.toPath()`.)

okio/src/jvmTest/kotlin/okio/utilJvm.kt
```diff
@@ -15,24 +15,26 @@
...
 ) {
-  assertEquals(bRelativeToA, b.relativeTo(a))
-  if (confirmResolutionInversion) {
-    assertEquals(b, a / b.relativeTo(a))
+  val actual = b.relativeTo(a)
+  assertEquals(bRelativeToA, actual)
+  try {
+    assertEquals(b.withUnixSlashes(), (a / actual).withUnixSlashes())
+  } catch (e: IllegalArgumentException) {
+    // This is also okay. It means we lose information and can't reverse the operation.
```
oldergod, line 34: When does that happen? All were passing on my end, except for the slash difference

swankjesse: I just removed the boolean parameter.

oldergod: But now that we do `withUnixSlashes()`, the assertion should never fail!

swankjesse: Sorry I misspoke. The information we don’t have is the directory the `..` erases.

oldergod:
Ready for review

okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -208,6 +208,20 @@ expect class Path internal constructor(bytes: ByteString) : Comparable<Path> {
...
 
+  /**
+   * Returns this path relative to [other]. This effectively inverts the resolve operator, `/`. For
+   * any two paths `a` and `b` that have the same root, `a / (b.relativeTo(a))` is equal to `b`. If
+   * both paths don't use the same slash, the resolved path will use the slash of the [other] path.
+   *
+   * @throws IllegalArgumentException if this path and the [other] path are not both
+   * [absolute paths][isAbsolute] or both [relative paths][isRelative], or if they are both
+   * [absolute paths][isAbsolute] but of different roots (C: vs D:, or C: vs \\server, etc.).
+   * It will also throw if the relative path is impossible to resolve such as "resumes" relative to
```
swankjesse, line 219: I think the resumes/taxes example is super confusing in this sentence

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -228,6 +227,83 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+    return when {
+      bytes.indexOf(SLASH) != -1 -> SLASH
+      bytes.indexOf(BACKSLASH) != -1 -> BACKSLASH
+      else -> null
+    }
+  }
+
+/** Returns the number of [value]s present in this [ByteString]. */
+private fun ByteString.count(value: ByteString): Int {
+  var count = 0
```
swankjesse, line 288: Is this function unused

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -228,6 +227,83 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+  }
+  for (i in firstNewSegmentIndex until thisSegments.size) {
+    buffer.write(thisSegments[i])
+    buffer.write(slash)
+  }
+  return buffer.toPath()
+}
+
+@ExperimentalFileSystem
+private fun Byte.equalsEitherSlash(other: Byte): Boolean {
```
swankjesse, line 272: Unsed

okio/src/commonMain/kotlin/okio/internal/-Path.kt
```diff
@@ -228,6 +227,83 @@ internal inline fun Path.commonResolve(child: Path): Path {
...
+      else -> {
+        pos = next + 1
+        count++
+      }
+    }
+  }
+  return count
+}
+
+private val Byte.isSlash: Boolean
```
swankjesse, line 302: Unused?

swankjesse (approved):

PR #1011, opened 2021-09-29
Organize target platforms

swankjesse:
I'm about to introduce new platforms and don't want to paste
the platforms list in a bunch of new places.


JakeWharton (approved):

swankjesse:
Ugh, I couldn’t get buildSrc to work and it’s unclear why. Just lots of things where the Kotlin plugin is crashing on inconsistent versions of things. I moved the new code into a `build-support` directory instead.

build-support/src/main/kotlin/BuildSupport.kt
```diff
@@ -0,0 +1,8 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+
+class BuildSupport : Plugin<Project> {
```
swankjesse, line 4: I couldn’t figure out how to use functions from this module in my `build.gradle.kts` files without also implementing the plugin interface.

JakeWharton: Super easy in Groovy Gradle 😁

swankjesse: TRUE

PR #1014, opened 2021-10-04
API and implementation for listRecursively

oldergod:
Nothing much yet, but that could be a start.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
@@ -70,6 +70,13 @@ internal open class JvmSystemFileSystem : FileSystem() {
     return result
   }
 
+  override fun listRecursively(dir: Path): Sequence<Path> {
+    val localList = list(dir).asSequence()
+    return localList.flatMap {
+      listOf(it) + listRecursively(it)
+    }
+  }
+
   override fun openReadOnly(file: Path): FileHandle {
     return JvmFileHandle(readWrite = false, randomAccessFile = RandomAccessFile(file.toFile(), "r"))
   }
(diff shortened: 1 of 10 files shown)


okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
```diff
@@ -70,6 +70,18 @@ internal open class JvmSystemFileSystem : FileSystem() {
     return result
   }
 
+  override fun listRecursively(dir: Path): Sequence<Path> {
+    if (!dir.toFile().isDirectory) return emptySequence()
```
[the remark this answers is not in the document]
swankjesse: Better to attempt list() and recover!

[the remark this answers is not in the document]
swankjesse: Also I think we aught to put this implementation in the superclass!

okio/src/jvmMain/kotlin/okio/ZipFileSystem.kt
```diff
@@ -110,6 +110,19 @@ class ZipFileSystem internal constructor(
     return entry.children.toList()
   }
 
+  override fun listRecursively(dir: Path): Sequence<Path> {
+    val canonicalDir = canonicalize(dir)
```
swankjesse, line 114: is this necessary? Won’t `list` do it anyway?

swankjesse: (not yet, because we need to recover from a list() failure)

okio/src/jvmMain/kotlin/okio/internal/ResourceFileSystem.kt
```diff
@@ -75,6 +76,18 @@ internal class ResourceFileSystem internal constructor(
     return result.toList()
   }
 
+  override fun listRecursively(dir: Path): Sequence<Path> {
+    if (!Files.isDirectory(dir.toNioPath())) return emptySequence()
```
swankjesse, line 80: as above, isDirectory isn’t sufficient

swankjesse (changes requested):
Let’s PAIR!

swankjesse (approved):

PR #1018, opened 2021-10-12
createSymlinks implementation for UnixPosix variant.

oldergod:


okio/src/appleMain/kotlin/okio/-ApplePosixVariant.kt
```diff
@@ -15,27 +15,48 @@
...
       throw errnoToIOException(errno)
     }
+
+    var symlinkTarget: Path? = null
+    if (stat.st_mode.toInt() and S_IFMT == S_IFLNK) {
+      // `path` is a symlink, let's resolve its target.
+      memScoped {
+        val buffer = allocArray<ByteVar>(4096)
+        readlink(path.toString(), buffer, 4096)
+        symlinkTarget = buffer.toKString().toPath()
```
oldergod, line 52: 4096 is somewhat arbitrary

okio/src/unixMain/kotlin/okio/-UnixPosixVariant.kt
```diff
@@ -113,6 +114,22 @@ internal actual fun PosixFileSystem.variantOpenReadWrite(file: Path): FileHandle
...
+internal actual fun PosixFileSystem.variantCreateSymlink(source: Path, target: Path) {
+  if (source.parent == null || !exists(source.parent!!)) {
+    throw IOException("parent directory does not exist: ${source.parent}")
+  }
+
+  if (exists(source)) {
+    throw IOException("already exists: $source")
+  }
+
+  val result = symlink(target.toString(), source.toString())
```
oldergod, line 127: I cannot use canonicalize (as we're doing in FakeFS) because this method throws if the file doesn't exist.
Tests seem to pass. Could that be a problem?

okio/src/appleMain/kotlin/okio/-ApplePosixVariant.kt
swankjesse: Also, if the return value is -1, you should do this:
```
      throw errnoToIOException(errno)
```
https://pubs.opengroup.org/onlinepubs/009696699/functions/readlink.html

swankjesse: Also I think it’s nicer to pass the result to toKString():

```
symlinkTarget = buffer.toKString(endIndex = byteCount).toPath()
```

swankjesse: Replace 4096 with `platform.posix.PATH_MAX`

swankjesse (approved):

okio/src/unixMain/kotlin/okio/-UnixPosixVariant.kt
swankjesse: I think this looks right as-is.

swankjesse: Also FakeFileSystem isn’t calling `canonicalize()`, it’s merely converting a potentially-relative path to an absolute path and calling that canonical.

PR #1021, opened 2021-10-13
Symlinks for NodeJS

oldergod:


okio-nodefilesystem/src/main/kotlin/okio/NodeJsFileSystem.kt
```diff
@@ -30,27 +30,39 @@ object NodeJsFileSystem : FileSystem() {
...
       if (e.errorCode == "ENOENT") return null // "No such file or directory".
       throw IOException(e.message)
     }
+
+    var symlinkTarget: Path? = null
+    if ((stat.mode.toInt() and S_IFMT) == S_IFLNK) {
+      try {
+        symlinkTarget = readlinkSync(pathString).toPath()
+      } catch (e: Throwable) {
+        throw e.toIOException()
```
swankjesse, line 58: there’s an ugly race where the file is a symlink on the 1st call and not on the 2nd call. I don’t think we need to do anything about it.

swankjesse (approved):

PR #1022, opened 2021-10-14
Fix nodefilesystem dependency coordinate

ccjernigan:
Looking at the published artifacts, it appears to have been renamed for the alpha-10 release.

https://search.maven.org/artifact/com.squareup.okio/okio-nodefilesystem 
https://search.maven.org/artifact/com.squareup.okio/okio-nodefilesystem-js


ccjernigan:
@oldergod Done

PR #1023, opened 2021-10-14
API and FakeFS tests for mustCreate/mustExist

oldergod:


okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -1891,6 +2018,14 @@ abstract class AbstractFileSystemTest(
     }
   }
 
+  // TODO(Benoit) Remove once all filesystems support it.
+  private fun supportsMustCreateMustExist(): Boolean {
```
swankjesse, line 2022: Nice.

PR #1027, opened 2021-10-15
Add macosArm64 target

BoD:
This adds the `macosArm64` target.

This required depending on `kotlin-gradle-plugin:1.5.31` on the `build-support` module (as `macosArm64()` is a recent API), which prompted to remove its declaration from `Dependencies.kt` - which in turn allowed a few more simplifications (thanks @martinbonnin) as stdlib is automatically imported by the Kotlin plugin nowadays.

### DIFF
--- a/android-test/build.gradle.kts
+++ b/android-test/build.gradle.kts
@@ -65,7 +65,6 @@ dependencies {
   androidTestImplementation(deps.androidx.testExtJunit)
   androidTestImplementation(deps.androidx.testRunner)
   androidTestImplementation(deps.animalSniffer.annotations)
-  androidTestImplementation(deps.kotlin.stdLib.common)
   androidTestImplementation(deps.kotlin.test.annotations)
   androidTestImplementation(deps.kotlin.test.common)
   androidTestImplementation(deps.kotlin.test.jdk)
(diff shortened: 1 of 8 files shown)


build-support/build.gradle.kts
```diff
@@ -20,3 +20,7 @@ gradlePlugin {
     }
   }
 }
+
+dependencies {
+  implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.31")
```
oldergod, line 25: Can this file refer to `buildSrc`? It'd be nice to keep the kotlin version in the `Dependencies.kt`
Here wouldn't be a place people would look at to update versions

oldergod (commented):
👍

build-support/build.gradle.kts
BoD: Unfortunately, it doesn't look like so, when I tried, `deps` was not resolving.

oldergod: I didn't realize `build-support` was its own project. ~Maybe we could just move the dependencies over there and get rid of `buildSrc`... that'd be a follow-up~ that didn't work.

swankjesse:
Nice.

PR #1029, opened 2021-10-18
Fix two Linux bugs

swankjesse:
We were always appending to the end on FileHandle. This fix writes in
the appropriate place by changing open flags.

We weren't returning the symlink target.

Also fix file opens to be as atomic as possible on all UNIX systems.


oldergod (approved):
GREAT

PR #1065, opened 2021-11-02
Add watchosX64 target

BoD:
This is a followup to https://github.com/square/okio/pull/1027 and https://github.com/square/okio/pull/1028, and adds `watchosX64` (Apple watchOS 64-bit simulator for watchOS 7.0 and newer, on x86_64 platforms)

### DIFF
--- a/build-support/src/main/kotlin/platforms.kt
+++ b/build-support/src/main/kotlin/platforms.kt
@@ -13,6 +13,7 @@ fun KotlinMultiplatformExtension.configureOrCreateNativePlatforms() {
   watchosArm32()
   watchosArm64()
   watchosX86()
+  watchosX64()
   watchosSimulatorArm64()
   // Required to generate tests tasks: https://youtrack.jetbrains.com/issue/KT-26547
   linuxX64()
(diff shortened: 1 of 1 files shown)


martinbonnin (approved):

oldergod (approved):
Thank you

build-support/src/main/kotlin/platforms.kt
```diff
@@ -46,7 +48,7 @@ val linuxTargets = listOf(
 
 val nativeTargets = appleTargets + linuxTargets + mingwTargets
 
-/** Note that size_t is 32-bit on all watchOS versions (ie. pointers are always 32-bit). */
+/** Note that size_t is 32-bit on legacy watchOS versions (ie. pointers are always 32-bit). */
```
swankjesse, line 51: Why the change to the word legacy? Arm64 is the architecture of the current Apple watch.

BoD: Woops sorry, you're right, I think I got confused because `watchosX64` is 64 bits and specific to recent WatchOS, but that's only the simulator, not the actual watch.

PR #1069, opened 2021-11-16
Update Kotlin to 1.6

oldergod:


JakeWharton:
You need to use macos-11 worker

JakeWharton:
See https://github.com/cashapp/turbine/pull/76

oldergod:
Sweet. I wonder what's their timeline to get macos-latest to 11: https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md

I'm a bit puzzled by the removal of a windows platform on CI. Doesn't cost much to keep one.

JakeWharton:
The cost is significant build complexity and problems deploying because two Sonatype staging repos are created and must be released together. Turbine has no platform-specific code therefore there is no reason to build on multiple platforms.

JakeWharton (approved):

PR #1083, opened 2022-01-09
Common ForwardingSource.

yschimke:
Useful for okhttp.

### DIFF
--- a/okio/build.gradle.kts
+++ b/okio/build.gradle.kts
@@ -105,6 +105,9 @@ kotlin {
     }
     val nonJvmMain by creating {
       kotlin.srcDir("src/hashFunctions/kotlin")
+      dependencies {
+        dependsOn(commonMain)
+      }
     }
     val nonJvmTest by creating {
       dependencies {
--- a/okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
+++ b/okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
@@ -15,31 +15,18 @@
  */
 package okio
 
-import java.io.IOException
-
 /** A [Source] which forwards calls to another. Useful for subclassing. */
 actual abstract class ForwardingSource actual constructor(
   /** [Source] to which this instance is delegating. */
-  @get:JvmName("delegate")
   actual val delegate: Source
 ) : Source {
   // TODO 'Source by delegate' once https://youtrack.jetbrains.com/issue/KT-23935 is fixed.
 
-  @Throws(IOException::class)
   actual override fun read(sink: Buffer, byteCount: Long): Long = delegate.read(sink, byteCount)
 
   actual override fun timeout() = delegate.timeout()
 
-  @Throws(IOException::class)
   actual override fun close() = delegate.close()
 
-  actual override fun toString() = "${javaClass.simpleName}($delegate)"
-
-  @JvmName("-deprecated_delegate")
-  @Deprecated(
-    message = "moved to val",
-    replaceWith = ReplaceWith(expression = "delegate"),
-    level = DeprecationLevel.ERROR
-  )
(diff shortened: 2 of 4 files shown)


okio/build.gradle.kts
```diff
@@ -105,6 +105,9 @@ kotlin {
     }
     val nonJvmMain by creating {
       kotlin.srcDir("src/hashFunctions/kotlin")
+      dependencies {
```
yschimke, line 108: Bit curious why intellij is complaining for me without this.

okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
```diff
@@ -0,0 +1,32 @@
...
+package okio
+
+/** A [Source] which forwards calls to another. Useful for subclassing. */
+actual abstract class ForwardingSource actual constructor(
+  /** [Source] to which this instance is delegating. */
+  actual val delegate: Source
+) : Source {
+  // TODO 'Source by delegate' once https://youtrack.jetbrains.com/issue/KT-23935 is fixed.
+
+  actual override fun read(sink: Buffer, byteCount: Long): Long = delegate.read(sink, byteCount)
```
yschimke, line 25: Curious about exception standards, should these be marked here?

okio/src/jvmMain/kotlin/okio/ForwardingSource.kt
```diff
@@ -18,22 +18,22 @@ package okio
 import java.io.IOException
 
 /** A [Source] which forwards calls to another. Useful for subclassing. */
```
oldergod, line 20: I don't think we need comments and todos in here anymore?

okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+/** A [Source] which forwards calls to another. Useful for subclassing. */
```
oldergod, line 18: Again, I don't think we need comments and todos

okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
oldergod: I think it's only needed if this wasn't set already in the expect class. I would look at other actuals, like FileSystem to compare.

oldergod (commented):
Why do you need to create a nonJvmMain one if they have the same content? Will it differ in follow-up?

yschimke:
> Why do you need to create a nonJvmMain one if they have the same content? Will it differ in follow-up?

I removed a deprecated member

```

  @JvmName("-deprecated_delegate")
  @Deprecated(
    message = "moved to val",
    replaceWith = ReplaceWith(expression = "delegate"),
    level = DeprecationLevel.ERROR
  )
  fun delegate() = delegate
```

okio/src/nonJvmMain/kotlin/okio/ForwardingSource.kt
yschimke: Following FileSystem I put them in.

oldergod (approved):

okio/build.gradle.kts
```diff
@@ -105,6 +105,9 @@ kotlin {
...
+        dependsOn(commonMain)
```
JakeWharton, line 109: This goes outside the `dependencies { }` block

Issue #1093, opened 2022-02-03
AES/GCM encryption with CipherSink fails on Android


sndnv:
Basically, I have a piece of code that grabs a file, splits it into chunks and encrypts each one individually before sending everything over the network.

For creating the encryption sink, [`cipherSink`](https://github.com/square/okio/blob/5cf6a5b93985050b64f327bbc8e2daefbdaf2979/okio/src/jvmMain/kotlin/okio/JvmOkio.kt#L201) is called and then a bunch of segments are written; the unit tests are running just fine and working as expected but when I tested this on Android, I got the following exception:

```
java.lang.IllegalStateException: Unexpected output size 8208 for input size 16
        at okio.CipherSink.update(CipherSink.kt:53)
        at okio.CipherSink.write(CipherSink.kt:40)
        ...
```
After some digging, I managed to narrow it down to differences in behavior between the default JVM crypto provider and the one used on Android, when calls are made to `cipher.update(...)` for AEAD ciphers. Normally, calling `update` returns encrypted data but on Android the encrypted data is returned all at once, after a call to `cipher.doFinal(...)`; all previous calls to `update` buffer the data but return nothing. This is intentional - see [google/conscript#657](https://github.com/google/conscrypt/issues/657), [google/conscript#697](https://github.com/google/conscrypt/issues/697) so it's unlikely to change.

Now we get to the trouble with [`CipherSink.update(...)`](https://github.com/square/okio/blob/5cf6a5b93985050b64f327bbc8e2daefbdaf2979/okio/src/jvmMain/kotlin/okio/CipherSink.kt#L51-L56) (and, I assume `CipherSource` since the [code](https://github.com/square/okio/blob/5cf6a5b93985050b64f327bbc8e2daefbdaf2979/okio/src/jvmMain/kotlin/okio/CipherSource.kt#L64-L69) is similar):
```
    var outputSize = cipher.getOutputSize(size)
    while (outputSize > Segment.SIZE) {
      check(size > blockSize) { "Unexpected output size $outputSize for input size $size" }
      size -= blockSize
      outputSize = cipher.getOutputSize(size)
    }
```
[...]

PR #1096, opened 2022-03-03
Provide fallback for broken AES/GCM streaming implementation on Android

MartinDevi:
Hacky fix for #1093 to have a fallback implementation when the standard behavior fails.

There's unfortunately not tests to validate the bug fix. The two possibilities for it that I see would be:
1. Somehow, likely using a mocking framework to re-recreate the buggy `Cipher` behavior on the JVM, by caching all the  data provided to the instance through `update`, until `doFinal` is called
2. Set up tests on an Android emulator, somehow


swankjesse (approved):

swankjesse:
Sorry for sitting on this for forever.

PR #1107, opened 2022-03-26
Fix 'NSData.toByteString'

MJegorovas:
If you try to use `NSData.toByteString` with empty NSData method crashes with `kotlin.ArrayIndexOutOfBoundsException`

### DIFF
--- a/okio/src/appleMain/kotlin/okio/ByteString.kt
+++ b/okio/src/appleMain/kotlin/okio/ByteString.kt
@@ -23,7 +23,7 @@ import platform.posix.memcpy
 fun NSData.toByteString(): ByteString {
   val data = this
   val size = data.length.toInt()
-  return if (size != 0 ) {
+  return if (size != 0) {
     ByteString(
       ByteArray(size).apply {
         usePinned { pinned ->
--- a/okio/src/appleMain/kotlin/okio/ByteString.kt
+++ b/okio/src/appleMain/kotlin/okio/ByteString.kt
@@ -23,13 +23,15 @@ import platform.posix.memcpy
 fun NSData.toByteString(): ByteString {
   val data = this
   val size = data.length.toInt()
-  return ByteString(
-    ByteArray(size).apply {
-      if (size != 0 ) {
+  return if (size != 0 ) {
+    ByteString(
+      ByteArray(size).apply {
         usePinned { pinned ->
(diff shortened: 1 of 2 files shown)


okio/src/appleMain/kotlin/okio/ByteString.kt
```diff
@@ -22,10 +22,13 @@ import platform.posix.memcpy
 
 fun NSData.toByteString(): ByteString {
   val data = this
+  val size = data.length.toInt()
```
oldergod, line 25: if `ByteString.EMPTY` is available, let's return that directly when the size is 0.

MJegorovas: It is. Thanks

okio/src/appleTest/kotlin/okio/AppleByteStringTest.kt
```diff
@@ -28,4 +28,10 @@ class AppleByteStringTest {
     val byteString = data.toByteString()
     assertEquals("Hello", byteString.utf8())
   }
+
+  @Test fun emptyNsDataToByteString() {
+    val data = ("" as NSString).dataUsingEncoding(NSUTF8StringEncoding) as NSData
+    val byteString = data.toByteString()
+    assertEquals("", byteString.utf8())
```
Egorand, line 35: Does `assertEquals(ByteString.EMPTY, byteString)` work here?

Egorand (approved):
Thank you!

swankjesse (approved):

oldergod:
@MJegorovas looks like spotless is sad.
You can run `./gradlew spotlessApply` on your branch and repush, that should make it happy.

If you have not already, please [sign our CLA](https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) so we can merge the PR. Thank you

okio/src/appleTest/kotlin/okio/AppleByteStringTest.kt
MJegorovas: Was doing like the first test which was asserting with string. Will change.

oldergod:
Thank you

PR #1123, opened 2022-06-22
Add NSInputStream.source() and BufferedSource.inputStream() functions for Apple's NSInputStream

jeffdgr8:
Modeled after JVM's [`InputStream.source()`](https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/JvmOkio.kt#L80-L117) and [`BufferedSource.inputStream()`](https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/RealBufferedSource.kt#L142-L174), add extension functions to similarly map to and from Apple's [`NSInputStream`](https://developer.apple.com/documentation/foundation/nsinputstream).


JakeWharton:
FYI #1131 moves a Darwin-specific extension function onto the actual type's companion so that it becomes visible across the Objective-C bridge. We will want to do that same here, but that can be done in a follow-up.

jeffdgr8:
@JakeWharton excellent. I'll take a look at moving to a companion object. Looks like once #1131 is merged, we'll have to duplicate `actual BufferedSource` in **appleMain** and **nonAppleMain** source sets, adding the extension to the **appleMain** companion?

I'm assuming `NSInputStream.source()` would be appropriate on `Source`'s companion. There's only a single interface defined in common though. Does it need to be converted to expect/actual across all source sets for this change? And would actuals be in **jvm**, **nonAppleMain**, and **appleMain**?

I have these extensions working in my KMM database bindings with a local maven build and things are working well in API tests, by the way.

okio/src/appleMain/kotlin/okio/Source.kt
```diff
@@ -0,0 +1,66 @@
...
+import kotlinx.cinterop.UnsafeNumber
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.convert
+import kotlinx.cinterop.reinterpret
+import kotlinx.cinterop.usePinned
+import platform.Foundation.NSInputStream
+import platform.darwin.UInt8Var
+
+/** Returns a source that reads from `in`. */
+fun NSInputStream.source(): Source = NSInputStreamSource(this)
```
swankjesse, line 27: I think this file should be called `source.kt` (lowercase name) . . . we’ve been doing `lowerCamelFileNames` for files with free functions, `UpperCamelFileNames` when the file matches the primary declaration class.

okio/src/appleMain/kotlin/okio/Source.kt
```diff
@@ -0,0 +1,66 @@
...
+    require(byteCount >= 0L) { "byteCount < 0: $byteCount" }
+    val tail = sink.writableSegment(1)
+    val maxToCopy = minOf(byteCount, Segment.SIZE - tail.limit)
+    val bytesRead = tail.data.usePinned {
+      val bytes = it.addressOf(tail.limit).reinterpret<UInt8Var>()
+      input.read(bytes, maxToCopy.convert()).toLong()
+    }
+    if (bytesRead < 0) throw IOException(input.streamError?.localizedDescription)
+    if (bytesRead == 0L) {
+      if (tail.pos == tail.limit) {
```
swankjesse, line 49: Nice

okio/src/appleTest/kotlin/okio/AppleBufferedSourceTest.kt
```diff
@@ -0,0 +1,115 @@
...
+    assertTrue(source.closed)
+
+    val byteArray = ByteArray(4)
+    byteArray.usePinned {
+      val cPtr = it.addressOf(0).reinterpret<UInt8Var>()
+
+      byteArray.fill(-5)
+      assertEquals(-1, nsis.read(cPtr, 4))
+      assertNotNull(nsis.streamError)
+      assertEquals("closed", nsis.streamError?.localizedDescription)
```
swankjesse, line 111: Nice

swankjesse (approved):
Love this. Thanks for your patience on my slow review.

okio/src/appleMain/kotlin/okio/BufferedSource.kt
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  override fun read(buffer: CPointer<uint8_tVar>?, maxLength: NSUInteger): NSInteger {
+    try {
+      val internalBuffer = bufferedSource.buffer
+
+      if (bufferedSource is RealBufferedSource) {
+        if (bufferedSource.closed) throw IOException("closed")
+
+        if (internalBuffer.size == 0L) {
```
swankjesse, line 61: Please rewrite this if clause as  

```
if (bufferedSource.exhausted()) return 0
```

That will guarantee the buffer has at least one byte in it after!

okio/src/appleMain/kotlin/okio/BufferedSource.kt
```diff
@@ -0,0 +1,125 @@
...
+
+      val toRead = minOf(maxLength.toInt(), internalBuffer.size).toInt()
+      return internalBuffer.readNative(buffer, toRead).convert()
+    } catch (e: Exception) {
+      error = e.toNSError()
+      return -1
+    }
+  }
+
+  override fun getBuffer(
```
swankjesse, line 75: I don’t think this is a big deal for now, but eventually it might be a problem ..  . .

Using usePinned to get the address of a byte within a buffer is fine, but I worry that the caller might need usePinned when they later read from the buffer.

swankjesse: (This is only an issue if Kotlin’s GC ever learns to relocate buffers)

okio/src/appleMain/kotlin/okio/BufferedSource.kt
jeffdgr8: Oh, nice. I see `RealBufferedSource.commonExhausted()` is this exact logic. I went ahead and replaced this same logic in other places as well https://github.com/square/okio/pull/1123/commits/3571012c6ab78f1c41865afe2c763b9369727499. Let me know if you'd rather have that reverted though.

okio/src/appleMain/kotlin/okio/Source.kt
jeffdgr8: Should `BufferedSource.kt` and `ByteString.kt`, also in this directory, be renamed `bufferedSource.kt`and `byteString.kt` as well? Or perhaps `BufferedSource.kt` should be `inputStream.kt`?

okio/src/appleMain/kotlin/okio/BufferedSource.kt
jeffdgr8: I wondered if this could be problematic, but couldn't think of a way to pass the unpinning to the caller. I figure typical usage would read the buffer immediately after calling the function and wouldn't hold the reference longer. Still a minute chance GC occurs right between unpinning and reading though, and even tinier chance the address becomes invalid.

Copying the buffer would nullify the usefulness of the API and violate the documented behavior of [returning in 0(1)](https://developer.apple.com/documentation/foundation/nsinputstream). I figured having it implemented is probably preferable to not, but alternatively we could just return false.

I wonder what the behavior would be to `pin()` the address without `unpin()`ing it. Would GC still collect it once the reference is released, just disallow relocating it? Or would this create a memory leak?

```
val pinned = s.data.pin()
buffer?.pointed?.value = pinned.addressOf(s.pos).reinterpret()
length?.pointed?.value = (s.limit - s.pos).convert()
return true
```

swankjesse: The memory leak question is also a good one. At the moment this API doesn't tell the GC about this reference to the ByteArray, so it could be collected (or recycled) before the caller gets to use it.

Maybe we should pin it here, and/or take other steps to avoid recycling, then unpin either on the next call to this method or on close(). That would work and it wouldn't leak.

okio/src/appleMain/kotlin/okio/Source.kt
jeffdgr8: @swankjesse after merging the latest upstream changes, the linter doesn't like the lowercase filename `source.kt`, wanting PascalCase. I renamed this file `NSInputStreamSource.kt`. Let me know if there's a preferable name.

serandel:
Naive question, is this PR still in mind? I see an approval from last year, but I don't know if it needs more, or it has been discarded. Talking to Jeff, this is the only blocker to have a Kotlin Multiplatform Couchbase Lite library, which would be really awesome.

swankjesse:
I’d like to backport this once it lands:
https://github.com/Kotlin/kotlinx-io/pull/174

jeffdgr8:
> I’d like to backport this once it lands: [Kotlin/kotlinx-io#174](https://github.com/Kotlin/kotlinx-io/pull/174)

Great, I'm planning to implement the additions there on this Okio branch as well.

jeffdgr8:
I backported the changes from https://github.com/Kotlin/kotlinx-io/pull/174 including support for `NSOutputStream` sink extensions and `NSRunLoop` scheduling.

swankjesse:
Oh that’s awesome. Thanks!

okio/build.gradle.kts
```diff
@@ -132,6 +132,11 @@ kotlin {
           nativeTest.dependsOn(nonJvmTest)
           nativeTest.dependsOn(nonWasmTest)
           createSourceSet("appleTest", parent = nativeTest, children = appleTargets)
+            .apply {
+              dependencies {
+                implementation(libs.kotlin.coroutines)
```
PromanSEW, line 137: Is it used only in tests? Should be `testImplementation` instead

jeffdgr8: Yes, it is only used in tests. But this is specifically the `appleTest` source set `KotlinDependencyHandler` dependency management block. `testImplementation` is only valid in the top-level `dependencies` block, and applies specifically for the `test` source set, which isn't specific enough for a KMP module with multiple platform target `main` and `test` source sets.

okio/src/appleMain/kotlin/okio/BuffersApple.kt
```diff
@@ -0,0 +1,100 @@
...
+internal fun Buffer.write(source: CPointer<uint8_tVar>, maxLength: Int) {
+  require(maxLength >= 0) { "maxLength ($maxLength) must not be negative" }
+
+  var currentOffset = 0
+  while (currentOffset < maxLength) {
+    val tail = writableSegment(1)
+
+    val toCopy = minOf(maxLength - currentOffset, Segment.SIZE - tail.limit)
+    tail.data.usePinned {
+      memcpy(it.addressOf(tail.pos), source + currentOffset, toCopy.convert())
```
fzhinkin, line 50:
```suggestion
      memcpy(it.addressOf(tail.limit), source + currentOffset, toCopy.convert())
```

okio/src/appleMain/kotlin/okio/BufferedSource.kt
```diff
@@ -0,0 +1,196 @@
...
+      if (runLoop == this.runLoop) {
+        delegateOrSelf.stream(this, event)
+      }
+    }
+  }
+
+  private fun checkBytes() {
+    val runLoop = runLoop ?: return
+    runLoop.performInModes(runLoopModes) {
+      if (runLoop != this.runLoop || isFinished) return@performInModes
```
Khang-NT, line 146: I don't think checkBytes in runLoop is appropriate, runLoop could be main thread, and should be only used for event callback.

PR #1125, opened 2022-06-26
Make pom.xml depend on the JVM artifact

swankjesse:
This should prevent Maven users from needing to change
their dependencies from okio to okio-jvm. More importantly,
it should also prevent them from inadvertently getting two
different Okio JVM artifacts on their classpaths at once.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -2,6 +2,8 @@ import aQute.bnd.gradle.BundleTaskConvention
 import com.diffplug.gradle.spotless.SpotlessExtension
 import com.vanniktech.maven.publish.MavenPublishBaseExtension
 import com.vanniktech.maven.publish.SonatypeHost
+import groovy.util.Node
+import groovy.util.NodeList
 import java.nio.charset.StandardCharsets
 import org.gradle.api.tasks.testing.logging.TestExceptionFormat
 import org.gradle.api.tasks.testing.logging.TestLogEvent.FAILED
@@ -87,6 +89,7 @@ allprojects {
   }
 
   plugins.withId("com.vanniktech.maven.publish.base") {
+    val publishingExtension = extensions.getByType(PublishingExtension::class.java)
     configure<MavenPublishBaseExtension> {
       publishToMavenCentral(SonatypeHost.S01)
       signAllPublications()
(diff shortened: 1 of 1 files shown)


swankjesse:
We end up with this:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <!-- This module was also published with a richer model, Gradle metadata,  -->
  <!-- which should be used instead. Do not delete the following line which  -->
  <!-- is to indicate to Gradle or any Gradle module metadata file consumer  -->
  <!-- that they should prefer consuming it instead. -->
  <!-- do_not_remove: published-with-gradle-metadata -->
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.squareup.okio</groupId>
  <artifactId>okio</artifactId>
  <version>3.1.0</version>
  <name>okio</name>
  <description>A modern I/O API for Java</description>
  <url>https://github.com/square/okio/</url>
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>http://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
  <developers>
    <developer>
      <id>square</id>
      <name>Square, Inc.</name>
    </developer>
  </developers>
  <scm>
    <connection>scm:git:git://github.com/square/okio.git</connection>
    <developerConnection>scm:git:ssh://git@github.com/square/okio.git</developerConnection>
    <url>https://github.com/square/okio/</url>
  </scm>
  <dependencies>
    <dependency>
      <groupId>com.squareup.okio</groupId>
      <artifactId>okio-jvm</artifactId>
      <version>3.1.0</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

swankjesse:
🎉 I tested this manually with a Maven project and it worked as expected.

staktrace (approved):

PR #1130, opened 2022-07-11
Seal the `BufferedSource`/`BufferedSink` hierarchy

JakeWharton:
These interfaces are not meant for third-party implementation. The `Buffer` type and the type returned by `.buffer()` are the only valid implementations.


swankjesse (approved):
# 🦭

PR #1131, opened 2022-07-11
Move `NSData` to `ByteString` conversion onto `ByteString` companion

### LINKED ISSUE #1113: `ByteString` exposed to Objective-C / Swift cannot see `NSData` helper
The Kotlin/Native Objective-C headers only expose things transitively referenced from APIs which are public in the module producing the XCFramework. The `ByteStringKt` class which only exists for darwin platforms to hold the `NSData` extension therefore does not get exposed, and cannot even be explicitly exposed since Kotlin won't let you reference that implicit enclosing type.

It would be nice if these platform-specific conversion functions existed on the companion object so that if `ByteString` is exposed all of the niceties to create one for your platform are available. Doing so requires more granular actual's which just adds to the existing pain for lack of https://youtrack.jetbrains.com/issue/KT-20427/.

Still probably worth doing.

JakeWharton:
This is a convenience for Objeective-C/Swift callers such that when a `ByteString` is exposed in your header you are also given this (and any future) conversion functions as automatically available.

Because `SegmentedByteString` relies on `ByteString`'s internal constructor which is only avaialble in the actuals it must be duplicated in the two source sets.

Closes #1113


oldergod (approved):

okio/src/appleMain/kotlin/okio/ByteString.kt
```diff
@@ -1,39 +1,221 @@
 /*
- * Copyright (C) 2020 Square, Inc.
+ * Copyright (C) 2018 Square, Inc.
```
JakeWharton, line 2: This old file was deleted, the other file moved here with the same name, and the extension hand written at the bottom. It is not actually changed but git thinks it's being smart.

swankjesse (approved):

PR #1132, opened 2022-07-20
Check thread isInterrupted before deleting on JvmSystemFileSystem

Goooler:
If we use `JvmSystemFileSystem.deleteRecursively()` to delete a folder that contains sub-files, then interrupt the thread, we want the delete action will be interrupted earlier.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
@@ -124,7 +124,7 @@ internal open class JvmSystemFileSystem : FileSystem() {
   }
 
   override fun delete(path: Path, mustExist: Boolean) {
-    if (Thread.currentThread().isInterrupted) {
+    if (Thread.interrupted()) {
       // If the current thread has been interrupted.
       throw InterruptedIOException("interrupted")
     }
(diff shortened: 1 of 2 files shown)


okio/src/jvmMain/kotlin/okio/JvmSystemFileSystem.kt
```diff
@@ -123,6 +124,10 @@ internal open class JvmSystemFileSystem : FileSystem() {
   }
 
   override fun delete(path: Path, mustExist: Boolean) {
+    if (Thread.currentThread().isInterrupted) {
```
swankjesse, line 127: Should be `interrupted()`, not `isInterrupted`.

Goooler: ae4d73fe1d05ee0ad53f7ab2469ebba9fea20603

swankjesse:
Fantastic. Wanna add a test case? Good to merge either way.

okio/src/jvmTest/kotlin/okio/JvmSystemFileSystemTest.kt
```diff
@@ -35,4 +40,16 @@ class JvmSystemFileSystemTest : AbstractFileSystemTest(
...
+    Thread.currentThread().interrupt()
+    try {
+      fileSystem.delete(base)
+      fail()
+    } catch (expected: InterruptedIOException) {
+      assertEquals("interrupted", expected.message)
+      assertFalse(Thread.interrupted())
+    }
+  }
+}
```
Goooler, line 55: If I need to copy this test into `NioSystemFileSystemTest`?

Goooler: Ready to merge?

swankjesse (approved):

PR #1134, opened 2022-07-31
Adopt Gradle version catalogs in Okio

swankjesse:
Also upgrade to Gradle 7.5 where version catalogs are on by default.


gradle/libs.versions.toml
```diff
@@ -0,0 +1,31 @@
...
+ktlint = "0.42.1"
+
+[libraries]
+android-gradle-plugin = { module = "com.android.tools.build:gradle", version = "4.2.2" }
+android-desugar-jdk-libs = { module = "com.android.tools:desugar_jdk_libs", version = "1.1.5" }
+androidx-test-ext-junit = { module = "androidx.test.ext:junit", version = "1.1.3" }
+androidx-test-runner = { module = "androidx.test:runner", version = "1.4.0" }
+kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test" }
+kotlin-test-junit = { module = "org.jetbrains.kotlin:kotlin-test-junit" }
+kotlin-time = { module = "org.jetbrains.kotlinx:kotlinx-datetime", version = "0.3.2" }
```
swankjesse, line 12: We need this to be 0.3.3 in Zipline to work around a bug. Rather than upgrading it here, I wanted to organize this so a robot could do that for me.

buildSrc/build.gradle.kts
```diff
@@ -1,7 +0,0 @@
-plugins {
```
swankjesse, line 1: No more buildSrc for you, JW

build-support/src/main/kotlin/kmp.kt
```diff
@@ -0,0 +1,5 @@
+// If false - JS targets will not be configured in multiplatform projects.
+val kmpJsEnabled = System.getProperty("kjs", "true").toBoolean()
+
+// If false - Native targets will not be configured in multiplatform projects.
+val kmpNativeEnabled = System.getProperty("knative", "true").toBoolean()
```
oldergod, line 5: sweet

oldergod (approved):

buildSrc/build.gradle.kts
JakeWharton: Thank you, JW

PR #1147, opened 2022-08-03
Fix ABA race in JVM SegmentPool

ejona86:
"Acquire" LOCK for recycle() instead of using compareAndSet(). This
avoids an ABA problem. This change increases the number of atomic writes
during recycle() from 1 to 2, but doesn't really influence the
concurrency level.

ThreadSanitizer detected a race on the JVM for `Segment.limit` with okio
3.1. The write is setting `limit = 0` and the read is computing the pool
size.
```
WARNING: ThreadSanitizer: data race (pid=823)
  Write of size 4 at 0x0000cea2e4f0 by thread T27:
    #0 okio.SegmentPool.take()Lokio/Segment; SegmentPool.kt:94
    #1 okio.Buffer.writableSegment$(I)Lokio/Segment; Buffer.kt:1444
    #2 okio.InputStreamSource.read(Lokio/Buffer;J)J JvmOkio.kt:92
    #3 okio.AsyncTimeout$source$1.read(Lokio/Buffer;J)J AsyncTimeout.kt:125
    #4 okio.RealBufferedSource.request(J)Z RealBufferedSource.kt:206
    #5 okio.RealBufferedSource.require(J)V RealBufferedSource.kt:199
    ...

  Previous read of size 4 at 0x0000cea2e4f0 by thread T23:
    #0 okio.SegmentPool.recycle(Lokio/Segment;)V SegmentPool.kt:109
    #1 okio.OutputStreamSink.write(Lokio/Buffer;J)V JvmOkio.kt:65
    #2 okio.AsyncTimeout$sink$1.write(Lokio/Buffer;J)V AsyncTimeout.kt:99
    ...
```

That specific race is benign as the segment will be discarded in
recycle(). However, there are other races with `limit` that this will
hide. In particular, lots of things become possible if you consider the
ABA problem that can happen if a segment is take()n and recycle()d
within the race window in recycle() between get() and compareAndSet().
With such races, it is possible for limit to be smaller than the
segments in the pool. For example:

```
Assume there is one segment in the pool.

Thread A                      Thread B
// in recycle():
val firstRef = firstRef()
val first = firstRef.get()
if (first === LOCK) return
// will be zero
val firstLimit = first?.limit ?: 0
if (firstLimit >= MAX_SIZE) return
                              s1 = take() // firstRef.get() is now null
                              s2 = take() // allocates
                              s3 = take() // allocates
                              recycle(s2)
                              recycle(s3)
                              recycle(s1)
segment.next = first
segment.pos = 0
// firstLimit is out-of-date and too small
segment.limit = firstLimit + Segment.SIZE
// compareAndSet succeeds
if (!firstRef.compareAndSet(first, segment))
```

Such races aren't generally that scary, as they will self-heal rapidly,
but they are legitimate races and a fix is easy.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
+++ b/okio/src/jvmMain/kotlin/okio/SegmentPool.kt
@@ -28,7 +28,9 @@ import java.util.concurrent.atomic.AtomicReference
  * On [take], a caller swaps the stack's next pointer with the [LOCK] sentinel. If the stack was
  * not already locked, the caller replaces the head node with its successor.
  *
- * On [recycle], a caller swaps the head with a new node whose successor is the replaced head.
+ * On [recycle], a caller swaps the stack's next pointer with the [LOCK] sentinel. If the stack was
+ * not already locked, the caller replaces the head node with a new node whose successor is the
+ * replaced head.
  *
  * On conflict, operations succeed, but segments are not pushed into the stack. For example, a
  * [take] that loses a race allocates a new segment regardless of the pool size. A [recycle] call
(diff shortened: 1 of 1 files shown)


swankjesse:
Great analysis, great fix. Thank you.

PR #1154, opened 2022-08-10
Remove workaround and publish sha256/sha512 checksums

martinbonnin:
Pretty sure this is not needed anymore as it was fixed by Sonatype. I've been running my projects without this and it's working fine these days.


JakeWharton:
We probably cargo culted this to about 20 projects. Keep 'em coming, if you want!

JakeWharton (approved):

PR #1181, opened 2022-12-31
Attempt to make filesystem tests less flaky

swankjesse:
I've been observing failures like this:

    expected 2022-12-31T16:00:40.999Z in 2022-12-31T16:00:41Z..2022-12-31T16:00:43Z (relaxed from 2022-12-31T16:00:41.002314Z..2022-12-31T16:00:41.002365Z)

### DIFF
--- a/okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
+++ b/okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
@@ -15,6 +15,12 @@
  */
 package okio
 
+import kotlinx.datetime.Clock
+import kotlinx.datetime.Instant
+import okio.ByteString.Companion.encodeUtf8
+import okio.ByteString.Companion.toByteString
+import okio.Path.Companion.toPath
+import okio.fakefilesystem.FakeFileSystem
 import kotlin.test.BeforeTest
 import kotlin.test.Ignore
 import kotlin.test.Test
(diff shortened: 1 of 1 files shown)


okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -2356,20 +2357,28 @@ abstract class AbstractFileSystemTest(
...
    *
+   * It also pads the result by 200 milliseconds because the host and file system may use different
+   * clocks, allowing the time on the CPU to drift behind the time on the file system.
+   *
    * https://docs.microsoft.com/en-us/windows/win32/sysinfo/file-times
    */
   private fun Instant.maxFileSystemTime(): Instant {
-    return Instant.fromEpochSeconds(plus(2.seconds).epochSeconds)
+    val paddedInstant = plus(200.milliseconds)
+    return Instant.fromEpochSeconds(paddedInstant.plus(2.seconds).epochSeconds)
```
yschimke, line 2381: Why go via epoch seconds?

yschimke (approved):
Seems relatively safe.

okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
swankjesse: It’s attempting to drop precision, consistent with what FAT32 and friends do.

Issue #1187, opened 2023-01-03
Okio use block is return unsafe


yschimke:
A non local return from the inline function was skipping the close.

https://github.com/square/okio/blob/039d996d8d04bcd7ecfcbf1da82b62010bd0a64f/okio/src/commonMain/kotlin/okio/Okio.kt#L51-L70

```
inline fun <T : Closeable?, R> T.use(block: (T) -> R): R {
```

See issue here https://github.com/square/okhttp/pull/7625/files#diff-c7abbac71d83fc93c0d3f76dbc609a0b64426b7a3f9a8548b1f1723b3e0a277c

yschimke:
Closed in https://github.com/square/okio/pull/1190

PR #1188, opened 2023-01-04
Ensure close is called in all use cases.

yschimke:
A non local return from the inline function was skipping the close.

Addresses https://github.com/square/okio/issues/1187

### DIFF
--- a/okio/src/commonTest/kotlin/okio/FakeFileSystemTest.kt
+++ b/okio/src/commonTest/kotlin/okio/FakeFileSystemTest.kt
@@ -15,6 +15,8 @@
  */
 package okio
 
+import okio.Path.Companion.toPath
+import okio.fakefilesystem.FakeFileSystem
 import kotlin.reflect.KClass
 import kotlin.test.Test
 import kotlin.test.assertEquals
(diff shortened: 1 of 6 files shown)


okio/src/commonTest/kotlin/okio/FakeFileSystemTest.kt
```diff
@@ -488,6 +488,21 @@ abstract class FakeFileSystemTest internal constructor(
     }
   }
 
+  @Test
+  fun closesWithUseBlock() {
+    fun testMethodWithUse() {
+      val sink = fakeFileSystem.sink(base / "all-files-includes-file")
+
+      sink.use {
+        return
```
swankjesse, line 497:
```suggestion
        return@testMethodWithUse
```

okio/src/commonTest/kotlin/okio/FakeFileSystemTest.kt
```diff
@@ -488,6 +488,21 @@ abstract class FakeFileSystemTest internal constructor(
     }
   }
 
+  @Test
+  fun closesWithUseBlock() {
+    fun testMethodWithUse() {
+      val sink = fakeFileSystem.sink(base / "all-files-includes-file")
```
swankjesse, line 494: Does this test have anything to do with FakeFileSystem? Seems like we should have a test for `use`.

swankjesse (approved):

swankjesse:
Ooooh I hate that this fix works.

swankjesse:
Ooooh I hate that this fix is necessary.

okio/src/commonTest/kotlin/okio/FakeFileSystemTest.kt
yschimke: Yep, you got me. :)

PR #1198, opened 2023-01-08
Always run Gradle on JDK 19

swankjesse:


swankjesse:
I’m hoping to unblock this renovate job:
https://github.com/square/okio/pull/1191

JakeWharton (approved):

yschimke (approved):
This makes a lot of sense. Also simpler to have same approach across all your active projects.

PR #1201, opened 2023-01-14
Enable automatic release

swankjesse:


JakeWharton:
Will this work since Okio publishes from two actions at once? Also is the windows publish still needed? Kotlin added cross compilation to windows from Mac and Linux a few versions ago.

swankjesse:
I expect it to work. The manual process requires me to promote two staging repositories. I expect this to work similarly.

JakeWharton (approved):

PR #1204, opened 2023-01-15
Get style from .editorconfig

swankjesse:
Also use Redwood's .editorconfig content and layout

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -150,9 +150,6 @@ subprojects {
   configure<SpotlessExtension> {
     kotlin {
       target("**/*.kt")
-      ktlint(libs.versions.ktlint.get()).userData(mapOf("indent_size" to "2"))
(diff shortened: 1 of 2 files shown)


build.gradle.kts
```diff
@@ -150,9 +150,6 @@ subprojects {
   configure<SpotlessExtension> {
     kotlin {
       target("**/*.kt")
-      ktlint(libs.versions.ktlint.get()).userData(mapOf("indent_size" to "2"))
```
JakeWharton, line 153: You still need `ktlint()` here

JakeWharton (approved):

build.gradle.kts
swankjesse: Very nice catch. I might be soured on DSLs.

JakeWharton: Only know because I did it first!

PR #1213, opened 2023-01-29
Update Spotless and Ktlint, and appease these robots

swankjesse:
I think it’s worse in general, but I’d rather have the inelegant formatting of a robot than wasting human attention on whitespace.


okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -461,33 +463,61 @@ internal inline fun Buffer.commonWriteDecimalLong(v: Long): Buffer {
...
-    else if (v < 1000000000000000L)
-      if (v < 10000000000000L) 13
-      else if (v < 100000000000000L) 14
-      else 15
-    else if (v < 100000000000000000L)
-      if (v < 10000000000000000L) 16
-      else 17
-    else if (v < 1000000000000000000L) 18
-    else 19
+    if (v < 100000000L) {
```
swankjesse, line 466: Yech

okio/src/jvmTest/kotlin/okio/BufferFactory.kt
```diff
@@ -59,7 +59,8 @@ enum class BufferFactory {
 
       return bufferWithRandomSegmentLayout(dice, largeByteArray)
     }
-  };
+  },
+  ;
```
swankjesse, line 63: Arguably a ktlint bug

JakeWharton: I think this looks right. Trailing comma means clean diffs. Semicolon is required to separate constants from functions.

JakeWharton: I think this looks right. Trailing comma means clean diffs. Semicolon is required to separate constants from functions.

okio/src/commonMain/kotlin/okio/internal/Buffer.kt
JakeWharton: Both versions are pretty bad.

JakeWharton (approved):

PR #1224, opened 2023-02-11
Add extension & nameWithoutExtension for Path

Goooler:
Ref:
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/extension.html https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io.path/java.nio.file.-path/name-without-extension.html


okio/src/commonMain/kotlin/okio/Path.kt
```diff
@@ -305,5 +305,17 @@ expect class Path internal constructor(bytes: ByteString) : Comparable<Path> {
...
+     * Returns the extension of this path (not including the dot),
+     * or an empty string if it doesn't have one.
+     */
+    val Path.extension: String
+
+    /**
+     * Returns the name of this file or directory without an extension,
+     * or an empty string if this path has zero path elements.
+     */
+    val Path.nameWithoutExtension: String
```
Goooler, line 319: I'm not sure if we should make them as fields or extensions.

JakeWharton:
Jesse and I briefly talked about this a long time ago. I was opposed to these at the time because they name it seems like an extension is well defined when usually there's a bunch of edge cases that you need to think about (and may want to handle differently). Here's what I wrote then:

>  The implementation leans me toward no. Plus if you’re doing something generally, such as displaying a list of files, you already have to special case multiple dots (.tar.gz) and dotfiles (.bash_profile) so what does it really buy you?

I'll leave it up to him, though.

swankjesse:
I’ve got a blog post about whether file systems are UIs or implementation details, and my gripe about extensions is related to that. https://publicobject.com/2020/10/06/files/

In particular, I think extensions are part of the UI-model of file systems. Application developers wanted their files to have appropriate icons beside them, and the file extension is the mechanism.

I think this model still exists! But we’re collectively moving past it. . . . more data is in databases vs. files, for example. 

Anyway, I’m reluctant to do this 'cause I’m unsure of the use case. I hate the idea of using the file extension to determine whether an application can do something or not.

Could you tell me your use case?

Goooler:
I use these extensions for displaying file icons in a file manager. Using named based extensions instead of reading metadata to show icons is a more general step when we just want to show a file list, sure, we must read metadata when displaying a photo even if it ends with `.zip`.

swankjesse:
I’m gonna close this 'cause I’m not ready to do all the work to do file extensions well. In particular, doing reasonable things with files like `.DS_Store` makes me anxious. It seems simpler to leave this tiny detail up to the users.

PR #1231, opened 2023-02-19
Support Duration in Timeout

### LINKED ISSUE #1227: Kotlin Duration and DurationUnit support
Hi Okio team, thanks for this amazing IO library !

Expressing the timeout as timeout (Long) + DurationUnit would allow to declare and access it in common module, be the Kotlin way of expressing a duration, and it is possible to add this new feature without any API change for Java users so it does not leak.

Below I provide 2 options to this feature, what do you think ?

### Option 1 : keep `timeoutNanos: Long` internally as now
in commonMain
```kotlin
expect open class Timeout {
  internal var timeoutNanos: Long
  // other stuff
}

// both methods below are extensions, so they are usable in Kotlin but not visible from Java
fun Timeout.timeout(timeout: Long, unit: DurationUnit): Timeout {
  require(timeout >= 0) { "timeout < 0: $timeout" }
  timeoutNanos = timeout.toDuration(unit).inWholeNanoseconds
  return this
}

val Timeout.timeoutNanoseconds: Long
  get() = timeoutNanos
```
in jvmMain
```kotlin
actual open class Timeout {
  internal actual var timeoutNanos = 0L
  // other stuff
```
### Option 2 : use `timeout: Duration` internally
TimeUnit to DurationUnit conversion is now [stable from Kotlin 1.8.0](https://kotlinlang.org/docs/whatsnew18.html#timeunit-conversion-between-java-and-kotlin)

in commonMain
```kotlin
expect open class Timeout {
  internal var timeout: Duration
}

// both methods below are extensions, so they are usable in Kotlin but not visible from Java
fun Timeout.timeout(timeout: Long, unit: DurationUnit): Timeout {
  require(timeout >= 0) { "timeout < 0: $timeout" }
  this.timeout = timeout.toDuration(unit)
  return this
}

val Timeout.timeout: Duration
  get() = timeout

```
in jvmMain
```kotlin
actual open class Timeout {
  internal actual var timeout = Duration.ZERO

  open fun timeout(timeout: Long, unit: TimeUnit): Timeout {
    require(timeout >= 0) { "timeout < 0: $timeout" }
    this.timeout = timeout.toDuration(unit.toDurationUnit())
    return this
  }
  // other stuff
```

This issue is a fresh thinking around the same subject of #1084

Goooler:
Closes #1227.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Timeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/Timeout.kt
@@ -19,6 +19,9 @@ import java.io.IOException
 import java.io.InterruptedIOException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.Condition
+import kotlin.time.Duration
+import kotlin.time.DurationUnit
+import kotlin.time.toTimeUnit
 
 actual open class Timeout {
   /**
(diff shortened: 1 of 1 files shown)


swankjesse (approved):
Thanks!

PR #1235, opened 2023-02-28
doc: Fix swapped NioPath.sink/source javadoc

I-Al-Istannen:
I noticed this when trying to write to a path. I believe the javadocs have been swapped.


swankjesse (approved):
Thanks!!

PR #1239, opened 2023-03-12
Make UnsafeCursor implement Closable interface in multiplatform (Fixes #1238)

### LINKED ISSUE #1238: UnsafeCursor does not implement Closable in common, js and native targets
UnsafeCursor does not implement Closable in [common](https://github.com/square/okio/blob/37e2f284cf62a1253633ae40e5f84b976c57ea0d/okio/src/commonMain/kotlin/okio/Buffer.kt#L340) target (therefore in js and native too). Jvm target has its [own class](https://github.com/square/okio/blob/37e2f284cf62a1253633ae40e5f84b976c57ea0d/okio/src/jvmMain/kotlin/okio/Buffer.kt#L610) with Closable implementation.

UnsafeCursor does not implement Closeable while docs recommend to use non-multiplatform `.use` extension for UnsafeCursor.

inklesspen1rus:
While docs recommend to use `Closable.use` extension to manage UnsafeCursor as resource, `Closable.use` extension is not applicable to UnsafeCursor because UnsafeCursor does not implement Closable interface.

This PR contains:

1. Multiplatform test to ensure Unsafe Cursor implements Closeable interface
2. Commit that fixes #1238 and ensures the test above

### DIFF
--- a/okio/src/commonMain/kotlin/okio/Buffer.kt
+++ b/okio/src/commonMain/kotlin/okio/Buffer.kt
@@ -337,7 +337,7 @@ expect class Buffer() : BufferedSource, BufferedSink {
    * You can reuse instances of this class if you like. Use the overloads of [Buffer.readUnsafe] and
    * [Buffer.readAndWriteUnsafe] that take a cursor and close it after use.
    */
-  class UnsafeCursor constructor() {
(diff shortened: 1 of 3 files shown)


swankjesse (approved):
👍🏻 thanks!

PR #1251, opened 2023-04-20
Import `AssetFileSystem`

### LINKED ISSUE #1097: Assets filesystem for Android platform
I _think_ this should be possible, but I'm curious what APIs won't work. I suspect the fact that the zip filesystem exists means we can pull this off, though.

JakeWharton:
Which wraps a `FileSystem` around Android's assets... obviously.

Closes #1097


okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
```diff
@@ -0,0 +1,156 @@
...
+import okio.Source
+import okio.source
+
+/**
+ * Expose this [AssetManager] as an Okio [FileSystem].
+ *
+ * Note: Assets are a read-only view on a file system and so any attempt to mutate
+ * will throw an [UnsupportedOperationException].
+ */
+fun AssetManager.asFileSystem(): FileSystem = AssetFileSystem(this)
```
swankjesse, line 36: 👍🏻

okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
```diff
@@ -0,0 +1,156 @@
...
+  override fun listOrNull(dir: Path): List<Path>? {
+    return try {
+      list(dir)
+    } catch (_: IOException) {
+      null
+    }
+  }
+
+  override fun openReadOnly(file: Path): FileHandle {
+    TODO("not implemented yet!")
```
swankjesse, line 118: Should do

okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
```diff
@@ -0,0 +1,156 @@
...
+  override fun openReadWrite(file: Path, mustCreate: Boolean, mustExist: Boolean): FileHandle {
+    throw UnsupportedOperationException("asset file systems are not writable")
+  }
+
+  override fun source(file: Path): Source {
+    return assets.open(canonicalizeInternal(file).toAssetRelativePathString()).source()
+  }
+
+  override fun sink(file: Path, mustCreate: Boolean): Sink {
+    throw UnsupportedOperationException("asset file systems are read-only")
```
swankjesse, line 130: Please replace these with IOException, for consistency with the `FileSystem` docs and `ResourceFileSystem`, which does this:

```
    throw IOException("$this is read-only")
```

okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
JakeWharton: When I wrote this however many years ago I'm pretty sure the resource file system didn't support it so I also decided not to. I see that has since changed, so I'll do an impl as well.

okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
```diff
@@ -0,0 +1,156 @@
...
+          isDirectory = true,
+        )
+      }
+    }
+    return null
+  }
+
+  override fun list(dir: Path): List<Path> {
+    val canonical = canonicalizeInternal(dir)
+    if (canonical.existsInternal()) {
```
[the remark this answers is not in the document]
JakeWharton: Yes the API for assets is not great. Definitely designed around knowing ahead-of-time what files you want and not for arbitrary querying.

okio-assetfilesystem/src/main/kotlin/okio/assetfilesystem/AssetFileSystem.kt
JakeWharton: I'm going to do this in a follow-up because it's much more complicated than I expected. Assets do not support random access so I've had to fake it.

Issue #1121, opened 2022-06-20
Add `okio-jvm` to the `okio-bom`


rickie:
We are working on the upgrade of `okio` and decided to use the `okio-bom`. 
As we are using Maven, we also had to switch to `okio-jvm` (see related issue https://github.com/square/okio/issues/1067).

However, now that we switched to the BOM, we also had to specify the `okio-jvm` dependency to make sure it uses the same version as in the `okio-bom`.
Would it be possible to add the `okio-jvm` in the `okio-bom` such that the dependency is provided "out of the box"?

bogdandrutu:
Can this be fixed please?

Stephan202:
Hey @Egorand! This issue is still relevant for users of the [maven-dependency-plugin](https://maven.apache.org/plugins/maven-dependency-plugin/analyze-only-mojo.html), as explicit references to a dependency require said dependency to be declared explicitly:
```
[INFO] --- dependency:3.5.0:analyze-only (analyze-dependencies) @ messaging-support ---
[ERROR] Used undeclared dependencies found:
[ERROR]    com.squareup.okio:okio-jvm:jar:3.3.0:runtime
```

So it'd be nice if `okio-jvm` can be added to `okio-bom` :pray:

rickie:
I filed: https://github.com/square/okio/pull/1256.

PR #1269, opened 2023-06-18
Collect BOM constraints automatically

swankjesse:
Closes: https://github.com/square/okio/issues/1121

### DIFF
--- a/build-support/src/main/kotlin/bom.kt
+++ b/build-support/src/main/kotlin/bom.kt
@@ -1,0 +1,45 @@
+import org.gradle.api.Project
+import org.gradle.api.artifacts.dsl.DependencyConstraintHandler
+import org.gradle.kotlin.dsl.getByType
+import org.gradle.kotlin.dsl.withType
+import org.jetbrains.kotlin.gradle.dsl.KotlinJsProjectExtension
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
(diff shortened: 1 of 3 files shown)


build-support/src/main/kotlin/bom.kt
```diff
@@ -0,0 +1,45 @@
...
+import org.jetbrains.kotlin.gradle.plugin.KotlinTarget
+
+/**
+ * Collect all the root project's multiplatform targets and add them to the BOM.
+ *
+ * Only published subprojects are included.
+ *
+ * This supports Kotlin/Multiplatform and Kotlin/JS subprojects.
+ */
+fun Project.collectBomConstraints() {
```
swankjesse, line 18: FYI @oldergod we might wanna copy-paste this into our other repos? Or maybe submit it as a new feature for the Vanniktech plugin? Note that I don’t yet support non-multiplatform Kotlin/JVM subprojects.

swankjesse: Oooh, just realized I gotta support the Kotlin-Android plugin to capture the AssetFileSystem.

oldergod: I like it very much indeed

oldergod (approved):

build-support/src/main/kotlin/bom.kt
swankjesse: Done.

PR #1271, opened 2023-06-19
Add abstract test with default nio FileSystem

oldergod:


okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -1355,10 +1356,15 @@ abstract class AbstractFileSystemTest(
     val to = base / "to.txt"
     from.writeUtf8("source file")
     to.writeUtf8("target file")
-    expectIOExceptionOnWindows {
```
oldergod, line 1358: this method only would fail on Windows but all the others "windows limitations" would not apply so I had to make a one case exception :/

oldergod:
rebased on master

swankjesse (approved):

PR #1272, opened 2023-06-19
Invert supportsFileHandle and supportsSymlink

oldergod:
This will force new implementation to pay attention.


okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
```diff
@@ -2312,28 +2312,20 @@ abstract class AbstractFileSystemTest(
...
-      "JvmSystemFileSystem",
-      "NioSystemFileSystem",
-      "NioFileSystemWrappingFileSystem",
-      "PosixFileSystem",
-      "NodeJsFileSystem",
-      -> true
-      else -> false
+      "AddFileSystemNameIfNotSupportingAny",
+      -> false
+      else -> true
```
oldergod, line 2317: Or maybe we could simply remove it?

PR #1277, opened 2023-06-21
Fix conversion of signed constants to unsigned

sifmelcara:
Signed integer constant cannot be automatically converted to unsigned integer anymore after Kotlin 1.9.0.

See https://youtrack.jetbrains.com/issue/KT-56583 for detail

### DIFF
--- a/okio/src/nativeMain/kotlin/okio/SizetVariant.kt
+++ b/okio/src/nativeMain/kotlin/okio/SizetVariant.kt
@@ -29,11 +29,11 @@ internal fun variantFread(
   target: CPointer<ByteVarOf<Byte>>,
   byteCount: UInt,
   file: CPointer<FILE>,
-): UInt = fread(target, 1, byteCount.convert(), file).convert()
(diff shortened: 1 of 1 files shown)


JakeWharton:
I had no idea you could use lowercase letters. I think we should always use uppercase because a lowercase L can be mistaken for the number 1.

JakeWharton:
Although weirdly in reviewing some of my own code I like lowercase `f` apparently.

Also fun/horrific fact: You can write `1uL` 😩

oldergod:
We don't need test to cover that?

sifmelcara:
Examples in official doc uses lower case `u` https://kotlinlang.org/docs/unsigned-integer-types.html#unsigned-integers-literals , but I can change it to upper case if you prefer

JakeWharton:
No. Thanks for the link, though. I'll discuss it somewhere else and won't hold this up any longer. Then maybe we can get enforcement from ktlint.

PR #1279, opened 2023-06-22
Revert public API breaking change

oldergod:
The changed happened here https://github.com/square/okio/commit/cd31c079e6935f4b8035baa989acff5f0caa6c7f#diff-aa9f6ad4ed05efc263a9cc5ab076514a3f851f6b0573d76580a99aa31fe5d21a 
This introduced a API breaking change from out latest tag `3.3.0`. Even though the package is internal, methods are public.


okio/api/okio.api
```diff
@@ -800,7 +800,7 @@ public final class okio/Utf8 {
 	public static synthetic fun size$default (Ljava/lang/String;IIILjava/lang/Object;)J
 }
 
-public final class okio/internal/-Utf8 {
+public final class okio/internal/_Utf8Kt {
```
swankjesse, line 803: Don’t we need a `-` here?

swankjesse: I think this is what it was doing before

swankjesse (approved):

okio/api/okio.api
oldergod: No, the API dump on `3.3.0` gives us

```
public final class okio/internal/_Utf8Kt {
	public static final fun commonAsUtf8ToByteArray (Ljava/lang/String;)[B
	public static final fun commonToUtf8String ([BII)Ljava/lang/String;
	public static synthetic fun commonToUtf8String$default ([BIIILjava/lang/Object;)Ljava/lang/String;
}
```

okio/src/commonMain/kotlin/okio/internal/-Utf8.kt
```diff
@@ -13,11 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@file:JvmName("-Utf8") // A leading '-' hides this class from Java.
```
oldergod, line 16: Here's the api dump from `3.3.0`

```
public final class okio/_JvmPlatformKt {
	public static final fun withLock (Ljava/util/concurrent/locks/ReentrantLock;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
}
```

swankjesse (approved):

PR #1285, opened 2023-07-07
Copyright to files in build-support

oldergod:
Was it on purpose?


Egorand (approved):
Probably an omission. You can configure spotless to check for missing license headers and auto-add them, see how KotlinPoet does it!

swankjesse (approved):

PR #1291, opened 2023-07-10
Convert remaining .java tests to .kt

swankjesse:
Some of these were there to confirm Java APIs worked. At this point that’s just testing that Kotlin compiler’s implementation of features like `@JvmName`. Not building Java makes it easier to test on more non-JVM platforms.


swankjesse:
This PR uses 2 commits so the conversion and rename can be split into separate steps.

oldergod (approved):

PR #1295, opened 2023-07-20
Fix FileHandle read/write with empty array

MJegorovas:
If you try to use `FileHandle.read/write` with empty ByteArray on Native, method crashes with kotlin.ArrayIndexOutOfBoundsException.


oldergod (approved):
🍠 thank you

swankjesse (approved):

swankjesse:
# 💯 Good catch. Thanks!

PR #1300, opened 2023-07-22
Read files with WASM

swankjesse:


oldergod (approved):

okio-wasifilesystem/src/wasmMain/kotlin/okio/internal/Wasi.kt
```diff
@@ -42,17 +42,33 @@ internal fun Pointer.readString(byteCount: Int): String {
...
+  }
+  return sink
+}
+
 internal fun MemoryAllocator.write(
   string: String,
 ): Pair<Pointer, size> {
   val bytes = string.encodeToByteArray()
-  return write(bytes) to bytes.size
+  val result = allocate(bytes.size + 1)
```
swankjesse, line 64: This was a bug before! Linux gets really wrecked unless we add a trailing \0 when writing strings

swankjesse: (even though we also provide the string length elsewhere)

swankjesse: FYI @sdeleuze I needed a trailing `\0` on strings for NodeJS to not crash on Linux.

PR #1308, opened 2023-07-25
Implement unlink and symlink

swankjesse:


okio-wasifilesystem/src/wasmMain/kotlin/okio/internal/preview1/Preview1.kt
```diff
@@ -130,6 +130,51 @@ internal external fun path_open(
...
+
+/**
+ * path_symlink(old_path: string, fd: fd, new_path: string) -> Result<(), errno>
+ *
+ * Create a symbolic link.
+ * Note: This is similar to `symlinkat` in POSIX.
+ */
+@WasmImport("wasi_snapshot_preview1", "path_symlink")
+internal external fun path_symlink(
+  old_path: PointerU8,
```
swankjesse, line 157: The names old_path and new_path look like a copy-paste mistake in the spec. These words confuse me when thinking of symlinks.

okio-wasifilesystem/src/wasmMain/kotlin/okio/internal/preview1/Preview1.kt
```diff
@@ -130,6 +130,51 @@ internal external fun path_open(
...
+ */
+@WasmImport("wasi_snapshot_preview1", "path_rename")
+internal external fun path_rename(
+  fd: fd,
+  old_path: PointerU8,
+  old_pathSize: size,
+  new_fd: fd,
+  new_path: PointerU8,
+  new_pathSize: size,
+): Int // should be Short??
```
oldergod, line 147: Have a strategy to know about this one?

oldergod: Or you mean it gives you an Int but you think it should actually be a short?

oldergod (approved):

okio-wasifilesystem/src/wasmMain/kotlin/okio/internal/preview1/Preview1.kt
swankjesse: I should do an experiment to see if they both work. This maps to errno, which is a 16-bit value, but there’s potentially also minimum sizes for these things? It’s all pretty new.

PR #1309, opened 2023-07-26
Implement FileMetadata for WasiFileSystem

swankjesse:
This doesn't yet handle no such file.


okio-wasifilesystem/src/wasmMain/kotlin/okio/WasiFileSystem.kt
```diff
@@ -50,7 +57,60 @@ object WasiFileSystem : FileSystem() {
   }
 
   override fun metadataOrNull(path: Path): FileMetadata? {
```
oldergod, line 59: kudos

PR #1312, opened 2023-07-27
Set up spotless header configuration

oldergod:
The feature I added to spotless makes it a bit easier to exclude files we don't own.
I had to move some comments/documentations below the first JVM instruction to avoid getting it removed.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -169,6 +169,11 @@ subprojects {
     kotlin {
       target("**/*.kt")
       ktlint(libs.versions.ktlint.get())
+      licenseHeaderFile(rootProject.file("gradle/license-header.txt"))
+      targetExcludeIfContentContainsRegex("" +
+        "// Copyright 2019-2023 the Contributors to the WASI Specification|" +
+        "// ktlint-disable filename|" +
+        "Licensed to the Apache Software Foundation \\(ASF\\) under one or more")
     }
   }
 
--- a/gradle/license-header.txt
+++ b/gradle/license-header.txt
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      https://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/gradle/license-header.txt
+++ b/gradle/license-header.txt
@@ -1,0 +1,15 @@
+/*
+ * Copyright (C) $YEAR Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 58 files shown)


build.gradle.kts
```diff
@@ -169,6 +169,11 @@ subprojects {
     kotlin {
       target("**/*.kt")
       ktlint(libs.versions.ktlint.get())
+      licenseHeaderFile(rootProject.file("gradle/license-header.txt"))
+      targetExcludeIfContentContainsRegex("" +
+        "// Copyright 2019-2023 the Contributors to the WASI Specification|" +
+        "// ktlint-disable filename|" +
```
oldergod, line 175: Otherwise, spotless would override the whole header get rid of the ktlint comment and fail on the next iteration.

Egorand: Are there many files that contain this header? You can exclude individual files, see how we do it e.g. in `filament/views/build.gradle`.

gradle/license-header.txt
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      https://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
Egorand, line 8: Why not HTTPS?

Egorand (approved):

okio/src/commonMain/kotlin/okio/internal/RealBufferedSink.kt
```diff
@@ -5,20 +5,20 @@
...
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-// TODO move to RealBufferedSink class: https://youtrack.jetbrains.com/issue/KT-20427
 @file:Suppress("NOTHING_TO_INLINE")
 
 @file:JvmName("-RealBufferedSink") // A leading '-' hides this class from Java.
 
+// TODO move to RealBufferedSink class: https://youtrack.jetbrains.com/issue/KT-20427
```
swankjesse, line 20: I believe this comment was intended to explain the suppression on line 16

okio/src/commonMain/kotlin/okio/internal/SegmentedByteString.kt
```diff
@@ -5,20 +5,20 @@
...
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-// TODO move to SegmentedByteString class: https://youtrack.jetbrains.com/issue/KT-20427
 @file:Suppress("NOTHING_TO_INLINE")
 
 @file:JvmName("-SegmentedByteString") // A leading '-' hides this class from Java.
 
+// TODO move to SegmentedByteString class: https://youtrack.jetbrains.com/issue/KT-20427
```
swankjesse, line 20: same

gradle/license-header.txt
oldergod: Because otherwise it's 200+ file changes and I wanna decrease the noise as much as possible. I'll follow up

oldergod:
That's a simpler solution indeed

JakeWharton:
Misk is not a great example to follow:
- The pom license name should be a SPDX identifier, not an arbitrary string
- Our open source policy recommends a license at the header of each file. One section implies it's required, the other says "best practice".
- A `LICENSE` or `LICENSE.txt` is required, and should have prevented the project from passing open source review

PR #1314, opened 2023-07-27
Attempt to fix publishing with WASM

swankjesse:
The new okio-wasifilesystem module is a multiplatform module with a single artifact. This configuration is broken on the latest version of Dokka.
    https://github.com/Kotlin/dokka/issues/3038


oldergod (commented):
LGTM!

PR #1321, opened 2023-08-01
Only create relative symlinks in WasiFileSystem

swankjesse:
Creating absolute symlinks is broken in WASI. This also works better for preopens.

### DIFF
--- a/okio-wasifilesystem/src/wasmMain/kotlin/okio/WasiFileSystem.kt
+++ b/okio-wasifilesystem/src/wasmMain/kotlin/okio/WasiFileSystem.kt
@@ -85,7 +85,7 @@ object WasiFileSystem : FileSystem() {
         if (dirNameErrno != 0) throw ErrnoException(dirNameErrno.toShort())
         val dirName = bufPointer.readString(size)
         val dirNamePath = dirName.toPath()
-        add(Preopen(dirNamePath.segmentsBytes, fd))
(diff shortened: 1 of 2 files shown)


okio-wasifilesystem/src/wasmMain/kotlin/okio/WasiFileSystem.kt
```diff
@@ -427,13 +428,25 @@ object WasiFileSystem : FileSystem() {
...
+        ?: throw FileNotFoundException("no preopen: $source")
+
+      // Always create symlinks relative to their source. Absolute symlinks are trouble because the
+      // absolute paths used by WASI are different from the absolute paths on the host file system.
+      val sourceParent = source.parent
+        ?: throw IOException("unexpected symlink source: $source")
+      val targetRelative = when {
+        target.isRelative -> target
+        else -> target.relativeTo(sourceParent)
+      }
```
oldergod, line 441: Is this even acknowledged as a bug by the WASI team?

oldergod (approved):

okio-wasifilesystem/src/wasmMain/kotlin/okio/WasiFileSystem.kt
swankjesse: Yeah, they’ve changed it in the forthcoming preview2.

PR #1334, opened 2023-08-16
Fix a bug where xlen larger than 0x7fff was rejected (1.x)

xstefank:
Backported from 81bce1a30af244550b0324597720e4799281da7b

Backported for 1.x. Please consider releasing this fix for 1.x stream so we don't need to update to 3.4.0 since it's a big jump.


jhoestevam:
Great! This will save a lot of time for us.

xstefank:
@swankjesse, @oldergod can you please consider releasing this fix for 1.x?

swankjesse (approved):

swankjesse:
1.17.6 is out. https://repo1.maven.org/maven2/com/squareup/okio/okio/1.17.6/

PR #1344, opened 2023-09-10
Handle JDK Deflater throwing NPE

yschimke:
From https://github.com/square/okhttp/issues/6719

Not sure it isn't a programming bug, but even if so, IOException seems appropriate.

### DIFF
--- a/okio/src/jvmTest/kotlin/okio/DeflaterSinkTest.kt
+++ b/okio/src/jvmTest/kotlin/okio/DeflaterSinkTest.kt
@@ -15,7 +15,6 @@
  */
 package okio
 
-import java.io.IOException
 import java.util.zip.Deflater
 import java.util.zip.Inflater
 import java.util.zip.InflaterInputStream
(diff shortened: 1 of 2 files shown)


okio/src/jvmTest/kotlin/okio/DeflaterSinkTest.kt
```diff
@@ -129,6 +128,29 @@ class DeflaterSinkTest {
     mockSink.assertLogContains("close()")
   }
 
+  /**
+   * This test deflates a single segment of without compression because that's
+   * the easiest way to force close() to emit a large amount of data to the
+   * underlying sink.
```
oldergod, line 134:
```suggestion
   * This test deflates a single segment without compression because that's
   * the easiest way to force close() and emit a large amount of data to the
   * underlying sink.
```

oldergod (approved):

okio/src/jvmTest/kotlin/okio/DeflaterSinkTest.kt
yschimke: I think the comment is garbage, copy and pasted.  I'll update to a more succinct and correct one.

PR #1359, opened 2023-09-30
Fix file leak when using `metadata` functions

### LINKED ISSUE #1356: ZipFileSystem.metadata leaves file open
**Version: 3.5.0**

Calling `metadata` on a `ZipFileSystem` leaves the ZIP file open in its parent `FileSystem`, at least as reported by `FakeFileSystem`.

Test:

```Kotlin
package com.sample

import okio.BufferedSink
import okio.Path.Companion.toPath
import okio.buffer
import okio.fakefilesystem.FakeFileSystem
import okio.openZip
import okio.sink
import java.util.zip.ZipEntry
import java.util.zip.ZipOutputStream
import kotlin.test.Test
import kotlin.test.assertTrue

class ZipOpenTest {

    @Test
    fun test() {
        val baseFileSystem = FakeFileSystem().apply {
            write("/my.zip".toPath()) {
                writeZip {
                    putEntry("some.file") {
                        writeUtf8("contents")
                    }
                }
            }
        }
        val zipFileSystem = baseFileSystem.openZip("/my.zip".toPath())
        // Call to `metadata` leaves file open
        assertTrue(zipFileSystem.metadata("/some.file".toPath()).isRegularFile)
        baseFileSystem.checkNoOpenFiles()
    }
}

// Some helpers

/**
 * Writes a ZIP file to a [BufferedSink].
 */
private inline fun <R> BufferedSink.writeZip(action: ZipOutputStream.() -> R): R {
    return ZipOutputStream(outputStream()).use(action)
}

/**
 * Adds a new ZIP entry named [name], populates it with [action], and closes the entry.
 */
private inline fun <R> ZipOutputStream.putEntry(name: String, action: BufferedSink.() -> R): R {
    putNextEntry(ZipEntry(name).apply { time = 0L })
    val sink = sink().buffer()
    return try {
        sink.action()
    } finally {
        sink.flush()
        closeEntry()
    }
}
```

This test fails:

```
expected 0 open files, but found:
    /my.zip
java.lang.IllegalStateException: expected 0 open files, but found:
    /my.zip
	at okio.fakefilesystem.FakeFileSystem.checkNoOpenFiles(FakeFileSystem.kt:166)
	at com.sample.ZipOpenTest.test(ZipOpenTest.kt:30)
	...
Caused by: java.lang.Exception: file opened for READ here
	at okio.fakefilesystem.FakeFileSystem.open(FakeFileSystem.kt:368)
	at okio.fakefilesystem.FakeFileSystem.open$default(FakeFileSystem.kt:302)
	at okio.fakefilesystem.FakeFileSystem.openReadOnly(FakeFileSystem.kt:295)
	at okio.ZipFileSystem.metadataOrNull(ZipFileSystem.kt:102)
	at okio.internal.-FileSystem.commonMetadata(FileSystem.kt:36)
	at okio.FileSystem.metadata(FileSystem.kt:33)
	at com.sample.ZipOpenTest.test(ZipOpenTest.kt:29)
	... 43 more
```

The same does not happen if, for example, `/some.file` is read using `zipFileSystem.read(...)`.

pablobaxter:
Discovered this issue via the [wire](https://github.com/square/wire) project.

The issue here is that the `ZipFileSystem` creates a `Source` object, which is then returned out of the `use` lambda, which the lambda is supposed to close the `FileHandle` that is created, however, because the `Source` object had not been closed, the `FileHandle` remains open.
```kotlin
val source = fileSystem.openReadOnly(zipPath).use { fileHandle ->
  fileHandle.source(entry.offset).buffer()
}
return source.readLocalHeader(basicMetadata)
```

Going into a bit more detail, the `openStreamCount` is not `0` because the `Source` object wasn't closed, so we never close the underlying `RandomAccessFile` that was opened.
```kotlin
@Throws(IOException::class)
final override fun close() {
  lock.withLock {
    if (closed) return
    closed = true
    if (openStreamCount != 0) return
  }
  protectedClose()
}
```

This fix ensures the `Source` object is closed once the metadata has been extracted. This issue also occurs when calling `exists()` in the `ZipFileSystem`.

Closes #1356

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/ZipFileSystem.kt
+++ b/okio/src/jvmMain/kotlin/okio/ZipFileSystem.kt
@@ -99,10 +99,11 @@ internal class ZipFileSystem internal constructor(
       return basicMetadata
     }
 
-    val source = fileSystem.openReadOnly(zipPath).use { fileHandle ->
-      fileHandle.source(entry.offset).buffer()
+    return fileSystem.openReadOnly(zipPath).use { fileHandle ->
+      return@use fileHandle.source(entry.offset).buffer().use { source ->
+        source.readLocalHeader(basicMetadata)
+      }
     }
-    return source.readLocalHeader(basicMetadata)
   }
 
   override fun openReadOnly(file: Path): FileHandle {
--- a/okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
+++ b/okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
@@ -1,0 +1,97 @@
+/*
+ * Copyright (C) 2023 Square, Inc. and others.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+import java.util.zip.ZipEntry
+import java.util.zip.ZipOutputStream
+import kotlin.test.assertEquals
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+import okio.Path.Companion.toPath
+import okio.fakefilesystem.FakeFileSystem
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+
+class FileLeakTest {
+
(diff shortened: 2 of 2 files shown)


JakeWharton:
#1356 has a test. Can we copy that in so as to not regress?

okio/src/jvmMain/kotlin/okio/ZipFileSystem.kt
```diff
@@ -99,10 +99,11 @@ internal class ZipFileSystem internal constructor(
       return basicMetadata
     }
 
-    val source = fileSystem.openReadOnly(zipPath).use { fileHandle ->
-      fileHandle.source(entry.offset).buffer()
+    return fileSystem.openReadOnly(zipPath).use { fileHandle ->
```
swankjesse, line 102: Much better

swankjesse (approved):
Thanks for finding this and also for fixing this. Your test is brilliant too. Yay!

PR #1387, opened 2023-12-12
Update Kotlin to 1.9.21

swankjesse:


okio/src/commonMain/kotlin/okio/CommonPlatform.kt
```diff
@@ -33,14 +33,16 @@ expect inline fun <T> Lock.withLock(action: () -> T): T
 internal expect fun newLock(): Lock
 
 expect open class IOException(message: String?, cause: Throwable?) : Exception {
-  constructor(message: String? = null)
+  constructor(message: String?)
```
JakeWharton, line 36: I'm betting the lack of a zero arg is going to break some people. Even if we don't use it (aside from in the test) seems prudent to add it here and below to avoid that.

JakeWharton (approved):

okio/src/commonMain/kotlin/okio/CommonPlatform.kt
swankjesse: Done. Easy fix.

PR #1391, opened 2023-12-14
Implement Timeout.cancel for waitUntilNotified

swankjesse:
This triggers a 'spurious wakeup' in the case where a timeout is canceled. This is the best we can do without a discrete mechanism to distinguish between a notify() and a timeout.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Timeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/Timeout.kt
@@ -19,6 +19,7 @@ import java.io.IOException
 import java.io.InterruptedIOException
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.locks.Condition
+import kotlin.concurrent.Volatile
 import kotlin.time.Duration
 import kotlin.time.DurationUnit
 import kotlin.time.toTimeUnit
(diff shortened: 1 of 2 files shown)


swankjesse:
I intend to implement Timeout.cancel() for awaitSignal(), intersectWith(), ForwardingTimeout, and AsyncTimeout in follow-up changes.

okio/src/jvmMain/kotlin/okio/Timeout.kt
```diff
@@ -32,6 +33,12 @@ actual open class Timeout {
   private var deadlineNanoTime = 0L
   private var timeoutNanos = 0L
 
+  /**
+   * A sentinel that is updated to a new object on each call to [cancel]. Sample this property
+   * before and after an operation to test if the timeout was canceled during the operation.
+   */
+  @Volatile private var cancelMark: Any? = null
```
swankjesse, line 40: I also considered doing `cancelCount: Int` but I think I like this better. It doesn’t have an overflow problem!

yschimke (approved):

okio/src/jvmMain/kotlin/okio/Timeout.kt
JakeWharton: I think it would have worked. Concurrent read, add, writes would still have changed the value. Overflowing still results in a new value. The only problem would have been if you called cancel 2^32 times during an operation it wouldn't be detected (but that's probably okay because it's impossible).

oldergod (approved):
ΛΓΤΜ

okio/src/jvmMain/kotlin/okio/Timeout.kt
swankjesse: That’s exactly the case I was considering. It’s practically impossible but ugly

PR #1394, opened 2023-12-15
Don't throw in awaitSignal if the timeout is canceled

swankjesse:


okio/src/jvmMain/kotlin/okio/Timeout.kt
```diff
@@ -193,10 +193,19 @@ actual open class Timeout {
...
 
-      if (nanosRemaining <= 0) throw InterruptedIOException("timeout")
+      // If there's time remaining, we probably got the call we were waiting for.
+      if (nanosRemaining > 0) return
+
+      // Return without throwing if this timeout was canceled while we were waiting. Note that this
+      // return is a 'spurious wakeup' because Condition.signal() was not called.
+      if (cancelMark !== cancelMarkBefore) return
+
+      throw InterruptedIOException("timeout")
```
JakeWharton, line 208: Not my favorite control flow structure, but probably is the most readable.

JakeWharton (approved):

okio/src/jvmMain/kotlin/okio/Timeout.kt
swankjesse: Note that this throw doesn’t get caught by the directly-enclosing try/catch. InterruptedIOException and InterruptedException are tricky.

JakeWharton: Yeah I was going to suggest moving the throw out of the try/catch, but then I really want the returns out of the try/catch too, but then local scoping becomes hard and it's all terrible and whatever this is fine.

PR #1396, opened 2023-12-15
Implement AsyncTimeout.cancel()

swankjesse:
This is a simple implementation that uses 3 booleans to keep state (inQueue, isCanceled, hadTimeoutWhenCanceled). I'd like to do a follow-up change to replace 3 booleans with a proper state variable, especially since that eliminates some impossible states (like inQueue and isCanceled).

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -48,6 +48,9 @@ open class AsyncTimeout : Timeout() {
   /** If scheduled, this is the time that the watchdog should time this out.  */
   private var timeoutAt = 0L
 
+  private var isCanceled = false
+  private var hadTimeoutWhenCanceled = false
+
   fun enter() {
     val timeoutNanos = timeoutNanos()
     val hasDeadline = hasDeadline()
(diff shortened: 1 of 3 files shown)


okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -270,24 +294,22 @@ open class AsyncTimeout : Timeout() {
 
     /** Returns true if the timeout occurred. */
     private fun cancelScheduledTimeout(node: AsyncTimeout): Boolean {
```
swankjesse, line 296: GitHub’s UI doesn’t make it obvious, but the only change in this function is moving withLock() to the callsites.

okio/src/jvmTest/kotlin/okio/AsyncTimeoutTest.kt
```diff
@@ -372,9 +372,124 @@ class AsyncTimeoutTest {
     assertEquals(of(*data), target.readByteString())
   }
 
+  @Test
+  fun enterCancelSleepExit() {
```
swankjesse, line 376: I’m self-satisfied with these test names

JakeWharton (approved):

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
Egorand: Looks pretty good with "Hide whitespaces", although I might be biased cause I've already seen this code!

PR #1397, opened 2023-12-16
Use a state machine to implement AsyncTimeout

swankjesse:
Previously we had 3 boolean state variables to track 4 possible states. That was more complex than necessary.

### DIFF
--- a/okio/api/okio.api
+++ b/okio/api/okio.api
@@ -44,7 +44,6 @@ public final class okio/-InflaterSourceExtensions {
 }
 
 public class okio/AsyncTimeout : okio/Timeout {
-	public static final field Companion Lokio/AsyncTimeout$Companion;
 	public fun <init> ()V
 	public final fun access$newTimeoutException (Ljava/io/IOException;)Ljava/io/IOException;
 	public fun cancel ()V
@@ -55,11 +54,6 @@ public class okio/AsyncTimeout : okio/Timeout {
 	public final fun source (Lokio/Source;)Lokio/Source;
 	protected fun timedOut ()V
 	public final fun withTimeout (Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
-}
-
-public final class okio/AsyncTimeout$Companion {
-	public final fun getCondition ()Ljava/util/concurrent/locks/Condition;
-	public final fun getLock ()Ljava/util/concurrent/locks/ReentrantLock;
 }
 
 public final class okio/Buffer : java/lang/Cloneable, java/nio/channels/ByteChannel, okio/BufferedSink, okio/BufferedSource {
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -28,19 +28,18 @@ import kotlin.concurrent.withLock
  * writing.
  *
  * Subclasses should override [timedOut] to take action when a timeout occurs. This method will be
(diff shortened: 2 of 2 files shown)


okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -250,77 +288,67 @@ open class AsyncTimeout : Timeout() {
...
-      AsyncTimeout.lock.withLock {
-        check(!node.inQueue) { "Unbalanced enter/exit" }
-        node.inQueue = true
-
-        // Start the watchdog thread and create the head node when the first timeout is scheduled.
-        if (head == null) {
-          head = AsyncTimeout()
-          Watchdog().start()
-        }
+    private fun insertIntoQueue(node: AsyncTimeout, timeoutNanos: Long, hasDeadline: Boolean) {
```
swankjesse, line 291: Most of the changes in this function are just whitespace. I moved `withLock {}` to the caller.

Screenshot 2023-12-16 at 10 03 42 AM

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -250,77 +288,67 @@ open class AsyncTimeout : Timeout() {
...
           prev.next = node.next
           node.next = null
-          return false
+          return
         }
         prev = prev.next
       }
 
-      // The node wasn't found in the linked list: it must have timed out!
-      return true
```
swankjesse, line 312: Now that we keep a state machine we don’t need to traverse the linked list to learn this

okio/api/okio.api
```diff
@@ -57,11 +56,6 @@ public class okio/AsyncTimeout : okio/Timeout {
 	public final fun withTimeout (Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
 }
 
-public final class okio/AsyncTimeout$Companion {
```
swankjesse, line 60: Ugh, I think these snuck in by accident. I don’t believe they belong in the public API, and I don’t think anyone ever used them.

JakeWharton (approved):

Egorand (approved):

yyaroshevich:
Hello,

I'm not sure if this is the right place to ask, but I hope @swankjesse, as the author of several timeout-related PRs, might be able to help. Please know that I ask this question with the utmost respect and curiosity, and it is not meant to be critical or blaming in any way.

Is there a specific reason why `AsyncTimeout` is still implemented as a manual linked list with a backing thread instead of using `ScheduledThreadPoolExecutor` in 2024? I understand that this class was initially written before JDK 8, but now it seems that much of the complex code could be simplified by using `ScheduledThreadPoolExecutor`. 

I'm asking because I've observed that list manipulations, such as linear time insertions into the linked list, have become a bottleneck in a back-end service that heavily relies on `OkHttpClient` and probably it is bottleneck in other apps which heavily uses `OkHttpClient`.

Thank you for any insights you can provide.

Best regards, Yury

swankjesse:
@yyaroshevich that’s a good point!

Wanna open a tracking bug for us to consider switching? At the time of creation we optimized for fewest allocations, which were particularly slow on Android.

One other option that we should investigate – could we accomplish better performance by searching backwards instead of forwards in the list? If all of the timeouts are symmetric (they often are), then inserts will usually be at the end. We could reduce contention without increasing allocation?

yyaroshevich:
> Wanna open a tracking bug for us to consider switching?

Sure, I've opened ticket where we can proceed discussion https://github.com/square/okio/issues/1488. Thanks a lot!

PR #1405, opened 2024-01-06
Add EnumOptions as an exhaustive alternative to Options

kyay10:
This facilitates using `when` exhaustively on it.

### DIFF
--- a/okio/api/okio.api
+++ b/okio/api/okio.api
@@ -396,6 +396,26 @@ public final class okio/DeflaterSink : okio/Sink {
 	public fun write (Lokio/Buffer;J)V
 }
 
+public abstract interface class okio/EnumOption {
+	public abstract fun getByteString ()Lokio/ByteString;
+}
+
+public abstract class okio/EnumOptions {
+	public static final field Companion Lokio/EnumOptions$Companion;
+	public fun <init> (Ljava/util/List;)V
+	public final fun getEntries ()Ljava/util/List;
+	public final fun getOptions ()Lokio/Options;
+	public static final fun of (Ljava/util/List;)Lokio/EnumOptions;
+}
+
+public final class okio/EnumOptions$Companion {
+	public final fun of (Ljava/util/List;)Lokio/EnumOptions;
+}
+
+public final class okio/EnumOptionsKt {
+	public static final fun select (Lokio/BufferedSource;Lokio/EnumOptions;)Lokio/EnumOption;
+}
+
 public abstract interface annotation class okio/ExperimentalFileSystem : java/lang/annotation/Annotation {
 }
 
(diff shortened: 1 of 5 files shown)


swankjesse (commented):
Thanks for the idea on this. I think it’s a solid idea.

I’m uncertain of the implementation. In particular I’d like to explore similar `when` friendly syntax without as much runtime overhead. We’re really sensitive to that in Okio! 

I’m also a bit anxious about the net new number of tests here. I don’t want a copy of all the old Options' code’s tests just because the Options were constructed differently. It’s more code to maintain, and more tests to run.

Next steps: I’m going to consider alternative implementations for the implied feature request ‘manual int mapping for options is unpleasant’.

I know this isn’t the outcome you wanted for this PR. My recommendation going forward is to get a ‘yes please build this’ message from us before investing too much of your time on pull requests.

swankjesse:
Thanks for this.

kyay10:
No problem at all, I completely get it! It didn't take me that much time. To be honest, I saw one of your talks about okio, saw that Options were very rudimentary, and thought hey I could fix this in an afternoon.

Would you like me to open an issue for this, so that we can discuss possible implementations? 

I can imagine you're very busy so of course it's not a priority at all. I just want to try some implementations out to see which would fit okio the best (I don't have extensive experience with the library so of course your feedback is much much appreciated).

swankjesse:
Thanks for being accommodating. 😅

I spent my lunch time thinking this over. I really like your idea of a `select()` function that uses something friendlier than ints. Here’s the smallest API I can think of to satisfy that requirement:

```kotlin
class TypedOptions<T : Any>(
  values: Iterable<T>,
  encode: (T) -> ByteString,
) {
  internal val list = values.toList()
  internal val options = Options.of(*list.map { encode(it) }.toTypedArray<ByteString>())
}

fun <T : Any> BufferedSource.select(typedOptions: TypedOptions<T>): T? {
  return when (val index = select(typedOptions.options)) {
    -1 -> null
    else -> typedOptions.list[index]
  }
}
```

And here’s what it looks like in practice:

```kotlin
  @Test
  fun happyPath() {
    val durationOptions = TypedOptions(DurationUnit.entries) {
      it.name.encodeUtf8()
    }

    val buffer = Buffer().writeUtf8("SECONDS,HOURS,DAYS")
    assertThat(buffer.select(durationOptions)).isEqualTo(DurationUnit.SECONDS)
    buffer.readByte()
    assertThat(buffer.select(durationOptions)).isEqualTo(DurationUnit.HOURS)
    buffer.readByte()
    assertThat(buffer.select(durationOptions)).isEqualTo(DurationUnit.DAYS)
  }
```

swankjesse:
Next steps: if the above works for you, I’ll add a couple tests & send a pull request!

kyay10:
Yep that absolutely works! Only nitpick I have would maybe be making the constructor just take the list of bytestrings instead, and making that an internal constructor, then having a "smart constructor" which is just a function that actually does the `map`ping business, so that there's no unnecessary lambda objects being created or stored around

swankjesse:
Here it is! https://github.com/square/okio/pull/1417

PR #1413, opened 2024-01-29
Fix bug in Pipe.fold where sink can incorrectly be left open

### LINKED ISSUE #1412: Pipe.fold: destination sink can remain open even after pipe closed
If a call to `Pipe.sink.close()` is invoked concurrently while `Pipe.fold()` is running, it can be missed by the folding thread, leaving the fold destination open, even though it should be closed. Depending on how the sink is used, this can lead to deadlock or slowness and timeout failures.

jhump:
This adds a new test case that demonstrates the bug described in #1412.

This also proposes a fix, by making sure the final loop iteration -- when the buffer is empty and the new destination is swapped in -- can close the destination sink if necessary.

Resolves #1412

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Pipe.kt
+++ b/okio/src/jvmMain/kotlin/okio/Pipe.kt
@@ -172,6 +172,7 @@ class Pipe(internal val maxBufferSize: Long) {
       // Either the buffer is empty and we can swap and return. Or the buffer is non-empty and we
       // must copy it to sink without holding any locks, then try it all again.
       var closed = false
+      var done = false
       lateinit var sinkBuffer: Buffer
       lock.withLock {
         check(foldedSink == null) { "sink already folded" }
@@ -181,26 +182,30 @@ class Pipe(internal val maxBufferSize: Long) {
           throw IOException("canceled")
         }
 
+        closed = sinkClosed
         if (buffer.exhausted()) {
           sourceClosed = true
           foldedSink = sink
-          return@fold
+          done = true
+          return@withLock
         }
 
-        closed = sinkClosed
         sinkBuffer = Buffer()
         sinkBuffer.write(buffer, buffer.size)
         condition.signalAll() // Notify the sink that it can resume writing.
       }
 
+      if (done) {
+        if (closed) {
+          sink.close()
+        }
+        return
+      }
+
       var success = false
       try {
         sink.write(sinkBuffer, sinkBuffer.size)
-        if (closed) {
-          sink.close()
-        } else {
-          sink.flush()
-        }
+        sink.flush()
         success = true
       } finally {
         if (!success) {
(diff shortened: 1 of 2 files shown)


okio/src/jvmMain/kotlin/okio/Pipe.kt
```diff
@@ -181,26 +182,30 @@ class Pipe(internal val maxBufferSize: Long) {
...
       }
 
-      var success = false
-      try {
-        sink.write(sinkBuffer, sinkBuffer.size)
+      if (done) {
         if (closed) {
           sink.close()
-        } else {
-          sink.flush()
```
jhump, line 202: It wasn't clear if this flush vs. close was really necessary. It felt like an optimization. I changed it because I didn't want the code to inadvertently close the sink 2x, and also didn't want to add yet-another piece of state to prevent such a double-close. The code seemed a little easier on the eyes to simply have the close called in one place, and accept the potential overhead and explicit flush right before close.

JakeWharton: > inadvertently close the sink 2x

The [documentation of `Sink`](https://square.github.io/okio/3.x/okio/okio/okio/-sink/close.html) requires the second call to be a no-op. So if it simplifies things here, you are free to do it.

jhump: I don't know that it simplifies anything to let it call `close` here, too. It would be an extra conditional and a couple of extra lines that IMO aren't much value. Feels like we should only be bothering to call `close` it at the end.

I was just explaining that by moving the call to close to the end, we _could_ end up calling flush needlessly, because we'll call it in the last iteration of the loop for the last chunk of data, but we might then immediately call close on return. We don't know for certain here whether we're going to close, so we do need the flush. A possible alternative would be to defer calling flush until we return from the loop, but then we might induce unsynchronized concurrent access to the destination sink -- by calling flush in the folding thread, which is stated to be safe to call concurrently with calls to write to the pipe's sink, which would also write to the destination sink, but from a different thread.

Sorry if that's hard to parse. Maybe it would be more clear to just review the change and see if it looks fine or if you'd prefer an alternate formulation. (It's small; most of the lines of code in this PR are the new test case.)

JakeWharton: I'm on mobile which is why I didn't. I'll reread this and review the diff later or tomorrow on a nice big screen.

jhump: (gentle reminder, if you have a moment to look)

JakeWharton: I did take a look, and it only served to convince me that @swankjesse should be the one looking!

swankjesse (approved):
This looks great. Thank you!

PR #1424, opened 2024-02-06
Don't check buffer exhaustion while filling ints

JakeWharton:
The array size is already derived from the buffer size.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/Options.kt
+++ b/okio/src/commonMain/kotlin/okio/Options.kt
@@ -71,10 +71,8 @@ class Options private constructor(
       val trieBytes = Buffer()
       buildTrieRecursive(node = trieBytes, byteStrings = list, indexes = indexes)
 
-      val trie = IntArray(trieBytes.intCount.toInt())
-      var i = 0
-      while (!trieBytes.exhausted()) {
-        trie[i++] = trieBytes.readInt()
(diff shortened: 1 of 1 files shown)


okio/src/commonMain/kotlin/okio/Options.kt
```diff
@@ -71,10 +71,8 @@ class Options private constructor(
       val trieBytes = Buffer()
       buildTrieRecursive(node = trieBytes, byteStrings = list, indexes = indexes)
 
-      val trie = IntArray(trieBytes.intCount.toInt())
-      var i = 0
-      while (!trieBytes.exhausted()) {
-        trie[i++] = trieBytes.readInt()
+      val trie = IntArray(trieBytes.intCount.toInt()) {
+        trieBytes.readInt()
```
oldergod, line 75: Don't you still need to loop over it?

JakeWharton: This is a loop. It's called once for each position in the array.

oldergod: oooooh I misread that the previouew code wasn't doing that in the block passed to `IntArray`.

oldergod (approved):

swankjesse (approved):

Issue #1364, opened 2023-10-10
Potential file leak in `FileSystem.read()` method


pablobaxter:
Continued investigation in file descriptor leaks led to this stack trace:
```
#1 /Users/pablobaxter/.gradle/caches/modules-2/files-2.1/com.squareup.wire/wire-schema-jvm/4.7.0/4f9c99c8217fc7d53fd11fa3e885559c62280227/wire-schema-jvm-4.7.0.jar by thread:main on Tue Oct 10 12:47:57 PDT 2023
        at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:173)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:347)
        at java.base/sun.net.www.protocol.jar.URLJarFile.<init>(URLJarFile.java:103)
        at java.base/sun.net.www.protocol.jar.URLJarFile.getJarFile(URLJarFile.java:72)
        at java.base/sun.net.www.protocol.jar.JarFileFactory.getOrCreate(JarFileFactory.java:106)
        at java.base/sun.net.www.protocol.jar.JarURLConnection.connect(JarURLConnection.java:132)
        at java.base/sun.net.www.protocol.jar.JarURLConnection.getInputStream(JarURLConnection.java:175)
        at java.base/java.net.URL.openStream(URL.java:1161)
        at java.base/java.lang.ClassLoader.getResourceAsStream(ClassLoader.java:1735)
        at okio.internal.ResourceFileSystem.source(ResourceFileSystem.kt:129)
        at com.squareup.wire.schema.CoreLoader.load(CoreLoader.kt:54)
        at com.squareup.wire.schema.internal.CommonSchemaLoader.load(CommonSchemaLoader.kt:148)
        at com.squareup.wire.schema.Linker.getFileLinker$wire_schema(Linker.kt:74)
        at com.squareup.wire.schema.Linker.link(Linker.kt:95)
        at com.squareup.wire.schema.internal.CommonSchemaLoader.loadSchema(CommonSchemaLoader.kt:103)
        at com.squareup.wire.schema.SchemaLoader.loadSchema(SchemaLoader.kt:71)
        at com.squareup.wire.schema.WireRun.execute(WireRun.kt:219)
        at com.squareup.wire.schema.WireRun.execute(WireRun.kt:211)
```

[...]

PR #1428, opened 2024-02-10
Prevent caching URLConnection files in ResourceFileSystem

pablobaxter:
This fix is for #1364. It prevents the issue where `ClassLoader.getResourceAsStream()` caches an open file due to `URLConnection.openStream()` works. More details in https://github.com/square/okio/issues/1364#issuecomment-1758260314.

### DIFF
--- a/okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
+++ b/okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
@@ -149,7 +149,7 @@ private inline fun <R> ZipOutputStream.putEntry(name: String, action: BufferedSi
 // This is a Linux only test for open file descriptors on the current process
 @OptIn(ExperimentalPathApi::class)
 private fun Path.assetFileNotOpen() {
-  val fds = Path("/proc/${ProcessHandle.current().pid()}/fd")
+  val fds = Path("/proc/self/fd")
   if (fds.isDirectory()) {
     // Linux: verify that path is not open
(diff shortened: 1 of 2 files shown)


okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
```diff
@@ -109,3 +145,18 @@ private inline fun <R> ZipOutputStream.putEntry(name: String, action: BufferedSi
     closeEntry()
   }
 }
+
+// This is a Linux only test for open file descriptors on the current process
```
swankjesse, line 149: Neato

swankjesse (approved):
This is rad. Good fix.

okio/src/jvmTest/kotlin/okio/FileLeakTest.kt
```diff
@@ -109,3 +145,18 @@ private inline fun <R> ZipOutputStream.putEntry(name: String, action: BufferedSi
...
+@OptIn(ExperimentalPathApi::class)
+private fun Path.assetFileNotOpen() {
+  val fds = Path("/proc/self/fd")
+  if (fds.isDirectory()) {
+    // Linux: verify that path is not open
```
joshfriend, line 154: Since this only works on linux, some kind of `assumeTrue` to skip the test on other platforms would be useful to prevent future "WTF" moments where the tests pass on macOS because they don't assert anything then fail in CI

swankjesse (approved):

PR #1432, opened 2024-02-14
Implement CRC32

swankjesse:
On Kotlin/Native this uses the built-in zlib

### DIFF
--- a/okio/build.gradle.kts
+++ b/okio/build.gradle.kts
@@ -91,11 +91,21 @@ kotlin {
       dependsOn(commonTest)
     }
 
+    val zlibMain by creating {
+      dependsOn(commonMain)
+    }
+
+    val zlibTest by creating {
+      dependsOn(commonTest)
+    }
+
     val jvmMain by getting {
+      dependsOn(zlibMain)
     }
     val jvmTest by getting {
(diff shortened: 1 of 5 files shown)


okio/build.gradle.kts
```diff
@@ -91,11 +91,21 @@ kotlin {
       dependsOn(commonTest)
     }
 
+    val zlibMain by creating {
```
swankjesse, line 94: These are targets that offer zlib libraries; currently JVM and Kotlin/Native

okio/src/nativeMain/kotlin/okio/internal/CRC32.kt
```diff
@@ -0,0 +1,46 @@
...
+import kotlinx.cinterop.CValuesRef
+import kotlinx.cinterop.UnsafeNumber
+import kotlinx.cinterop.addressOf
+import kotlinx.cinterop.usePinned
+import platform.zlib.crc32
+import platform.zlib.uBytefVar
+
+@OptIn(UnsafeNumber::class)
+actual class CRC32 {
+  var crc = crc32(0u, null, 0u)
```
JakeWharton, line 27:
```suggestion
  private var crc = crc32(0u, null, 0u)
```

JakeWharton (approved):

PR #1434, opened 2024-02-15
Do expect/actual for shared code in Deflater

swankjesse:


okio/src/jvmMain/kotlin/okio/DeflaterSink.kt
```diff
@@ -13,8 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
-@file:JvmName("-DeflaterSinkExtensions")
```
swankjesse, line 17: extension moved to the multiplatform zlibMain source set

okio/src/nativeMain/kotlin/okio/Deflater.kt
```diff
@@ -66,47 +70,61 @@ internal class Deflater : DataProcessor() {
...
 
-    source.usePinned { pinnedSource ->
-      target.usePinned { pinnedTarget ->
-        val sourceByteCount = sourceLimit - sourcePos
-        zStream.next_in = when {
-          sourceByteCount > 0 -> pinnedSource.addressOf(sourcePos) as CPointer<UByteVar>
-          else -> null
-        }
-        zStream.avail_in = sourceByteCount.toUInt()
+  internal val dataProcessor: DataProcessor = object : DataProcessor() {
```
swankjesse, line 75: Super annoying; I can’t extend `DataProcessor` in Kotlin/Native because that doesn’t match the inheritance hierarchy in Kotlin/JVM.

okio/src/nativeMain/kotlin/okio/Inflater.kt
```diff
@@ -35,70 +36,86 @@ import platform.zlib.z_stream_s
...
+      nativeHeap.free(zStream)
+    }
   }
 
-  override fun close() {
-    if (closed) return
-    closed = true
+  actual constructor() : this(false)
+
+  @OptIn(UnsafeNumber::class)
```
swankjesse, line 112: These should be safe in practice! We don’t expose target-specific numbers in our public API

okio/src/zlibMain/kotlin/okio/DeflaterSink.kt
```diff
@@ -0,0 +1,57 @@
...
+ * Aggressive flushing of this stream may result in reduced compression. Each
+ * call to [flush] immediately compresses all currently-buffered data;
+ * this early compression may be less effective than compression performed
+ * without flushing.
+ *
+ * This is equivalent to using [Deflater] with the sync flush option.
+ * This class does not offer any partial flush mechanism. For best performance,
+ * only call [flush] when application behavior requires it.
+ */
+expect class DeflaterSink
```
JakeWharton, line 37: Any reason to make these public? The JVM ones have to remain public for compatibility, but these don't expose any API on top of the underlying `Sink`.

okio/src/nativeTest/kotlin/okio/DeflaterTest.kt
```diff
@@ -23,177 +23,237 @@ import kotlin.test.assertTrue
...
-  fun closeIsIdemptent() {
+  fun endIsIdemptent() {
     val deflater = Deflater()
-    deflater.close()
-    deflater.close()
+    deflater.end()
+    deflater.end()
   }
+
+  private fun Deflater(nowrap: Boolean) = Deflater(Z_BEST_COMPRESSION, nowrap)
```
JakeWharton, line 258: No way to put this default value onto the expect so you don't need to define this function?

JakeWharton (approved):

okio/src/zlibMain/kotlin/okio/DeflaterSink.kt
swankjesse: This API is handy to implement web sockets, SPDY/3, and PNG. We might use it in a future multiplatform Paparazzi!

JakeWharton: I'm referring specifically to the class. The extension below already provides access to the functionality (although its return type should probably be just `Sink`).

okio/src/nativeTest/kotlin/okio/DeflaterTest.kt
swankjesse: Not as far as I know.

```
> Task :okio:compileKotlinJvm FAILED
e: file:///Volumes/Development/okio/okio/src/jvmMain/kotlin/okio/-JvmPlatform.kt:45:1 Default argument values inside expect declaration 'Deflater' are not allowed if it is actualized via typealias. Possible fix is to remove default argument values in members:
    public constructor Deflater(level: Int = ..., nowrap: Boolean = ...)
```

okio/src/zlibMain/kotlin/okio/DeflaterSink.kt
swankjesse: I’d like for it to remain public so we keep the option of having APIs like finishDeflate() public. Or APIs to supply a Dictionary mid-compression if that exotic feature is used.

PR #1437, opened 2024-02-20
Use constant files in ZipFileSystemTest

swankjesse:
This should make it possible to run this test on platforms where programatically creating .zip files is difficult.


swankjesse:
Probably the biggest victory here is that these tests now run on Windows (JVM).

JakeWharton (approved):

oldergod (commented):
I love it

PR #1438, opened 2024-02-21
Handle DOS dates in Kotlin/Multiplatform

swankjesse:
This might be foolish. The total amount of new code is
relatively small (45 lines of code), but that code does
something we don't otherwise want to be doing in Okio -
date math.

But this unblocks implementing ZipFileSystem in Kotlin/Native.

### DIFF
--- a/okio/src/commonTest/kotlin/okio/OkioTesting.kt
+++ b/okio/src/commonTest/kotlin/okio/OkioTesting.kt
@@ -102,3 +102,5 @@ expect fun assertRelativeToFails(
   b: Path,
   sameAsNio: Boolean = true,
 ): IllegalArgumentException
+
+expect fun <T> withUtc(block: () -> T): T
--- a/okio/src/nativeMain/kotlin/okio/internal/-ZlibNative.kt
+++ b/okio/src/nativeMain/kotlin/okio/internal/-ZlibNative.kt
@@ -17,3 +17,62 @@
 package okio.internal
 
 internal actual val DEFAULT_COMPRESSION: Int = platform.zlib.Z_DEFAULT_COMPRESSION
+
+/**
+ * Roll our own date math because Kotlin doesn't include a built-in date math API, and the
+ * kotlinx.datetime library doesn't offer a stable at this time.
+ *
+ * Also, we don't necessarily want to take on that dependency for Okio.
+ *
+ * This implementation assumes UTC.
+ *
+ * This code is broken for years before 1970. It doesn't implement subtraction for leap years.
+ *
+ * This code is broken for out-of-range values. For example, it doesn't correctly implement leap
+ * year offsets when the month is -24 or when the day is -365.
+ */
+internal actual fun datePartsToEpochMillis(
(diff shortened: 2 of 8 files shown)


okio/src/nativeMain/kotlin/okio/internal/-ZlibNative.kt
```diff
@@ -17,3 +17,62 @@
 package okio.internal
 
 internal actual val DEFAULT_COMPRESSION: Int = platform.zlib.Z_DEFAULT_COMPRESSION
+
+/**
+ * Roll our own date math because Kotlin doesn't include a built-in date math API, and the
+ * kotlinx.datetime library doesn't offer a stable at this time.
```
swankjesse, line 23:
```suggestion
 * kotlinx.datetime library doesn't offer a stable release at this time.
```

okio/src/nativeMain/kotlin/okio/internal/-ZlibNative.kt
```diff
@@ -17,3 +17,62 @@
...
+ * Also, we don't necessarily want to take on that dependency for Okio.
+ *
+ * This implementation assumes UTC.
+ *
+ * This code is broken for years before 1970. It doesn't implement subtraction for leap years.
+ *
+ * This code is broken for out-of-range values. For example, it doesn't correctly implement leap
+ * year offsets when the month is -24 or when the day is -365.
+ */
+internal actual fun datePartsToEpochMillis(
```
JakeWharton, line 34: Why not use this in common? Do we gain anything from using the Java APIs?

okio/src/commonTest/kotlin/okio/OkioTesting.kt
```diff
@@ -102,3 +102,5 @@ expect fun assertRelativeToFails(
   b: Path,
   sameAsNio: Boolean = true,
 ): IllegalArgumentException
+
+expect fun <T> withUtc(block: () -> T): T
```
JakeWharton, line 106: ...it would eliminate this otherwise weird function that is only for JVM too.

JakeWharton (approved):

okio/src/nativeMain/kotlin/okio/internal/-ZlibNative.kt
swankjesse: Quite tragically I haven’t yet done the work to use the ‘local time zone’ on Kotlin/Native. So for everything but UTC the Kotlin/Native DOS dates will be off by the time zone offset.

I’ll spend a small effort attempting to fix this? The “extended timestamps” extension (0x5455) means we’ll probably get the correct answer anyway.

swankjesse: One other option that might make the code simpler: use Java APIs for the timezone adjustment only. That way I can ship one common implementation that uses timezone information where it’s available.

swankjesse:
I’m gonna do the timezone fixups in a follow-up!

PR #1439, opened 2024-02-22
Get openZip working on KotlinNative

swankjesse:


okio-testing-support/src/commonMain/resources/okio/zipfilesystem/empty.zip
JakeWharton, line 1: Minimal zip is 22 bytes. Look at this chunker at a whopping 41!!!

okio/build.gradle.kts
```diff
@@ -97,6 +97,9 @@ kotlin {
 
     val zlibTest by creating {
       dependsOn(commonTest)
+      dependencies {
+        implementation(libs.test.assertk)
```
JakeWharton, line 101: I will fix this wasm problem tomorrow (although I don't control when releases happen).

JakeWharton (approved):

okio/src/zlibMain/kotlin/okio/internal/FixedLengthSource.kt
JakeWharton, line 1: We should really make this public (with a little design tweak for OkHttp). We're long-past the rule of three on it.

okio/build.gradle.kts
swankjesse: I was playing with it. I ran into this:

```
* What went wrong:
Execution failed for task ':assertk:wasmWasiNodeTest'.
> command '/Users/jwilson/.gradle/nodejs/node-v21.0.0-v8-canary202309143a48826a08-darwin-arm64/bin/node' exited with errors (exit code: 1)
```

Totally unclear what I did wrong. Trying to borrow config from Okio wasn’t helping much!

okio/src/zlibMain/kotlin/okio/internal/FixedLengthSource.kt
swankjesse: Can do. Follow-up!

PR #1445, opened 2024-02-24
Tag the language in code samples

swankjesse:


okio-nodefilesystem/src/main/kotlin/okio/FsJs.kt
```diff
@@ -62,7 +62,7 @@
  *
  * This will create a file with a full set of APIs to copy-paste from.
  *
- * ```
+ * ```kotlin
```
JakeWharton, line 65: This one isn't Kotlin

JakeWharton (approved):

okio-nodefilesystem/src/main/kotlin/okio/FsJs.kt
oldergod, line 65:
```suggestion
 * ```
```

oldergod (approved):

PR #1449, opened 2024-03-03
Get ZipFileSystem to prefer NTFS timestamps when present

swankjesse:
They don't have a time bomb in 2038.


okio/src/zlibTest/kotlin/okio/ZipFileSystemGoTest.kt
```diff
@@ -0,0 +1,49 @@
...
+  private val fileSystem = SYSTEM_FILE_SYSTEM
+  private var base = okioRoot / "okio-testing-support" /
+    "src/commonMain/resources/go/src/archive/zip/testdata"
+
+  @Test
+  fun timeWinzip() {
+    val zipFileSystem = fileSystem.openZip(base / "time-winzip.zip")
+    val fileMetadata = zipFileSystem.metadata("test.txt".toPath())
+    assertEquals(
+      Instant.parse("2017-11-01T04:11:57.244Z"),
```
JakeWharton, line 37: Shouldn't this have an import?

JakeWharton (approved):

okio/src/zlibTest/kotlin/okio/ZipFileSystemGoTest.kt
swankjesse: Yes, it absolutely should. Why the heck does it compile?

PR #1453, opened 2024-03-05
Fix `Closeable.use` `NullPointerException`

### LINKED ISSUE #1450: `Closeable.use` throws NPE when `block` returns nullable
`block` return type `R` is defined as `<R>`, not `<R: Any>`. If `block` produces `null` (e.g. `Int?`) and no exception is thrown, return will cause a `KotlinNullPointerException`.

A cast as `R` instead of a double banger is needed on result

image

05nelsonm:
Closes #1450

### DIFF
--- a/okio/src/nonWasmTest/kotlin/okio/UseTest.kt
+++ b/okio/src/nonWasmTest/kotlin/okio/UseTest.kt
@@ -25,4 +25,11 @@ class UseTest {
 
     fakeFileSystem.checkNoOpenFiles()
   }
+
+  @Test
+  fun acceptsNullReturn() {
+    object : Closeable {
+      override fun close() {}
+    }.use { null }
+  }
 }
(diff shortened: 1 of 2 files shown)


okio/src/nonWasmTest/kotlin/okio/UseTest.kt
```diff
@@ -25,4 +25,11 @@ class UseTest {
 
     fakeFileSystem.checkNoOpenFiles()
   }
+
+  @Test
+  fun acceptsNullReturn() {
+    object : Closeable {
+      override fun close() {}
+    }.use { null }
```
JakeWharton, line 33: Can you add an assertNull or assertEquals here just for posterity.

05nelsonm: Resolved in commit f8dfad3

PR #1455, opened 2024-03-11
Create FileSystem.SYSTEM property in shared source set

zsmb13:
This PR creates a new source set called `systemFileSystem` that `jvm` and `native` depend on, and makes `FileSystem.SYSTEM` available there.

We can't remove the existing `SYSTEM` declarations in the companion objects of `FileSystem`, and we can't introduce a companion object in the shared source set as it could not be further refined in more specific source sets.

Therefore, this new declaration is added as an `expect` extension on the companion. This lets code targeting JVM and native platforms resolve to the extension in common code, while they will actually use the properties of the companions directly when compiled.

To make this change easier to test, we add additional platforms to the `samples` project, and a sample function that exercises the new property.


JakeWharton (approved):
Clever! LGTM once the sample build issues are fixed.

swankjesse (approved):
Long overdue. Nice fix.

swankjesse (approved):

PR #1419, opened 2024-02-05
Speed up CI

Goooler:
Android and macOS builds are much faster for now.

https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source

PR #1461, opened 2024-03-28
Correct documentation links in ZipFileSystem.kt

vanniktech:
Currently, it just seems broken:

Screenshot 2024-03-28 at 10 41 16

But I think this should fix it.

### DIFF
--- a/okio/src/zlibMain/kotlin/okio/ZipFileSystem.kt
+++ b/okio/src/zlibMain/kotlin/okio/ZipFileSystem.kt
@@ -24,7 +24,7 @@ import okio.internal.readLocalHeader
 import okio.internal.skipLocalHeader
 
 /**
- * Read only access to a [zip file][zip_format] and common [extra fields][extra_fields].
+ * Read only access to a [zip file](zip_format) and common [extra fields](extra_fields).
  *
  * [zip_format]: https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE_6.2.0.txt
  * [extra_fields]: https://opensource.apple.com/source/zip/zip-6/unzip/unzip/proginfo/extra.fld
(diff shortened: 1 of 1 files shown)


okio/src/zlibMain/kotlin/okio/ZipFileSystem.kt
```diff
@@ -24,7 +24,7 @@ import okio.internal.readLocalHeader
 import okio.internal.skipLocalHeader
 
 /**
- * Read only access to a [zip file][zip_format] and common [extra fields][extra_fields].
+ * Read only access to a [zip file](zip_format) and common [extra fields](extra_fields).
  *
  * [zip_format]: https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE_6.2.0.txt
  * [extra_fields]: https://opensource.apple.com/source/zip/zip-6/unzip/unzip/proginfo/extra.fld
```
oldergod, line 30: Doesn't look like it works for me. The links are there but nothing happens as I click on them.

image

I have to inline them to see something working.

image


```suggestion
 * Read only access to a [zip file](https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE_6.2.0.txt)
 * and common [extra fields](https://opensource.apple.com/source/zip/zip-6/unzip/unzip/proginfo/extra.fld).
```

oldergod:
What's this? Why is there no github details bound to it?

msfjarvis:
> What's this? Why is there no github details bound to it?
> 
> image

https://github.com/square/okio/pull/1419#issuecomment-2005725815

GitHub still expects the macos-11 job as a required check but it no longer exists.

JakeWharton:
I had fixed it, which is why the auto-merge triggered.

Also let's try to use squash for multi-commit PRs in the future unless each commit is a separate, tested unit.

PR #1469, opened 2024-04-15
fix for relativeTo with paths containing dots

### LINKED ISSUE #1468: Path.relativeTo does not respect 'current directory dot'
java.nio
```
Path.of("foo").relativeTo(Path.of(".")) 
// returns Path("foo")
```

java.io.
```
File("foo").relativeTo(File("."))
// returns File("foo")
```

okio
```
"foo".toPath().relativeTo(".".toPath())
// ⚠️ returns Path("../foo") 
```

oldergod:
Fixes #1468

Fix A (this PR)

image

Fix B (if you prefer)

image


okio/src/commonMain/kotlin/okio/internal/Path.kt
```diff
@@ -250,6 +250,11 @@ internal inline fun Path.commonRelativeTo(other: Path): Path {
     "Impossible relative path to resolve: $this and $other"
   }
 
+  if (other.segmentsBytes.size == 1 && other.segmentsBytes.first() == DOT) {
```
Egorand, line 253: nit:

```suggestion
  if (other.singleOrNull() == DOT) {
```

Egorand (approved):
Early return feels better to me!

oldergod:
LMAO (?) but Jvm8 so `java.io` doesn't do the "right" thing and fails with `java.lang.AssertionError: expected:<Desktop/documents/a> but was:<../Desktop/documents/a>`

swankjesse (approved):

okio/src/commonMain/kotlin/okio/internal/Path.kt
swankjesse: Or this:
```suggestion
  if (other.bytes == DOT) {
```

oldergod: :+1:

Issue #1466, opened 2024-04-05
Extension mechanism for FileSystem


swankjesse:
There’s a bunch of FileSystem features that we don’t implement:

 * ACLs & Permissions
 * Locks
 * Watches
 * Volumes & Capacity

We should make it possible to get at extended features like these, without necessarily having all of these features in the core API.

Users could add an interface like this:

```
interface FileLocker {
  fun tryAcquire(path: Path): FileLock?

  interface FileLock : Closeable
}
```

You’d use it like this:

```
val fileSystem: FileSystem = ...

val locker = fileSystem.extension<FileLocker>()
val lock = locker.tryAcquire(path)
...
```

Finally you’d attach extensions in a way similar to `CoroutineContext`:

```
val fileLocker = MyFileLocker()
val fileSystemWithLocker = FileSystem.SYSTEM + fileLocker
```

swankjesse:
It’d also be nice to use this to simplify testing for libraries like SQLite that go direct to the filesystem by default.

```
val database = fileSystem.openDatabase("data.sqlite".toPath())
```

```
fun FileSystem.openDatabase(path: Path): SqliteDatabase {
  extension<SqliteExtension>().open(path)
}

interface SqliteExtension {
  fun open(path: Path): SqliteDatabase
}

/**
 * Returns a file system that includes the extension.
 * 
 * @param inMemory true to return database instances that aren’t durable on disk; appropriate
 *    for use with FakeFileSystem. Otherwise the
 *    databases are written to FileSystem.SYSTEM.
 */
fun applySqlite(
  fileSystem: FileSystem,
  inMemory: Boolean
): FileSystem

**Comment by swankjesse** [comment]
I think we need more stuff to support the `chroot` use case, where one FileSystem maps paths upon another.

Perhaps something like this?

```kotlin
interface FileSystemExtension {
  fun interface Factory<E : FileSystemExtension> {
    fun create(host: FileSystemExtension.Host): E
  }

  interface Host {
    fun onPathParameter(path: Path, functionName: String, parameterName: String): Path
    fun onPathResult(path: Path, functionName: String): Path
  }
}
```

```kotlin
fun <E : FileSystemExtension> FileSystem.extend(
  factory: FileSystemExtension.Factory<E>,
): FileSystem
```

This makes it a bit more annoying to create extensions, but it gives them the stuff they need to do 1:1 path mapping.

Also getting the implementation right is tricky, cause we need to potentially apply multiple layers of mappings.

[...]

PR #1470, opened 2024-04-15
Initial APIs for FileSystem extensions

swankjesse:
Closes: https://github.com/square/okio/issues/1466


okio/src/commonMain/kotlin/okio/internal/FileSystem.kt
```diff
@@ -69,6 +71,23 @@ internal fun FileSystem.commonCopy(source: Path, target: Path) {
   }
 }
 
+fun <T : Any> FileSystem.commonExtend(extensionType: KClass<T>, extension: T): FileSystem {
+  // If this file system is already an extension wrapper, replace it rather than wrapping again.
+  // Note that this optimization doesn't apply to ForwardingFileSystem subclasses, only to the
+  // ForwardingFileSystem base class.
+  if (this::class === ForwardingFileSystem::class) {
```
oldergod, line 78: I don't understand why we don't want to, or cannot support subclasses with `this is ForwardingFileSystem` ?

oldergod (approved):
I like the APIs

okio/src/commonMain/kotlin/okio/internal/FileSystem.kt
swankjesse: This coalesces two ForwardingFileSystem wrappers into one. But it only works if there’s no other methods overridden in the ForwardingFileSystem. If there are, then we shouldn’t replace the users’ subtype with the base type.

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -376,6 +378,18 @@ expect abstract class FileSystem() {
   @Throws(IOException::class)
   abstract fun createSymlink(source: Path, target: Path)
 
+  /**
+   * Returns a new file system that forwards all calls to this, and that also returns [extension]
+   * when it is requested.
+   *
+   * When [extensionType] is requested on the returned file system, it will return [extension],
+   * regardless of what is returned by this file system.
```
oldergod, line 386: So you can shadow extensions ?

oldergod: yes, I sas the `overrideExtension` test

oldergod (approved):

okio/src/commonMain/kotlin/okio/FileSystem.kt
```diff
@@ -376,6 +378,18 @@ expect abstract class FileSystem() {
...
+   */
+  open fun <E : FileSystemExtension> extend(extensionType: KClass<E>, extension: E): FileSystem
```
yschimke, line 388: Should there be a way to freeze these? And reopen via a new builder() type pattern.

Otherwise FileSystem.System is at risk of monkey patching.

yschimke: Not an issue since it returns a new one.

MohammedKHC:
@swankjesse why is this closed?
It seems so good!

PR #1500, opened 2024-08-02
Avoid use symlink in project file

iseki0:
It doesn't work on Windows platform


iseki0:
@oldergod approve this plz. It blocks me

iseki0:
@oldergod Should we re-run it until it success?

PR #1501, opened 2024-08-02
Implement transferTo method in RealBufferedSource.inputStream

iseki0:
Avoid unnecessary memory copying, during transfer data between streams.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/RealBufferedSource.kt
+++ b/okio/src/jvmMain/kotlin/okio/RealBufferedSource.kt
@@ -184,10 +184,9 @@ internal actual class RealBufferedSource actual constructor(
             val read = source.read(buffer, Segment.SIZE.toLong())
             if (read == -1L) break
           }
-          count = try {
-            Math.addExact(count, buffer.size)
-          } catch (ignored: ArithmeticException) {
-            Long.MAX_VALUE
+          count += buffer.size
+          if (count < 0) {
+            count = Long.MAX_VALUE
           }
           buffer.writeTo(out)
         }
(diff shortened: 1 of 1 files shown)


iseki0:
Since I can't run the check process... I manually edit it.

iseki0:
@oldergod Looks like it's not relate to me?

iseki0:
@oldergod 😉 It's OK

iseki0:
@swankjesse Any updates? 😊

okio/src/jvmMain/kotlin/okio/RealBufferedSource.kt
```diff
@@ -174,6 +175,24 @@ internal actual class RealBufferedSource actual constructor(
...
+        while (true) {
+          if (buffer.size == 0L) {
+            val read = source.read(buffer, Segment.SIZE.toLong())
+            if (read == -1L) break
+          }
+          count = try {
+            Math.addExact(count, buffer.size)
+          } catch (ignored: ArithmeticException) {
+            Long.MAX_VALUE
+          }
```
JakeWharton, line 191: I'm assuming the implementation of `addExact` is using a conditional to check for overflow and then throw an exception. So rather than run a conditional, throw, and catch, why don't we just do the conditional ourselves. Plus we know that neither value will ever be negative and can simply check for wrap-around.
```suggestion
          count += buffer.size
          if (count < 0) {
            count = Long.MAX_VALUE
          }
```

JakeWharton (approved):

okio/src/jvmMain/kotlin/okio/RealBufferedSource.kt
iseki0: LGTM, I just following some other implementations.

iseki0:
@JakeWharton done

PR #1507, opened 2024-08-22
fix(common): some HMacSHA tests compared sha values

i10416:
some test cases for Hmac SHA compared sha values instead of hmac sha values.


JakeWharton:
Thanks!

PR #1525, opened 2024-09-24
New API, FileSystem.close()

swankjesse:


okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -218,6 +227,7 @@ class FakeFileSystem(
 
   /** Don't throw [FileNotFoundException] if the path doesn't identify a file. */
   private fun canonicalizeInternal(path: Path): Path {
+    check(!closed) { "closed" }
```
swankjesse, line 230: All other functions call through here

JakeWharton (approved):

PR #1529, opened 2024-10-04
Update FakeFileSystem.kt

GeorgCantor:
This built-in function is highly optimized for copying arrays and will generally perform better than a manual loop due to lower overhead and potential optimizations by the JVM.


okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -732,9 +732,7 @@ class FakeFileSystem(
       val fileOffsetInt = fileOffset.toInt()
       val toCopy = minOf(file.data.size - fileOffsetInt, byteCount)
       if (toCopy <= 0) return -1
-      for (i in 0 until toCopy) {
-        array[i + arrayOffset] = file.data[i + fileOffsetInt]
-      }
+      System.arraycopy(file.data, fileOffsetInt, array, arrayOffset, toCopy)
```
JakeWharton, line 735: We need to use Kotlin's abstraction on this which works in multiplatform contexts.
```suggestion
      file.data.copyInto(array, arrayOffset, fileOffsetInt, fileOffsetInt + toCopy)
```

PR #1531, opened 2024-10-05
Update Buffer.kt

GeorgCantor:
**Reduced Complexity**: The previous nested `if` statements for calculating the width were replaced with a single `when` expression.


JakeWharton:
This change slows down the code, changing it from a biased binary search to a linear scan. We're actually going to change it to a constant time operation, I just haven't gotten around to it yet. See https://www.romainguy.dev/posts/2024/down-another-rabbit-hole/

PR #1548, opened 2024-11-14
Improve algorithm to count digits in Long

Egorand:
Based on "[Down Another Rabbit Hole](https://www.romainguy.dev/posts/2024/down-another-rabbit-hole/)" by @romainguy - thanks!


okio/src/commonMain/kotlin/okio/internal/Buffer.kt
```diff
@@ -539,6 +483,34 @@ internal inline fun Buffer.commonWriteDecimalLong(v: Long): Buffer {
...
+  99999999999L,
+  999999999999L,
+  9999999999999L,
+  99999999999999L,
+  999999999999999L,
+  9999999999999999L,
+  99999999999999999L,
+  999999999999999999L, // For 18 digits (index 18), the largest value is 999999999999999999.
+  Long.MAX_VALUE, // For 19 digits (index 19), the largest value is MAX_VALUE.
+)
```
Egorand, line 512: This is a slight deviation from the original table proposed in the article:

```kotlin
private val PowersOfTen = longArrayOf(
    0,
    10,
    100,
    1000,
    10000,
    100000,
    1000000,
    10000000,
    100000000,
    1000000000,
    10000000000,
    100000000000,
    1000000000000,
    10000000000000,
    100000000000000,
    1000000000000000,
    10000000000000000,
    100000000000000000,
    1000000000000000000
)
```

@swankjesse & I discovered that the original table wouldn't work for the `Long.MAX_VALUE` test case, as it has more digits than 10^18, and 10^19 is outside of the range of Long. The code instead compares against largest numbers that can be represented with a given digit count, which allowed us to fit another entry into the table!

romainguy: I saw that and it's a great improvement!

JakeWharton:
The original code made it up into kotlinx.io. Do you want to send them a PR, too?

Egorand:
> Based on "[Down Another Rabbit Hole](https://www.romainguy.dev/posts/2024/down-another-rabbit-hole/)" by @romainguy - thanks!

Sure, done! https://github.com/Kotlin/kotlinx-io/pull/413

PR #1559, opened 2024-12-17
Tee sample

swankjesse:


swankjesse:
https://en.wikipedia.org/wiki/Tee_(command)

JakeWharton (approved):

samples/src/jvmMain/kotlin/okio/samples/TeeSink.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    sinkB.close()
+  }
+
+  override fun timeout() = sinkA.timeout()
+}
+
+fun main() {
+  val a = System.out.sink()
```
oldergod, line 72: what is this magic

JakeWharton: `System.out` is a `PrintStream`, `PrintStream` is an `OutputStream`, we have `OutputStream.sink()`.

PR #1571, opened 2025-01-08
Correct Node integration's source set directories

JakeWharton:
These should have been renamed when we migrated from the `.js` plugin to `.multiplatform`, but nothing caught the fact that the compilations became empty.

They are, in fact, working now:

Screenshot 2025-01-08 at 11 05 14 AM


JakeWharton:
Er, wtf... we don't block on approval?

PR #1574, opened 2025-01-10
Handle source close while reading on Pipe

nicolashaan:
Handle source close while reading on Pipe.
Does it look like a legit behavior to implement?


swankjesse:
It’s not safe to close a source you’re currently reading. You want pipe.cancel() instead.

swankjesse:
We’d be better off if we had a Source.cancel() API, but I didn’t do that at the time.

Java’s InputStream uses synchronized to kind of support this, but we also don’t do synchronized in Okio. This is a feature!

PR #1597, opened 2025-02-27
Migrate to new sonatype and signing credentials

JakeWharton:


oldergod (approved):
Should I do it for Wire too?

JakeWharton:
I will get to all repos eventually

Issue #1608, opened 2025-03-21
DeflaterSink may use the same buffer for both input and output to Java's Deflater.deflate() call


boyuzhang-msft:
In `DeflaterSink` class in jvmMain, its `write()` method may recycle `head` Segment, which is used as input buffer in `deflater.setInput()` call. Later on this `head` Segment might be returned by `buffer.writableSegment(1)` in `DeflaterSink.deflate()` method, and it's being used as output buffer in `deflater.deflate()` call.

In some workflow (see example below), the same region of this recycled `head` Segment is being used as both input and output buffer to Java's `Deflater` calls, which may result in undefined behavior.

Flow with issue: when I use okhttp to send a compressed message using `webSocket.send("Hello, WebSocket! test")`, okhttp's `WebSocketWriter` calls into `MessageDeflater.deflate()`, which calls into okio's `DeflaterSink.write()` and `DeflaterSink.flush()`. In this flow, the same region of `head` Segment is being used as both input and output buffer to Java's `Deflater` methods.

I added logs to print out the buffer object (`head.data` and `s.data` respectively) used in Java's `deflater.setInput` (after L39 of https://github.com/square/okio/blob/master/okio/src/jvmMain/kotlin/okio/DeflaterSink.kt) and `deflater.deflate` (after L70 of the same file) calls.

```
2025-03-20 10:46:37.276 25536-25657 okhttp3.internal.ws.Mes com.sample.app         I  Hello, WebSocket! test
// This is  DeflaterSink.write() call
2025-03-20 10:46:37.283 25536-25658 okio.DeflaterSink       com.sample.app         I  deflater set input array object: [B@9cf71d5, pos: 0, todeflate: 22}, input data: [72, 101, 108, 108, 111, 44, 32, 87, 101, 98, 83, 111, 99, 107, 101, 116, 33, 32, 116, 101, 115, 116, 0, 0, 0,
2025-03-20 10:46:37.317 25536-25658 okio.DeflaterSink       com.sample.app         I  syncFlush: false, output buffer object: [B@1c08dea, deflated: 0, pos: 0, len: 8192,  s: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 

// This is DeflaterSink.flush() call
[...]

PR #1609, opened 2025-03-21
Do not recycle head Segment in DeflaterSink

boyuzhang-msft:
This change only recycles `head` Segment when the sink is closed, to avoid the risk of using the same buffer for both input and output for Java's Deflater calls.

Fixes bug: https://github.com/square/okio/issues/1608


swankjesse:
Obsoleted by https://github.com/square/okio/pull/1612

PR #1626, opened 2025-05-26
New API, BufferedSource.indexOf(ByteString, fromIndex, toIndex)

swankjesse:
This is surprisingly interesting. To minimize unnecessary reads for toIndex it is necessary to check whether a prefix of the query matches a suffix of the currently-loaded data.

This read-avoidance is useful in practice. When doing HTTP multipart decoding the caller may scan for a boundary separator with a bounded range, and we don't want to block reading when doing so won't impact the result of the call.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/RealBufferedSource.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/RealBufferedSource.kt
@@ -33,6 +33,7 @@ import okio.Segment
 import okio.Sink
 import okio.buffer
 import okio.checkOffsetAndCount
+import okio.minOf
 
 internal inline fun RealBufferedSource.commonRead(sink: Buffer, byteCount: Long): Long {
   require(byteCount >= 0L) { "byteCount < 0: $byteCount" }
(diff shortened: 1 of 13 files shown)


okio/src/commonMain/kotlin/okio/internal/RealBufferedSource.kt
```diff
@@ -332,22 +333,66 @@ internal inline fun RealBufferedSource.commonIndexOf(b: Byte, fromIndex: Long, t
...
+ * 'look', we know the result is -1 without loading more data.
+ */
+private fun matchPossibleByExpandingBuffer(
+  buffer: Buffer,
+  bytes: ByteString,
+  fromIndex: Long,
+  toIndex: Long,
+): Boolean {
+  // Load new data if the match could come entirely in that new data.
+  if (buffer.size < toIndex) return true
```
yschimke, line 382: Should this case be restricted further, considering bytes.size? 

As a contrived example `buffer.size` == 4, `toIndex` = 6 and `bytes` == "elongated"

yschimke (approved):
Looks like a good fit for https://github.com/square/okhttp/pull/8665

But minimal okio review from me.

mpawliszyn (approved):

PR #1633, opened 2025-06-04
New Socket interface

squarejesse:
I haven't yet hooked this up to `java.net.Socket`, but I intend to in a follow-up PR.

### DIFF
--- a/okio/api/okio.api
+++ b/okio/api/okio.api
@@ -374,12 +374,6 @@ public final class okio/ByteString$Companion {
 	public final fun read (Ljava/io/InputStream;I)Lokio/ByteString;
 }
 
-public abstract interface class okio/Channel {
-	public abstract fun cancel ()V
-	public abstract fun getSink ()Lokio/Sink;
-	public abstract fun getSource ()Lokio/Source;
-}
-
 public final class okio/CipherSink : okio/Sink {
 	public fun <init> (Lokio/BufferedSink;Ljavax/crypto/Cipher;)V
 	public fun close ()V
@@ -757,6 +751,12 @@ public abstract interface class okio/Sink : java/io/Closeable, java/io/Flushable
 	public abstract fun flush ()V
 	public abstract fun timeout ()Lokio/Timeout;
 	public abstract fun write (Lokio/Buffer;J)V
+}
+
+public abstract interface class okio/Socket {
+	public abstract fun cancel ()V
+	public abstract fun getSink ()Lokio/Sink;
+	public abstract fun getSource ()Lokio/Source;
 }
 
 public abstract interface class okio/Source : java/io/Closeable {
--- a/okio/src/commonMain/kotlin/okio/Socket.kt
+++ b/okio/src/commonMain/kotlin/okio/Socket.kt
@@ -21,11 +21,11 @@ package okio
  * Send data to the peer by writing to [sink], and read data from the peer by reading from [source].
  *
  * This can be implemented by a plain TCP socket. It can also be layered to add features like
- * security (as in a TLS channel) or connectivity (as in a proxy channel).
+ * security (as in a TLS socket) or connectivity (as in a proxy socket).
  *
  * Closing the [source] does not impact the [sink], and vice versa.
  */
-interface Channel {
+interface Socket {
   val source: Source
   val sink: Sink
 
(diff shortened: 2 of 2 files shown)


okio/src/commonMain/kotlin/okio/Channel.kt
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+package okio
+
+/**
+ * A pair of streams for interactive communication with another machine.
+ *
+ * Send data to the peer by writing to [sink], and read data from the peer by reading from [source].
+ *
+ * This can be implemented by a plain TCP socket. It can also be layered to add features like
```
yschimke: The Socket (TCP, SSL) case is interesting, because I assume we'd want some of the features from https://github.com/square/okhttp/pull/8410/files#diff-e108767b1cd01f2dccec45ffa13f1463516eed6ebebf1c25e8bb961bc88922f8

```
  var soTimeout: Int
  val peerName: String
  val localPort: Int
  val inetAddress: InetAddress?
  val localAddress: InetAddress
  val remoteSocketAddress: SocketAddress?
  ```

File: okio/src/commonMain/kotlin/okio/Channel.kt [hunk 2]
```diff
@@ -0,0 +1,44 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio
+
+/**
+ * A pair of streams for interactive communication with another machine.
```
yschimke: is Channel always open when created?

Can it be created for something being opened? or created before opening?

yschimke: Why specifically "another machine"?

okio/src/commonMain/kotlin/okio/Channel.kt
```diff
@@ -0,0 +1,44 @@
...
+ * security (as in a TLS channel) or connectivity (as in a proxy channel).
+ *
+ * Closing the [source] does not impact the [sink], and vice versa.
+ */
+interface Channel {
+  val source: Source
+  val sink: Sink
+
+  /**
+   * Fail any in-flight and future operations. After canceling:
```
yschimke: Comparing with https://github.com/square/okhttp/blob/a1863a71f476558389a2f855ca12b34c8eb2a118/mockwebserver/src/main/kotlin/mockwebserver3/internal/duplex/RealStream.kt#L30

That seems have an orderly shutdown, while this seems immediately disruptive.  Mainly because of the "in-flight" statement.

Is that intended?

JakeWharton:
Are there any other naming ideas for this? Overloading "channel" from coroutines but as a single-ended bidirectional thing as opposed to a dual-ended unidirectional thing doesn't seem ideal.

I don't really have any good suggestions though…

okio/src/commonMain/kotlin/okio/Channel.kt
squarejesse: I think we’d nee those on a specific concrete implementation, but the interface itself doesn’t need that stuff. Kind of like how InputStream doesn’t tell you which `File` it’s reading from.

okio/src/commonMain/kotlin/okio/Channel.kt
squarejesse: I think it’s the same as for `Source` . . . it’s created open and its lifecycle ends when it’s streams are both closed.

squarejesse: I was inspired by `java.net.Socket` which has similar language:

> A socket is an endpoint for communication between two machines.

In particular I wanted to draw a distinction between this and `Pipe`, which is a single stream confined to a single machine.

okio/src/commonMain/kotlin/okio/Channel.kt
squarejesse: I think they’re both supposed to be immediately disruptive. I think of cancel as ‘please stop consuming any resources, the work you’re doing is no longer necessary’

squarejesse:
@JakeWharton I think you’re right about the name collision with Kotlin’s `Channel` type. I think `Socket` is the next best name here. It’s consistent with what Java people would expect. The name collision with `java.net.Socket` is unfortunate, though we’ve had a name collision with Java’s `Buffer` forever and it really hasn’t caused much grief.

PR #1636, opened 2025-06-11
New concrete implementation of okio.Socket

swankjesse:
This turned out to be much trickier than I had expected. The default java.net.Socket inputStream and outputStream close the entire Socket when they are closed. This offers different behavior - the two streams are closed independently. When both are closed the socket is closed and its resources are released.

### DIFF
--- a/okio/src/jvmTest/kotlin/okio/SocketTest.kt
+++ b/okio/src/jvmTest/kotlin/okio/SocketTest.kt
@@ -253,6 +253,7 @@ class SocketTest(val factory: Factory = Factory.Default) {
     assertThat(javaNetSocket.isClosed).isFalse()
   }
 
+  @Suppress("ktlint:trailing-comma-on-declaration-site")
   enum class Factory {
     /** Implements an okio.Socket using the `java.net.Socket` API on OS sockets. */
     Default {


okio/src/commonMain/kotlin/okio/Socket.kt
```diff
@@ -24,6 +24,17 @@ package okio
...
+ *
+ * You must close the source and the sink to release the resources held by this socket. If you're using both from the
+ * same thread, you can do that with nested `use` blocks:
+ *
+ * ```kotlin
+ * socket.source.use { source ->
+ *   socket.sink.use { sink ->
+ *     readAndWrite(source, sink)
+ *   }
+ * }
```
swankjesse, line 36: I decided against including a `close()` function on `okio.Socket` because it’s too difficult to use correctly. In particular if when you close the whole darn `Socket` that’s very often going to be a race.

okio/src/jvmMain/kotlin/okio/internal/-Atomic.kt
```diff
@@ -0,0 +1,31 @@
...
+ * Returns the new value of the bit field if a change was made, or 0 if no change was made.
+ */
+internal fun AtomicInteger.setBitOrZero(bit: Int): Int {
+  while (true) {
+    val current = get()
+    if (current and bit != 0) return 0 // Bit is already set.
+    val updated = current or bit
+    if (compareAndSet(current, updated)) return updated
+  }
+}
```
swankjesse, line 31: Little side quest here. I’m not that excited about bitpacking our state, but I do really want to atomically set two booleans.

okio/src/jvmTest/kotlin/okio/AsyncSocket.kt
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.LinkedBlockingDeque
+import kotlin.concurrent.thread
+
+/**
+ * Queues reads and writes onto a background thread. This prevents potential deadlock if such
+ * operations were done on the caller's thread.
+ */
+class AsyncSocket(
```
swankjesse, line 26: Test-only code!

okio/src/jvmMain/kotlin/okio/internal/-Atomic.kt
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+package okio.internal
+
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * Returns the new value of the bit field if a change was made, or 0 if no change was made.
+ */
+internal fun AtomicInteger.setBitOrZero(bit: Int): Int {
```
JakeWharton: Since you're not doing a shift of 1 with this value, I would call it instead something like a mask or bits plural.

JakeWharton (approved):

okio/src/jvmMain/kotlin/okio/internal/-Atomic.kt
swankjesse: Good call. Done!

okio/src/jvmMain/kotlin/okio/JvmOkio.kt
```diff
@@ -139,33 +139,8 @@ fun Socket.source(): Source {
...
-        // https://code.google.com/p/android/issues/detail?id=54072
-        logger.log(Level.WARNING, "Failed to close timed out socket $socket", e)
-      } else {
-        throw e
-      }
-    }
-  }
-}
+@JvmName("socket")
+fun Socket.asOkioSocket(): okio.Socket = DefaultSocket(this)
```
ktabouguia-sq, line 143: 😎

okio/src/jvmTest/kotlin/okio/AsyncSocket.kt
ktabouguia-sq: Nice!

okio/src/jvmTest/kotlin/okio/SocketTest.kt
```diff
@@ -0,0 +1,303 @@
...
+            sinkPipe.cancel()
+          }
+        }
+
+        val aToB = Pipe(1024)
+        val bToA = Pipe(1024)
+
+        return PipeSocket(aToB, bToA) to PipeSocket(bToA, aToB)
+      }
+    };
```
oldergod, line 299: That's very cool

okio/src/jvmTest/kotlin/okio/SocketTest.kt
```diff
@@ -0,0 +1,303 @@
...
+        val socketA = SocketFactory.getDefault().createSocket()
+        socketA.connect(InetSocketAddress(localhost, serverSocket.localPort))
+
+        val socketB = socketBFuture.get()
+        return socketA.asOkioSocket() to socketB.asOkioSocket()
+      }
+    },
+
+    /** Implements an okio.Socket using a pair of in-memory pipes. */
+    Pipes {
```
ktabouguia-sq, line 280: Did you also intend to use this in tests?

ktabouguia-sq (approved):
Looks great!! 🎉

okio/src/jvmTest/kotlin/okio/SocketTest.kt
swankjesse: The tests exercise both thanks to `@Burst`.

PR #1638, opened 2025-06-12
New API, inMemorySocketPair()

swankjesse:
I'm using Array<Socket> instead of Pair<Socket, Socket> because Array is _slightly_ more friendly to Java-language consumers.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/internal/PipeSocket.kt
+++ b/okio/src/jvmMain/kotlin/okio/internal/PipeSocket.kt
@@ -1,0 +1,34 @@
+/*
+ * Copyright (C) 2025 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package okio.internal
+
+import okio.Pipe
+import okio.Sink
+import okio.Socket
+import okio.Source
+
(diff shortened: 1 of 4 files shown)


okio/src/jvmMain/kotlin/okio/internal/PipeSocket.kt
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package okio.internal
+
+import okio.Pipe
+import okio.Sink
+import okio.Socket
+import okio.Source
+
+internal class PipeSocket(val sinkPipe: Pipe, val sourcePipe: Pipe) : Socket {
```
swankjesse, line 23: Is this the right name? `InMemorySocket` might better fit the public API?

okio/src/jvmTest/kotlin/okio/SocketTest.kt
```diff
@@ -272,33 +272,15 @@ class SocketTest(val factory: Factory = Factory.Default) {
...
-
-        return PipeSocket(aToB, bToA) to PipeSocket(bToA, aToB)
-      }
+      override fun createSocketPair() = inMemorySocketPair(1024)
     };
 
-    /** Returns a pair of mutually-connected sockets. */
-    abstract fun createSocketPair(): Pair<Socket, Socket>
+    /** Returns two mutually-connected sockets. */
+    abstract fun createSocketPair(): Array<Socket>
```
swankjesse, line 284: I checked our API dump and there’s very little we’re exporting from the kotlin stdlib in Okio’s API. The only one I saw was `Options` and `TypedOptions` both expose `kotlin.collections.AbstractList` as their supertype, instead of `java.util.AbstractList`.

JakeWharton (approved):

okio/src/jvmMain/kotlin/okio/internal/PipeSocket.kt
JakeWharton: Is it intended to be public in the future?

swankjesse: Nope, not public. But the name will show up in toString() etc.

JakeWharton: I like this name

PR #1645, opened 2025-06-25
Update to kotlinx-datetime:0.7.0 and fix binary incompatible changes.

colinrtwhite:
`kotlinx-datetime:0.7.0` completely removes `kotlinx.datetime.Clock` and `kotlinx.datetime.Instant` in favour of the `kotlin.time` versions ([more info](https://github.com/Kotlin/kotlinx-datetime?tab=readme-ov-file#deprecation-of-instant)). This creates a couple issues:

- We currently use `kotlinx.datetime.Instant` in `FakeFileSystem`'s public API. We probably have to take a binary incompatible change here and update to the new class.
- `kotlin.time` classes are marked with `@ExperimentalTime` which will propagate to our `FakeFileSystem`'s constructor  unless we add a secondary constructor: `constructor() : this(Clock.System)` that hides the experimental arg.
- Currently all okio consumers have to use `0.7.0-0.6.x-compat` (a compat version that retains the removed classes) if they want to update to `0.7.0` and use `FakeFileSystem`.

### DIFF
--- a/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
+++ b/okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
@@ -18,8 +18,8 @@ package okio.fakefilesystem
 import kotlin.jvm.JvmField
 import kotlin.jvm.JvmName
 import kotlin.reflect.KClass
-import kotlinx.datetime.Clock
-import kotlinx.datetime.Instant
+import kotlin.time.Clock
+import kotlin.time.Instant
 import okio.ArrayIndexOutOfBoundsException
 import okio.Buffer
 import okio.ByteString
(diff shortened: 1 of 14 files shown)


JakeWharton:
> We probably have to take a binary incompatible change here and update to the new class.

I believe this was the concession when we originally took the unstable dependency. It's a test only dependency (or should be) so it's okay that we take this hit.

okio-fakefilesystem/src/commonMain/kotlin/okio/fakefilesystem/FakeFileSystem.kt
```diff
@@ -71,6 +71,9 @@ class FakeFileSystem(
   val clock: Clock = Clock.System,
 ) : FileSystem() {
 
+  // Avoid propagating 'ExperimentalTime' if 'clock' is not specified.
+  constructor() : this(Clock.System)
```
colinrtwhite, line 75: Not sure if it's worth maintaining this secondary constructor forever to avoid propagating `@ExperimentalTime` to `FakeFileSystem()`. Let me know what you think

JakeWharton: Seems like a good idea. We can do `@JvmOverloads` once this type goes stable.

swankjesse (approved):
Thanks for all the attention to compatibility!

mathieularue:
Is there any chance that this will be included in an upcoming release? Thanks!

JakeWharton:
It will be released today or tomorrow, yes.

swankjesse:
When this build completes, the release is out. https://github.com/square/okio/actions/runs/15945624170

PR #1662, opened 2025-07-15
feat(JVM): BufferedSource.utf8CodePoints

MukjepScarlet:
Reference to CharSequence.codePoints


MukjepScarlet:
I'm not sure if it's applicable to place the implementation in the interface file. I think the implementations of this function are same in `RealBufferedSource` and `Buffer` because it only depends on the implementation of `readUtf8CodePoint`.

okio/src/jvmMain/kotlin/okio/BufferedSource.kt
```diff
@@ -121,6 +123,11 @@ actual sealed interface BufferedSource : Source, ReadableByteChannel {
   @Throws(IOException::class)
   actual fun readUtf8CodePoint(): Int
 
+  /**
+   * Returns an [IntStream] of UTF-8 code point values from this source.
+   */
+  fun utf8CodePoints(): IntStream = commonUtf8CodePoints()
```
JakeWharton: I'm not sure we want to define this JVM-only API when Kotlin is likely to provide a `Codepoint` type in the future that we can put into common.

okio/src/jvmMain/kotlin/okio/internal/BufferedSource.kt
```diff
@@ -0,0 +1,40 @@
...
+package okio.internal
+
+import java.util.PrimitiveIterator
+import java.util.Spliterator
+import java.util.Spliterators
+import java.util.stream.IntStream
+import java.util.stream.StreamSupport
+import okio.BufferedSource
+
+internal fun BufferedSource.commonUtf8CodePoints(): IntStream {
```
JakeWharton: This is only called from one place and on only one platform. Seems like it should just be inlined to that call site as there's nothing "common" about it.

MukjepScarlet: Is it okay to put it in the BufferedSource interface (JVM)?

JakeWharton: I'd say either there or as an extension function. I guess the question is whether we want to expose it on the interface for Java users (assuming we want to expose it at all).

MukjepScarlet: Maybe we can provide it as a util(extension) function instead of a member.

okio/src/jvmMain/kotlin/okio/BufferedSource.kt
MukjepScarlet: Will it be like a value class wrapping `Int` or `UInt`? If we want to provide a Java-friendly API, a wrapped Codepoint might be bad because we don't have structs on JVM :p

squarejesse:
Here’s the `Sequence` one-liner:

```
    val sequence = generateSequence { 
      if (!source.exhausted()) source.readUtf8CodePoint() else null
    }
```

This one isn’t an `IntStream`, and it boxes the ints, so it isn’t a perfect comparison.

swankjesse:
Closing this 'cause it’s very easy to do on-top of the base library!

Issue #1603, opened 2025-03-10
BufferedSource.readUtf8Line as Sequence<String>(kt) and Stream<String>(java)


MukjepScarlet:
We have `BufferedReader.lines(): Stream<String>` in Java, so I hope that `BufferedSource` can also integrate similar APIs.

MukjepScarlet:
I did an extension for that:
```kotlin
fun BufferedSource.utf8Lines(): Iterator<String> =
    object : AbstractIterator<String>() {
        override fun computeNext() {
            val nextLine = readUtf8Line()
            if (nextLine != null) {
                setNext(nextLine)
            } else {
                close()
                done()
            }
        }
    }
```
One issue is that `IOException` is ignored here

MukjepScarlet:
I plan to do this...

PR #1663, opened 2025-07-15
feat: BufferedSource.utf8LineSequence

MukjepScarlet:
The `BufferedReader` of Java has member `lines(): Stream<String>` since Java 8, and Kotlin provides `BufferedReader.lineSequence(): Sequence<String>`. This PR is in order to create similar stream-style functions for BufferedSource too.

TODO: Tests and docs are missing now

closes https://github.com/square/okio/issues/1603


swankjesse:
Closing this 'cause it’s very easy to do on-top of the base library! Especially when using `exhausted()` to check before attempting to read a line.

PR #1683, opened 2025-08-19
refactor: improve Options.of

MukjepScarlet:
Replace `List<Int>` with `IntArray`. Less memory usage on JVM.
Replace `removeAt` with `subList`. Zero-copy.

The only regret is I have to copy the array because KMP doesn't have `Array.binarySearch`.

### DIFF
--- a/okio/src/commonMain/kotlin/okio/Options.kt
+++ b/okio/src/commonMain/kotlin/okio/Options.kt
@@ -33,47 +33,63 @@ class Options private constructor(
   override fun get(index: Int) = byteStrings[index]
 
   companion object {
+    private val EMPTY = Options(emptyArray(), intArrayOf(0, -1))
+
     @JvmStatic
     fun of(vararg byteStrings: ByteString): Options {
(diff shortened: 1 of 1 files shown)


swankjesse (commented):
I don’t believe this is going to be a performance improvement in practice.

For example, `isRemoved` is unlikely to be true in practice, so optimizing for that case is unnecessary.

What led you to make these changes? Have you observed `Options.of()` causing performance problems? I suspect the best mitigation is to cache the result of it.

MukjepScarlet:
The main point is `removeAt`'s array allocation (if user don't know the behavior).

> I suspect the best mitigation is to cache the result of it.

Yes... Unexperienced users usually create it on each usage. But this action seems to be out of control of the lib itself.

swankjesse:
I’m going to close this ’cause I don’t think it’s the best way to achieve our performance goals. Reusing an `Options` instance is gonna work better than making the `Options` construction more efficient!

PR #1685, opened 2025-08-26
Adopt Burst's new TestInterceptors

swankjesse:


swankjesse:
This won’t work until TestInterceptors are available in a proper Burst release. But I wanted to confirm the API looks right and Okio seemed like a good enough candidate.

okio-testing-support/src/jvmMain/kotlin/okio/TestExecutor.kt
```diff
@@ -0,0 +1,63 @@
...
+import java.util.concurrent.ScheduledExecutorService
+import java.util.concurrent.ScheduledFuture
+import java.util.concurrent.ThreadFactory
+import java.util.concurrent.TimeUnit
+import kotlin.time.Duration
+
+/**
+ * Manages a `ExecutorService` and shuts it down after the test.
+ */
+class TestExecutor(
```
swankjesse, line 31: Not totally sold on this name, as it doesn’t execute tests

JakeWharton: TestingExecutor?

JakeWharton (approved):

PR #1690, opened 2025-08-30
Set apiVersion and languageVersion to 2.0 for backwards compatibility

hugoncosta:
Following the discussion in https://github.com/square/okhttp/issues/9032 and the PR in https://github.com/square/okhttp/pull/9040, the purpose of this change is to ensure Okio remains compatible with Kotlin 2.0. 

My change is only within the Okio package - to make the entire project compatible, there would need to be a few changes in other modules, more specifically adapting the usage of the new Kotlin Clock (see https://github.com/search?q=repo%3Asquare%2Fokio%20kotlin.time.Clock&type=code).


hugoncosta:
Similar to the OkHttp PR mentioned above, limiting the scope of this change by only ensuring that the source sets remain compatible with Kotlin 2.0 - tests, which are already using 2.1+ features (like the new Clock and Instant) can remain on the latest version.

okio/build.gradle.kts
```diff
@@ -63,6 +63,14 @@ kotlin {
...
       }
     }
+    // Some sourceSets do not end with Test, easier to use the negative
+    matching { !it.name.endsWith("Test") }.all {
+      // Setting these to allow for backwards compatibility
+      languageSettings {
+        apiVersion = "2.0"
+        languageVersion = "2.0"
+      }
+    }
```
Goooler, line 73:
```suggestion
  tasks.withType<KotlinCompile>().configureEach {
    if (name.contains("test")) return@configureEach // Skip test source sets.
  
    compilerOptions {
      // Pin language level to 2.0 to ensure compatibility with older Gradle versions and other libraries that depend on okhttp.
      // https://docs.gradle.org/current/userguide/compatibility.html#kotlin
      languageVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
      apiVersion = org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
    }
  }
```

yschimke: This caused some issues committing to the OkHttp PR

https://github.com/square/okhttp/pull/9040

omarismail94:
You probably also want to set the `coreLibrariesVersion` field in the `kotlin` extension to `2.0.21` Something like:
```
  afterEvaluate {
    extensions.getByType(KotlinProjectExtension::class.java)
      .coreLibrariesVersion = "2.0.21"
  }
```

JakeWharton:
Thanks for kicking this off. This has been fixed in https://github.com/square/okio/pull/1770.

PR #1714, opened 2025-10-09
Recover from an SSLSocket crash

swankjesse:
See: https://github.com/square/okhttp/issues/9123


swankjesse:
FYI @yschimke I changed it to use `InputStream.close()` / `OutputStream.close()` instead. These bottom out in the same code but I think closing the stream is more semantic.

Issue #1728, opened 2025-10-25
Add the ability to create `Path` as-is from ByteString


MohammedKHC:
I'm writing a git clone in Kotlin and relying on Okio paths for various functionalities.

A key challenge is the creation of symlinks. For example, I need to create a symlink that points to ./README.md.

Currently, using String.toPath resolves the path as README.md. While both ./README.md and README.md refer to the same file, they are not treated identically in the context of Git. Git hashes the symlink's source and target as is. Any modification to this representation (e.g., from ./README.md to README.md) causes Git to register the file as having changed.

This issue is related to #1460.

Proposed Solution
Simply make the constructor Path(ByteString) public. This would allow for the creation of a Path object that retains the exact byte representation required for Git's hashing mechanism.

Does this proposal seem acceptable?

squarejesse:
I don’t think Okio’s Path is a good fit for your use case. You should consider forking it? Maybe create `GitPath` that better fits Git’s model of a path.

MohammedKHC:
@squarejesse
I think forking it will just complicate things.
Everything else works extremely well!
Deflater, Inflater, Buffer.
I just want to be able to create a Path as-is.
And this will likely be useful for other things as well.

I think creating unsafeCreate, UnsafePathApi (like @JakeWharton said) seems reasonable.

Thanks for your time!

squarejesse:
No problem!

Agreed, don’t fork the whole project! But if you need to do fancy stuff with paths, you might want to copy-paste parts from the Path class. There’s some code in here that you might need.
https://github.com/square/okio/blob/master/okio/src/commonMain/kotlin/okio/internal/Path.kt

MohammedKHC:
@squarejesse I really want to be able to use the Okio Path, and it really works well, it's just that sometimes I need the path to be as-is. In java for example if i do `java.nio.file.Paths.get("./hello/./hello/../hi")` and then call toString it I would get the same ugly path.
It would only be normalized when I call `Path.normalize`
But in okio when I create the same path with or without normalizing it will remove the leading dot.
Can you please reconsider it?

Thanks again for your time! 
And no problem if you don't want to do it. Thanks for your hard work and responsiveness!

swankjesse:
Yeah. The main reason I don't want to do it is cause it creates new cases to handle for other code that operates on Path instances.

PR #1729, opened 2025-10-25
Add Path.unsafeCreate(bytes: ByteString)

MohammedKHC:
Closes: #1728

### DIFF
--- a/okio/api/okio.api
+++ b/okio/api/okio.api
@@ -734,6 +734,7 @@ public final class okio/Path$Companion {
 	public static synthetic fun get$default (Lokio/Path$Companion;Ljava/io/File;ZILjava/lang/Object;)Lokio/Path;
 	public static synthetic fun get$default (Lokio/Path$Companion;Ljava/lang/String;ZILjava/lang/Object;)Lokio/Path;
 	public static synthetic fun get$default (Lokio/Path$Companion;Ljava/nio/file/Path;ZILjava/lang/Object;)Lokio/Path;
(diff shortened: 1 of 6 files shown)


JakeWharton:
Ignoring whether or not it makes sense to allow this, it should not be so easy to create a `Path` this way. It should be `Path.unsafeCreate(ByteString)` or `Path.uncheckedOf(ByteString)` or something, and probably come with an opt-in annotation for `@UnsafePathApi`.

MohammedKHC:
Fair enough. Will implement unsafeCreate, UnsafePathApi
This just makes more sense.

MohammedKHC:
I replaced it with `Path.unsafeCreate(bytes: ByteString)`
And also marked it with UnsafePathApi. (requried opt-in).
And added tests.
And basic examples.
Also fixed the okio.api.

MohammedKHC:
The ci failed.., What does this even mean??
> There were 1 lint error(s), they must be fixed or suppressed.
  src/commonTest/kotlin/okio/PathTest.kt:LINE_UNDEFINED ktlint(java.lang.NoClassDefFoundError) Could not initialize class org.jetbrains.kotlin.com.intellij.openapi.util.objectTree.ThrowableInterner (...)
  Resolve these lints or suppress with `suppressLintsFor`

squarejesse:
Added a note on the issue. Thanks for the submission but I think it’s better for Okio to not attempt to satisfy this use case!

PR #1733, opened 2025-11-06
Apply @IgnorableReturnValue where appropriate in Okio

swankjesse:


swankjesse:
I’ve added `@file:MustUseReturnValue` to every file in the public API. (Can I do this at the module level?)

And I’ve added `@IgnorableReturnValue` to every function that fits. I was pretty generous with these and put them on all the overrides and expect/actuals.

I haven’t done Okio-internals for us to take advantage of this capability internally! We can do that later, with a bit less urgency.

swankjesse:
Ooooh I can save myself a lot of boilerplate with this:

```
kotlin {
    compilerOptions {
        freeCompilerArgs.add("-Xreturn-value-checker=full")
    }
}
```

Will do.

swankjesse:
*  [ ] Don’t merge this ’til we’re on Kotlin 2.3.0, since the annotations themselves are changing.

JakeWharton:
I think this might end up being at odds with #1690.

Also not sure what implications it has, but the annotation has been renamed in Kotlin 2.3.0 to be plural.

Issue #1737, opened 2025-11-14
Nested AsyncTimeout seem to conflict


yschimke:
Not sure if Okio or OkHttp bug.

See https://github.com/square/okhttp/pull/9193

Given

callTimeout - 3 seconds
readTimeout - 2 seconds
2 hosts (so readTimeout used twice)

Then we get failure after 4 seconds, instead of expected 3

But if we **increase* readTimeout

callTimeout - 3 seconds
readTimeout - 4 seconds
2 hosts (so readTimeout used twice)

Then we get failure after 3 seconds.

So increasing a configured timeout, decreases the time to timeout.

I'm wondering whether readTimeout firing, is somehow resetting the callTimeout?

PR #1738, opened 2025-11-14
Fix compareTo function for priority queue in Asynctimeout

ktabouguia-sq:
## Description

Fixes: https://github.com/square/okio/issues/1737

The  priority queue used by the async timeout class should return the timeout with earliest expiration time, but it was instead returning the timeout with the smallest `timeoutNanos` value.

This PR fixes that issue by updating the `AsyncTimeout.CompareTo` function compare the `timeoutAt` value instead of the `timeoutNanos` value.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
+++ b/okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
@@ -49,7 +49,8 @@ open class AsyncTimeout : Timeout() {
   @JvmField internal var index: Int = -1
 
   /** If scheduled, this is the time that the watchdog should time this out.  */
-  private var timeoutAt: Long = 0L
+  internal var timeoutAt: Long = 0L
+    private set
 
   fun enter() {
     val timeoutNanos = timeoutNanos()
@@ -96,9 +97,6 @@ open class AsyncTimeout : Timeout() {
    * elapsed and the timeout should occur immediately.
    */
   internal fun remainingNanos(now: Long) = timeoutAt - now
-
-  /** If scheduled, this is the expiration time of this timeout  */
-  internal fun timeoutAt(): Long = timeoutAt
 
   /**
    * Sets the timeoutAt value as a sum of [now] and the time to wait for this timeout.
@@ -508,8 +506,8 @@ internal class PriorityQueue {
    */
   @Suppress("NOTHING_TO_INLINE")
   private inline operator fun AsyncTimeout.compareTo(other: AsyncTimeout): Int {
-    val a = timeoutAt()
-    val b = other.timeoutAt()
+    val a = timeoutAt
+    val b = other.timeoutAt
     return 0L.compareTo(b - a)
   }
 }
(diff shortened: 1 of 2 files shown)


okio/src/jvmTest/kotlin/okio/internal/PriorityQueueTest.kt
```diff
@@ -47,6 +50,50 @@ class PriorityQueueTest {
...
+      .let { before.add(it) }
+
+    assertThat(before.toDebugString()).isEqualTo(
+      """
+      |...............5................
+      |......10..............20........
+      |..28......29......21.......2....
+      |
+      """.trimMargin(),
+    )
```
ktabouguia-sq, line 70: I wonder if we should update these strings now to use the timeoutAt value. 
Now they are are not as intuitive as they were intended to be given that we are using timeoutAt value, not timeoutNanos.

ktabouguia-sq: I can achieve that by setting now to 0, for example

okio/src/jvmTest/kotlin/okio/internal/PriorityQueueTest.kt
```diff
@@ -20,12 +20,15 @@ import assertk.assertions.isEqualTo
...
+import kotlin.time.Duration.Companion.nanoseconds
+import kotlin.time.Duration.Companion.seconds
 import okio.AsyncTimeout
 import okio.PriorityQueue
 import org.junit.Test
 
 class PriorityQueueTest {
-  private val TIME_UNIT = TimeUnit.SECONDS
+  private val TIME_UNIT = TimeUnit.NANOSECONDS
+  var now = 0L
```
ktabouguia-sq: We are now using timeoutAt value in node value in debug string, so set now to 0 epoch nanos to simplify debug string formatting below.

swankjesse: Nice! This is a `var` but it isn’t mutated anywhere?

ktabouguia-sq: Ah right. Let me update it!

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
```diff
@@ -97,6 +97,28 @@ open class AsyncTimeout : Timeout() {
    */
   internal fun remainingNanos(now: Long) = timeoutAt - now
 
+  /** If scheduled, this is the expiration time of this timeout  */
+  internal fun timeoutAt(): Long = timeoutAt
```
swankjesse: nit: I think callers should just read the field. It’s also internal?

swankjesse: Or update `timeoutAt` to be like so?

```
internal var timeoutAt: Long = 0L
  private set
```

swankjesse (approved):

okio/src/jvmMain/kotlin/okio/AsyncTimeout.kt
ktabouguia-sq: Updated!

PR #1726, opened 2025-10-23
Add androidNative targets.

MohammedKHC:
This adds support for androidNative targers.
At first, I added a new sourceSet for androidNative and copied the LinuxPosixVariant into it.
And it worked without any modifications, so I removed the androidNative sourceSet and added androidNative targets to linuxTargets. I think this is fine since android is based on linux.

Also DEFFILEMODE is not avaliable on androidNative so i replaced it with
```kotlin
private const val DEFFILEMODE = 0b110110110 /* octal 666 */
```
Based on my small research both are the same.
But please tell me if you think there are other ways.

Closes: #1692

PR #1739, opened 2025-11-15
Add androidNative targets.

### LINKED ISSUE #1242: Consider supporting all official K/N targets
In `kotlinx.*` libraries we finally publish artifacts for all officially-supported targets. To mitigate potential questions about various implementations aspects (including the level of support in tooling, overall quality level and regular testing), we refer to [official recommendations](https://kotlinlang.org/docs/native-target-support.html#for-library-authors).

It would be nice if Okio could do the same in order to make the overall targets support across the libraries more consistent (i.e. will enable other libraries [that depend on Okio] to support lower-tier targets); in particular, we've already encountered it in our `kotlinx-serialization-okio` integration: https://github.com/Kotlin/kotlinx.serialization/blob/79076e26ef860967688403d27cd8313d8bf9ba66/gradle/native-targets.gradle#L111

### LINKED ISSUE #1692: Support for androidNativeX64 and androidNativeArm64 targets
Hello!

Do you plan to add support for `androidNativeX64` and `androidNativeArm64` targets?
Especially for the file-system API.
I could probably try to provide a PR, but I will need some help in this.

Thanks a lot

MohammedKHC:
This adds support for androidNative targets.

This is the second attempt at trying to do this. Now I can confirm that it does compile and work.

The problem with the first attempt (#1726) was that it didn't account for the fact that platform.posix.DIR doesn't exist on androidNative. Therefore, opendir, readdir, and closedir can't be used from common code. This fixes this by using expect/actual for all this stuff.

To run the tests. Run the following commands in order.
(You need to have an Android Emulator running first.)
```
./gradlew androidNativeX64TestBinaries
adb push /path/to/okio/okio/build/bin/androidNativeX64/debugTest/test.kexe /data/local/tmp
rm -r /path/to/okio/build
adb push /path/to/okio /data/local/tmp/okio
adb shell
export OKIO_ROOT=/data/local/tmp/okio
/data/local/tmp/test.kexe
```

Closes: #1692, Closes #1242


MohammedKHC:
<details>
  <summary>Outdated..</summary>
To run the tests on the Android Emulator run
```
./gradlew androidNativeX64TestBinaries && adb push okio/build/bin/androidNativeX64/debugTest/test.kexe /data/local/tmp && adb shell /data/local/tmp/test.kexe
```

For me all the tests passed except those:
```
[  FAILED  ] 23 tests, listed below:
[  FAILED  ] okio.NativeSystemFileSystemTest.listOnRelativePathWhichIsNotDotReturnsRelativePaths
[  FAILED  ] okio.NativeSystemFileSystemTest.listOrNullOnRelativePathWhichIsNotDotReturnsRelativePaths
[  FAILED  ] okio.ZipFileSystemGoTest.timeWinzip
[  FAILED  ] okio.ZipFileSystemTest.emptyZip
[  FAILED  ] okio.ZipFileSystemTest.emptyZipWithPrependedData
[  FAILED  ] okio.ZipFileSystemTest.zipWithFiles
[  FAILED  ] okio.ZipFileSystemTest.zipWithDeflate
[  FAILED  ] okio.ZipFileSystemTest.zipWithStore
[  FAILED  ] okio.ZipFileSystemTest.zipWithFileComments
[  FAILED  ] okio.ZipFileSystemTest.zipWithFileModifiedDate
[  FAILED  ] okio.ZipFileSystemTest.zipWithFileOutOfBoundsModifiedDate
[  FAILED  ] okio.ZipFileSystemTest.zipWithDirectoryModifiedDate
[  FAILED  ] okio.ZipFileSystemTest.zipWithModifiedDate
[  FAILED  ] okio.ZipFileSystemTest.zipWithEmptyDirectory
[  FAILED  ] okio.ZipFileSystemTest.zipWithSyntheticDirectory
[  FAILED  ] okio.ZipFileSystemTest.zip64
[  FAILED  ] okio.ZipFileSystemTest.zipWithArchiveComment
[  FAILED  ] okio.ZipFileSystemTest.cannotReadZipWithSpanning
[  FAILED  ] okio.ZipFileSystemTest.cannotReadZipWithEncryption
[  FAILED  ] okio.ZipFileSystemTest.zipTooShort
[  FAILED  ] okio.ZipFileSystemTest.filesOverlap
[  FAILED  ] okio.ZipFileSystemTest.canonicalizationValid
[  FAILED  ] okio.ZipFileSystemTest.canonicalizationInvalidThrows
```
Will investigate further and update you.
</details>

MohammedKHC:
<details>
  <summary>Also, Outdated</summary>
UPDATE: I got most tests running.
Those are the only failing tests.
```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] okio.NativeSystemFileSystemTest.canonicalizeDotReturnsCurrentWorkingDirectory
[  FAILED  ] okio.NativeSystemFileSystemTest.listOnRelativePathWhichIsNotDotReturnsRelativePaths
[  FAILED  ] okio.NativeSystemFileSystemTest.listOrNullOnRelativePathWhichIsNotDotReturnsRelativePaths
```

To run tests now, do the following on the terminal.
```
./gradlew androidNativeX64TestBinaries
adb push /path/to/okio/okio/build/bin/androidNativeX64/debugTest/test.kexe /data/local/tmp
rm -r /path/to/okio/build
adb push /path/to/okio /data/local/tmp/okio
adb shell
cd /data/local/tmp
export OKIO_ROOT=/data/local/tmp/okio
./test.kexe
```
</details>

MohammedKHC:
UPDATE again.
`canonicalizeDotReturnsCurrentWorkingDirectory` fails because i changed the cwd to /data/local/tmp..
The test expects it to be "/". And it does not fail if i don't do the cd.

Now the only failling tests are `listOnRelativePathWhichIsNotDotReturnsRelativePaths` and `listOrNullOnRelativePathWhichIsNotDotReturnsRelativePaths`

Best way to run the tests as of now.
```
./gradlew androidNativeX64TestBinaries
adb push /path/to/okio/okio/build/bin/androidNativeX64/debugTest/test.kexe /data/local/tmp
rm -r /path/to/okio/build
adb push /path/to/okio /data/local/tmp/okio
adb shell
export OKIO_ROOT=/data/local/tmp/okio
/data/local/tmp/test.kexe
```

MohammedKHC:
After reading the `listOnRelativePathWhichIsNotDotReturnsRelativePaths` code, it seems that the test itself is flaky. It only creates the required `okio.api` test file when one of the following is true:
`isFakeFileSystem || isWrappingJimFileSystem || isWasiFileSystem`

Knowing that, I can say that Okio does work as intended on Android native, which is useful for a bunch of Android tech apps such as Termux or Android IDEs.

MohammedKHC:
@JakeWharton should i edit `listOnRelativePathWhichIsNotDotReturnsRelativePaths` test or just ignore it on Android.
And do you think that the approach of using actual/expect is fine with opendir and this stuff?
Maybe there might be a better way?

MohammedKHC:
@JakeWharton are you free to look at this?

I think there are two ways to make opendir/closedir and this stuff on Android native 
1. Use expect/actual with a typealias and top level functions.
2. Use expect/actual class with a name like DirEntry or something 

And also it would be nice if we could create a new nativeNonAndroid sourceset. Although if we made it then we can't use the option 1, as the platform.posix.DIR is a typealias on Linux and we can't create a typealias to a typealias.
And this would kina complicate the targets hierarchy.

All the tests succeeded expect listOnRelativePathWhichIsNotDotReturnsRelativePaths and listOrNullOnRelativePathWhichIsNotDotReturnsRelativePaths

I can probably fix this tests and edit the github workflow to run them all for androidNative.

What do you think?

Info:
I really need this to be merged as I plan to use it in my 
[Rustroid Rust IDE](https://rustroid.is-a.dev) (Iam writing some native tool that should run on it in Kotlin)

MohammedKHC:
@JakeWharton any update on this one?
I can fix the tests. And run the tests on the Android emulator (on CI), if you want to.

JakeWharton:
Nope. I'm not working right now, so probably won't be looking at this for a month or two.

okio/build.gradle.kts
```diff
@@ -28,6 +28,11 @@ plugins {
  *   |   |-- mingw
  *   |   |   '-- mingwX64
  *   |   '-- unix
+ *   |       |-- androidNative
```
swankjesse: I would have expected this under linux

okio/src/androidNativeMain/kotlin/okio/AndroidNativePosixVariant.kt
```diff
@@ -0,0 +1,53 @@
...
+      size = stat.st_size,
+      createdAtMillis = stat.st_ctim.epochMillis,
+      lastModifiedAtMillis = stat.st_mtim.epochMillis,
+      lastAccessedAtMillis = stat.st_atim.epochMillis,
+    )
+  }
+}
+
+internal actual val DEFFILEMODE: Int get() = 0b110110110 /* octal 666 */
+internal actual typealias DIR = cnames.structs.DIR
```
swankjesse: Interesting that this is different! I’m learning

okio/src/androidNativeMain/kotlin/okio/AndroidNativePosixVariant.kt
```diff
@@ -0,0 +1,53 @@
...
+import kotlinx.cinterop.ptr
+import platform.posix.ENOENT
+import platform.posix.S_IFDIR
+import platform.posix.S_IFMT
+import platform.posix.S_IFREG
+import platform.posix.errno
+import platform.posix.lstat
+import platform.posix.stat
+
+internal actual fun PosixFileSystem.variantMetadataOrNull(path: Path): FileMetadata? {
```
swankjesse, line 29: (This is identical to linux. Possibly building another layer of directories isn’t worth the effort here? I gotta ponder that)

swankjesse (commented):
Looks very promising. Thanks for driving this.

I wanna look into making androidNativeMain a child of linuxMain

okio/build.gradle.kts
MohammedKHC: Actually me too! 
 It's all because androidNative opendir/closedir/dir are not under `platform.posix`.
So you cannot really make anroidNative extend linux.. 
I can probably fix this by using a custom cinterop and defining them as macros on the `platform.posix` package.

Or by introducing an extra sourceset like `linuxNonAndroidNative`
and placing the old opendir code there.

What are your thoughts?

IIRC kotlinx.io solved this by introducing an extra sourceset.

MohammedKHC:
@swankjesse I tried another approach on the latest commit. Can you review it and tell me what do you think?
Now androidNative is a child of linux. and there is no code duplication. But androidNative now contains what I would call a workaround to `platform.posix.DIR`..

MohammedKHC:
Oops looks like the workaround didn't work on CI..
Although I am sure that this worked on my Linux machine.. will investigate more..

okio/src/nativeMain/kotlin/okio/internal/DIR.kt
```diff
@@ -0,0 +1,34 @@
...
+ * POSIX directory APIs (`opendir`, `readdir`, `closedir`) cannot be used
+ * directly.
+ *
+ * These expect declarations provide platform-specific implementations
+ * for all `DIR`-related functionality.
+ */
+internal expect class DIR : CPointed
+internal expect fun opendir(path: String): CPointer<DIR>?
+internal expect fun readdir(dir: CValuesRef<DIR>): CPointer<dirent>?
+internal expect fun closedir(dir: CValuesRef<DIR>): Int
```
MohammedKHC: I believe this is the cleanest **working** approach.

In the previous commit, I attempted to manually introduce a `platform.posix.DIR` typealias, but it only worked locally when building using `./gradlew androidNativeX64TestBinaries`.

okio/src/unixMain/kotlin/okio/internal/PosixDirectory.kt
```diff
@@ -0,0 +1,34 @@
...
+ * POSIX directory APIs (`opendir`, `readdir`, `closedir`) cannot be used
+ * directly.
+ *
+ * [PosixDirectory] provides platform-specific implementation
+ * for `DIR`-related functionality.
+ */
+internal expect class PosixDirectory(path: Path): Closeable {
+  val isInvalid: Boolean
+  fun nextEntry(): CPointer<dirent>?
+}
```
MohammedKHC, line 34: This is the cleanest *working* approach so far.
I probably should’ve marked this PR as a draft from the beginning…

Man, everything feels way more complicated than it needs to be.

Why doesn’t platform.posix just expose DIR on Android?!

JakeWharton: Seems like these should all be `value class`es to avoid the memory allocation. Does anything prevent that?

MohammedKHC: @JakeWharton 
The problem is that the actual implementation needs to store
```kotlin
private val dir = opendir(path.toString())
```
And if you meant that the value should be DIR. Then this is kinda complicated.
as we cannot use expect/actual for DIR in only two source sets. (androidNative, nativeNonAndroid)

that's because this code will not compile under nativeNonAndroid
```kotlin
actual typealias DIR = platform.posix.DIR
```
It will fail with an error cannot create a typealias to a typealias.
As platform.posix.DIR is a typealias on Linux.

we could create a new nativeNonLinux source set that will exclude androidNative and linux.

What do you think?

EDIT: I just realized that we could do this
```kotlin
internal expect value class PosixDirectory(private val dir: COpaquePointer) : Closeable {
  fun nextEntry(): CPointer<dirent>?
  override fun close()
}

internal expect fun openPosixDirectory(path: Path): PosixDirectory?
```
and
```kotlin
internal actual value class PosixDirectory(private val dir: COpaquePointer) : Closeable {
  actual fun nextEntry() = readdir(dir.reinterpret())
  actual override fun close() {
    closedir(dir.reinterpret()) // Ignore errno from closedir.
  }
}

internal actual fun openPosixDirectory(path: Path): PosixDirectory? {
  return opendir(path.toString())?.let(::PosixDirectory)
}
```
Does this seem good?
And also we can make `openPosixDirectory` inline.
Do you think it's worth it?
And btw, I really hate the Kotlin compiler advising me to not use inline on functions..

MohammedKHC:
@swankjesse Is it possible to merge this before the next release?
Also maybe tweak the changelog wording a bit — mentioning Android API 30 might be confusing since Android support is being added with this PR.
https://github.com/square/okio/blob/296831e835185a1c5f7094d15129e03444941155/CHANGELOG.md?plain=1#L19

okio/src/unixMain/kotlin/okio/internal/PosixDirectory.kt
MohammedKHC: @JakeWharton @swankjesse ?

sky130:
Hi @MohammedKHC,
I tried building your branch on my local device, but it failed. I looked at the source code and found that functions on the arm32 architecture accept int type, while functions on the arm64 architecture accept long type. Do you have time to investigate this issue? If not, I will try to resolve it later.

---
source code

image

---

``` log
> Task :okio:compileKotlinAndroidNativeArm32 FAILED
e: file:///Users/***/Desktop/projects/okio/okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt:47:7 Argument type mismatch: actual type is 'Long', but 'Int' was expected.
e: file:///Users/***/Desktop/projects/okio/okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt:54:9 Operator '==' cannot be applied to 'Int' and 'Long'.

Execution failed for task ':okio:compileKotlinAndroidNativeArm32'.
> Compilation finished with errors
```

MohammedKHC:
Hi! @sky130 
Thanks for catching this! I fixed it just now. https://github.com/square/okio/pull/1739/commits/dfef41683a7a23c63bb88d277323a2a873831118
I forgot to compile to arm32 after the new statx implementation got merged, sorry!

MohammedKHC:
@JakeWharton can this be merged anytime soon?
I almost forgot about it.

sky130:
Hi @MohammedKHC ,

I meet another problem now. The app crashes with a **SIGSYS signal** on Android API 26(AVD).

Stack trace points to `okio.PosixFileSystem.variantMetadataOrNull()` calling the `statx` syscall. On  Android API 26, `statx` is blocked by `seccomp` and the kernel **directly sends SIGSYS** to the process. Execution never reaches the fallback `lstat` path because the signal terminates the app immediately.

Do you have time to look into this issue? If you don't have time, I'll try to fix it.

```
Fatal signal 31 (SIGSYS), code 1 in tid 29348 (package-name)
(kfun:okio#variantMetadataOrNull__at__okio.PosixFileSystem(okio.Path){}okio.FileMetadata?+1136)
(kfun:okio.PosixFileSystem#metadataOrNull(okio.Path){}okio.FileMetadata?+88)
(kfun:okio.FileSystem#metadataOrNull(okio.Path){}okio.FileMetadata?-trampoline+64)
(kfun:okio.internal#commonExists__at__okio.FileSystem(okio.Path){}kotlin.Boolean+132)
(kfun:okio.FileSystem#exists(okio.Path){}kotlin.Boolean+80)
(kfun:okio.internal#commonCreateDirectories__at__okio.FileSystem(okio.Path;kotlin.Boolean){}+536)
(kfun:okio.FileSystem#createDirectories(okio.Path;kotlin.Boolean){}+108)
(kfun:okio.FileSystem#createDirectories$default(okio.Path;kotlin.Boolean;kotlin.Int){}+172)
```

swankjesse (commented):
Sorry for sitting on this for such a long time.

I think this is good stuff. My only request is to change `PosixDirectory` from an `expect class` to an interface, that'll have different implementations on each platform.

ALSO – I’d prefer to be able to run some kind of androidNative test cases in CI. Any idea what’s involved in that?

JakeWharton:
Build test kexe, start emulator, adb push, adb shell to run. It should work like a normal test since the output will be forwarded and modern emulators (finally) propagate the exit code.

MohammedKHC:
So I ran tests on my Android phone and all of them succeeded..
However the CI seems to be stuck on
`okio.CommonBufferedSourceTest_OneByteAtATimeSink.indexOfByteStringAcrossSegmentBoundaries`
any idea why is that???

Edit
looks like its just stuck for some reason on different tests..
https://github.com/lysine-dev/okio/actions/runs/31183239226/job/92881400062
https://github.com/lysine-dev/okio/actions/runs/31181796110/job/92876601036
https://github.com/lysine-dev/okio/actions/runs/31181106776/job/92874319704
https://github.com/lysine-dev/okio/actions/runs/31180989417/job/92873942140

does this mean that the emulator can't handle all those tests?

PR #1770, opened 2026-02-04
Ensure compatibility with older Kotlin consumers

omarismail94:
We delegate the build configuration to Tapmoc and set the Kotlin version to 2.1 as the library uses APIs from this version that do not exist in 2.0.

One caveat:
`project.configureJavaCompatibility(8)` is not used as the code would compile with JDK 8. This will fail as `RealBufferedSource.kt` uses a Java API not available in Java 8: `InputStream.transferTo`. To use `project.configureJavaCompatibility`, the min supported Java version would need to be bumped to 11

TESTED: ./gradlew clean check . I also published to local repository and checked the .module files for the  JVM and Android artifacts and the kotlin-stdlib was correctly set to 2.1.21


build-support/src/main/kotlin/BuildSupport.kt
```diff
@@ -13,11 +13,119 @@
...
+
+  private fun Project.configureCommonKotlin() {
+    tasks.withType(KotlinCompilationTask::class.java).configureEach {
+      compilerOptions {
+        freeCompilerArgs.addAll("-Xexpect-actual-classes")
+        apiVersion.set(KotlinVersion.KOTLIN_2_1)
+        languageVersion.set(KotlinVersion.KOTLIN_2_1)
+      }
+    }
+    afterEvaluate {
```
swankjesse: I think we should be able to do _withType_ to avoid `afterEvaluate()`.

JakeWharton:
Need to use Tapmoc plugin. Much easier.

omarismail94:
the tapmoc plugin does not work with KMP.  I tried it locally, and it adds both the the KGP `kotlin-stdlib`  and the one set for compatibility (2.1.21) in the module file of the KMP targets:

```
"dependencies": [
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib",
          "version": {
            "requires": "2.2.21"
          }
        },
        {
          "group": "org.jetbrains.kotlin",
          "module": "kotlin-stdlib",
          "version": {
            "requires": "2.1.21"
          }
        }
```


For now, Ill add the plugin to the samples project (which isn't published), but keep the current behaviour I have

build-support/src/main/kotlin/BuildSupport.kt
omarismail94: fixed

omarismail94:
@JakeWharton @swankjesse  tests now pass. 

Jake - if you think tapmoc can still work, I can open it as a feature request issue against the repo. For now, Im hoping itd be fine to merge as is

omarismail94:
@JakeWharton  Actually, ignore my previous comment. I migrated to Tapmoc, and wow, the code is so much simpler! 

One caveat:
`project.configureJavaCompatibility(8)` is not used as the code would compile with JDK 8. This will fail as `RealBufferedSource.kt` uses a Java API not available in Java 8:` InputStream.transferTo`. To use `project.configureJavaCompatibility`, the min supported Java version would need to be bumped to 11

Issue #1755, opened 2025-12-19
`FileMetadata.creationTime` does not return the file creationTime on unix.


MohammadKHC:
`stat.ctime` refers to **change time**, not creation time.

In fact, POSIX does not provide a way to retrieve the creation time.
However, since Linux 4.11, there is a new `statx` system call, which provides
`stx_btime` as the creation (birth) time.

The JVM’s `UnixFileAttributes.creationTime()` already checks whether the birth
time is available and, if it is, returns it. Otherwise, it falls back to
`mtime`, not `ctime`.

See:
https://github.com/openjdk/jdk/blob/master/src/java.base/unix/classes/sun/nio/fs/UnixFileAttributes.java#L176

Note: It seems like macOS already provides a true file creation time via `st_birthtimespec`
in the original `stat` struct.

swankjesse:
Can fix!

MohammedKHC:
Fixed by #1772

PR #1772, opened 2026-02-05
Switch to FileMetadata to use statx instead of stat on Linux and Apple platforms

swankjesse:
https://github.com/square/okio/issues/1755

### DIFF
--- a/okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
+++ b/okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
@@ -32,10 +32,16 @@ import okio.internal.linux.statx_timestamp
 import platform.posix.ENOENT
 import platform.posix.ENOSYS
 import platform.posix.errno
+import platform.posix.lstat
+import platform.posix.stat
 import platform.posix.syscall
 
+/**
+ * Prefer `statx()` if it's available. Fall back to `stat()` which doesn't have a field for
+ * `createdAt`.
+ */
 internal actual fun PosixFileSystem.variantMetadataOrNull(path: Path): FileMetadata? {
-  return memScoped {
+  memScoped {
     val statx = alloc<statx>()
     val result = syscall(
       __NR_statx.toLong(),
@@ -45,20 +51,40 @@ internal actual fun PosixFileSystem.variantMetadataOrNull(path: Path): FileMetad
       STATX_BASIC_STATS or STATX_BTIME,
       statx.ptr,
     )
-    if (result != 0L) {
-      if (errno == ENOSYS) return null
-      if (errno == ENOENT) return null
-      throw errnoToIOException(errno)
+    if (result == 0L) {
+      return FileMetadata(
+        isRegularFile = statx.stx_mode.toInt() and S_IFMT == S_IFREG,
+        isDirectory = statx.stx_mode.toInt() and S_IFMT == S_IFDIR,
+        symlinkTarget = symlinkTarget(statx.stx_mode.toInt(), path),
+        size = statx.stx_size.toLong(),
+        createdAtMillis = when {
+          statx.stx_mask and STATX_BTIME != 0U -> statx.stx_btime.epochMillis
+          else -> statx.stx_mtime.epochMillis
+        },
+        lastModifiedAtMillis = statx.stx_mtime.epochMillis,
+        lastAccessedAtMillis = statx.stx_atime.epochMillis,
+      )
     }
-    return@memScoped FileMetadata(
-      isRegularFile = statx.stx_mode.toInt() and S_IFMT == S_IFREG,
-      isDirectory = statx.stx_mode.toInt() and S_IFMT == S_IFDIR,
-      symlinkTarget = symlinkTarget(statx.stx_mode.toInt(), path),
-      size = statx.stx_size.toLong(),
-      createdAtMillis = statx.stx_btime.epochMillis,
-      lastModifiedAtMillis = statx.stx_mtime.epochMillis,
-      lastAccessedAtMillis = statx.stx_atime.epochMillis,
-    )
+
+    // Recover if statx() isn't available. It first appeared in Linux in 4.11 (2017-04-30) and
+    // Android in API 30 (2020-09-08).
+    if (errno == ENOSYS) {
+      val stat = alloc<stat>()
+      if (lstat(path.toString(), stat.ptr) == 0) {
+        return FileMetadata(
+          isRegularFile = stat.st_mode.toInt() and S_IFMT == S_IFREG,
+          isDirectory = stat.st_mode.toInt() and S_IFMT == S_IFDIR,
+          symlinkTarget = symlinkTarget(stat.st_mode.toInt(), path),
+          size = stat.st_size,
+          createdAtMillis = stat.st_mtim.epochMillis,
+          lastModifiedAtMillis = stat.st_mtim.epochMillis,
+          lastAccessedAtMillis = stat.st_atim.epochMillis,
+        )
+      }
+    }
+
+    if (errno == ENOENT) return null
+    throw errnoToIOException(errno)
   }
 }
 


MohammedKHC:
This is good!
But also I think that we may use statx call on Linux.
It's only downside is that we need to use custom cinterop. and check if it's available at runtime..

Please refer to this code
https://github.com/MohammadKHC/mokio/blob/main/mokio/src/linuxMain/kotlin/metadata/FileMetadata.linux.kt#L40
and
https://github.com/MohammadKHC/mokio/blob/main/mokio/src/nativeInterop/linux/statx.def

And if we couldn't use statx then we should really let the creation time be mtime.. as in some cases (ex, modifying the file permissions)  will make ctime newer than mtime!

okio-testing-support/build.gradle.kts
```diff
@@ -67,3 +70,34 @@ kotlin {
...
 }
+
+/*
+ * AbstractFileSystemTest expects a file with specific metadata:
+ *
+ *   path: build/AbstractFileSystemTestFiles/metadata.txt
+ *   createdAt: 2026-01-01T01:01:01Z
+ *   lastModifiedAt: 2026-02-02T02:02:02Z
+ *
+ * It would be nicer to do this in the test itself with exec(), but we don't yet have a handy
```
MohammedKHC, line 86: What if okio provided an api to set file attributes?
Maybe something like this
https://github.com/MohammadKHC/mokio/blob/main/mokio/src/commonMain/kotlin/metadata/FileAttribute.kt

Although iam sure that you can even do better than it.
Iam really bad at writing documentation for things.  I usually just leave the code as is..

swankjesse:
@MohammadKHC your mokio project is rad!

I hope you don’t mind I’d like to copy your `Process` class (and the code that supports it) into `okio-testing-support`. I want to do processes in our tests! I would like to later promote that code into the main production artifact, but I want to start small by running it in our tests. There’s a lot of little design decisions in these things and I don’t yet have an intuition on how it should work. Forking processes is forbidden on iOS!

MohammedKHC:
@swankjesse

I absolutely don’t mind — and I’m glad you liked Mokio.

I actually built it hoping some parts might eventually be useful for Okio, so feel free to copy the Process class and any supporting code.

There may be some imperfect design decisions, but it works for my use case.

Mokio doesn’t support iOS for the same reason you mentioned: forking processes isn’t supported there. I considered a custom source set or stubs, but I decided it wasn’t worth the extra complexity for my use case.

Happy to see where this goes if it’s promoted later.

Thanks!

okio/src/unixMain/kotlin/okio/UnixPosixVariant.kt
```diff
@@ -188,11 +187,11 @@ internal fun variantPwrite(
 
 @OptIn(UnsafeNumber::class)
 internal val timespec.epochMillis: Long
-  get() = tv_sec * 1000L + tv_sec / 1_000_000L
+  get() = tv_sec * 1000L + tv_nsec / 1_000_000L
```
swankjesse, line 190: Longstanding severe bug!

okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
```diff
@@ -15,32 +15,53 @@
...
+    val result = syscall(
+      __NR_statx.toLong(),
+      AT_FDCWD,
+      path.toString(),
+      AT_SYMLINK_NOFOLLOW,
+      STATX_BASIC_STATS or STATX_BTIME,
+      statx.ptr,
+    )
+    if (result != 0L) {
+      if (errno == ENOSYS) return null
```
MohammedKHC: I think that we really should fallback to `lstat` when `statx` is not available.
This is actually the case on Android Native pre API 30.
https://android.googlesource.com/platform/bionic/+/master/libc/include/sys/stat.h#334

okio/src/linuxMain/headers/okio_statx.h
```diff
@@ -0,0 +1,6 @@
+/*
+ * Symbols to call statx() via syscall().
+ */
+extern int __NR_statx;
+extern int AT_FDCWD;
+extern int AT_SYMLINK_NOFOLLOW;
```
MohammedKHC: These are macros. So this would not work..
See https://elixir.bootlin.com/linux/v4.14.13/source/include/uapi/linux/fcntl.h#L78

you can just copy the macros and paste them here.

okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
```diff
@@ -15,32 +15,53 @@
...
-      symlinkTarget = symlinkTarget(stat.st_mode.toInt(), path),
-      size = stat.st_size,
-      createdAtMillis = stat.st_ctim.epochMillis,
-      lastModifiedAtMillis = stat.st_mtim.epochMillis,
-      lastAccessedAtMillis = stat.st_atim.epochMillis,
+      isRegularFile = statx.stx_mode.toInt() and S_IFMT == S_IFREG,
+      isDirectory = statx.stx_mode.toInt() and S_IFMT == S_IFDIR,
+      symlinkTarget = symlinkTarget(statx.stx_mode.toInt(), path),
+      size = statx.stx_size.toLong(),
+      createdAtMillis = statx.stx_btime.epochMillis,
```
MohammedKHC: Please note that `stx_btime` may be a dummy value if `statx` is supported by the system and the fs doesn't support it.

we should check if the returned `stx_mode` contains `STATX_BTIME`
and if it doesn't contain it we should fallback to `lastModifed` time.

from `statx` man page
> If a filesystem does not support a field or if it has an
> unrepresentable value (for instance, a file with an exotic type),
> then the mask bit corresponding to that field will be cleared in
> stx_mask even if the user asked for it and a dummy value will be
> filled in for compatibility purposes if one is available (e.g., a
> dummy UID and GID may be specified to mount under some
> circumstances).

swankjesse: Nice.

okio/src/linuxMain/headers/okio_statx.h
swankjesse: Any idea what I can paste for `__NR_statx` ?

swankjesse: I think it’s this:
https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/unistd.h#L699C1-L699C23

MohammedKHC: I think hardcoding the syscall number is a bad idea.. because it differs per CPU architecture.

We should probably include the `sys/syscall.h` file and use `SYS_statx` instead.

From `unistd.h`

> This file contains the system call numbers, based on the
>  layout of the x86-64 architecture, which embeds the
>  pointer to the syscall in the table.

swankjesse: After some grinding with different options, I think the least bad option is to hardcode it for each of our 2 supported platforms. As far as I can tell this yields the exact same constant I’d get by pulling it from `sys/syscall.h` at buildtime, but this way I don’t need to get GitHub actions to build on every architecture we support.

MohammedKHC: Okay! However I think that we should support all the main 4 archs.
arm64, arm32, x86_64, x86.
this is to be able to support android native in the feature.

This code should do it.
```c
#ifdef __x86_64__
#define __NR_statx 332
#elif defined(__arm__)
#define __NR_statx 397
#elif defined(__aarch64__)
#define __NR_statx 291
#elif defined(__i386__)
#define __NR_statx 383
#else
#error "Unknown arch!"
#endif
```

The syscall numbers are copied from here
https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md

okio/src/linuxMain/headers/okio_statx.h
```diff
@@ -1,6 +1,6 @@
 /*
  * Symbols to call statx() via syscall().
  */
-extern int __NR_statx;
-extern int AT_FDCWD;
-extern int AT_SYMLINK_NOFOLLOW;
+#define __NR_statx 332
```
MohammedKHC, line 8:
```suggestion
#ifdef __x86_64__
#define __NR_statx 332
#elif defined(__arm__)
#define __NR_statx 397
#elif defined(__aarch64__)
#define __NR_statx 291
#elif defined(__i386__)
#define __NR_statx 383
#else
#error "Unknown arch!"
#endif
```

source: https://chromium.googlesource.com/chromiumos/docs/+/master/constants/syscalls.md

okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
```diff
@@ -15,32 +15,79 @@
...
+    // Android in API 30 (2020-09-08).
+    if (errno == ENOSYS) {
+      val stat = alloc<stat>()
+      if (lstat(path.toString(), stat.ptr) == 0) {
+        return FileMetadata(
+          isRegularFile = stat.st_mode.toInt() and S_IFMT == S_IFREG,
+          isDirectory = stat.st_mode.toInt() and S_IFMT == S_IFDIR,
+          symlinkTarget = symlinkTarget(stat.st_mode.toInt(), path),
+          size = stat.st_size,
+          createdAtMillis = stat.st_mtim.epochMillis,
```
swankjesse, line 79: behavior change from `st_ctim`

okio-testing-support/build.gradle.kts
swankjesse: I don’t believe that it’s an API many people want?

okio/src/linuxMain/kotlin/okio/LinuxPosixVariant.kt
swankjesse: Done.

okio/src/linuxMain/headers/okio_statx.h
swankjesse: Perfecto. Done.

okio-testing-support/build.gradle.kts
swankjesse: Some background [here](https://publicobject.com/2020/10/06/files/), where I argue that files are less frequently UI and therefore setting metadata is not as useful as it once was.

There is definitely a use for it, but it’s difficult to do well. For example, I don’t believe cloud file systems do a particularly good job of preserving metadata!

MohammedKHC: I read your article, and I mostly agree with it.

It’s not just cloud file systems that don’t preserve file metadata—Git doesn’t either. Aside from whether a file is executable, Git ignores file metadata entirely.

https://softwareengineering.stackexchange.com/questions/350403/why-doesnt-git-set-the-file-time
https://superuser.com/a/1422661

That said, I think that we do need to support at least one piece of metadata: whether a file is executable.
This is very common on Unix systems.

Setting file creation or last-modified times only really makes sense for things like restoring family photo dates in the gallery 😄

And other attributes don’t seem necessary unless you’re building something very specialized.

However they are kinda nice to have in a filesystem API!

swankjesse: Yeah, they are nice to have in a FileSystem API. I certainly prefer metadata over access control!

okio/src/unixMain/kotlin/okio/UnixPosixVariant.kt
swankjesse: I assume we got away with it cause it's usually zero? Otherwise we wouldn't have passed our tests.

okio-testing-support/src/commonMain/kotlin/okio/TestingCommon.kt
```diff
@@ -60,6 +68,9 @@ val FileMetadata.lastAccessedAt: Instant?
     return Instant.fromEpochMilliseconds(lastAccessedAt)
   }
 
+fun fromIso8601String(iso8601String: String): Instant =
+  Instant.fromEpochMilliseconds(Instant.parse(iso8601String).toEpochMilliseconds())
```
JakeWharton, line 70: Why the round trip through millis after parsing?

MohammedKHC: It's probably to ignore some nanosecond precision..
Although I am not sure if it's intentional or not..

Edit:  after checking the `assertInRange` code I think that this is redundant.
Sorry for replying before checking the code 😶‍🌫️

okio-testing-support/src/commonMain/kotlin/okio/TestingCommon.kt
```diff
@@ -42,6 +42,14 @@ expect fun isBrowser(): Boolean
...
 expect fun isWasm(): Boolean
 
+/**
+ * Returns true if the host file system probably exposes metadata like file creation time.
+ *
+ * The file system that GitHub actions gives us doesn't do anything when we `touch` a file.
+ */
+val fileSystemHasGoodMetadata: Boolean
+  get() = getEnv("GITHUB_WORKSPACE") == null
+
```
MohammedKHC, line 50: After reviewing the code again, I don’t think this is a GitHub issue.

GNU touch itself does not modify the creation time (and on Linux there’s no way to do that anyway).

Relevant code:
https://github.com/coreutils/coreutils/blob/3d35e82b9b0460769c1966b1ef8acc0b5e5c8348/src/touch.c#L138-L160
https://github.com/coreutils/coreutils/blob/3d35e82b9b0460769c1966b1ef8acc0b5e5c8348/src/touch.c#L218

I’m not sure about macOS and Windows, though.

okio-testing-support/src/commonMain/kotlin/okio/TestingCommon.kt
swankjesse: Yep. Just a bug. Fixed!

swankjesse: (introduced by deleting our commonMain Instant for Wasm)

okio-testing-support/src/commonMain/kotlin/okio/TestingCommon.kt
swankjesse: Ahh, it’s a Linux issue, not a GitHub issue. That makes sense.

swankjesse: I was able to do validate our code worked successfully running Linux in Docker on my Mac, but that would have relied on my Mac’s `touch` command.

PR #1774, opened 2026-02-09
Fix result of an 'errnoToIOException' call is not thrown.

MohammedKHC:


JakeWharton:
Seems like we should change that function to throw directly and return `Nothing` to avoid things like this from happening.

JakeWharton (approved):

swankjesse:
Tests not catching this is annoying. Lemme add that.

swankjesse (commented):
Ahh, this was passing 'cause the code still throws when it falls through. 

https://github.com/square/okio/pull/1787

PR #1780, opened 2026-02-12
Fix `FileHandle` thread safety on native platforms

### LINKED ISSUE #1778: FileHandle is not thread-safe on native due to no-op Lock implementation.
`FileHandle` uses `Lock.withLock()` to protect `closed` and `openStreamCount`, but on native platforms the Lock implementation does nothing:

https://github.com/square/okio/blob/79aa26755c77df8c4d0233926c7308fd353ad697/okio/src/nonJvmMain/kotlin/okio/NonJvmPlatform.kt#L42-L48

This means `FileHandle` has race conditions when used from multiple threads.

- `openStreamCount` can be corrupted
- Resource leaks possible when `close()` races with `source()`/`sink()`
- Methods check `!closed` but state can change between check and use

Can we use `AtomicBoolean` and `AtomicInt` instead of `Lock`?
Or implement it on native using `pthread_mutex`?

MohammedKHC:
Replace no-op Lock with `AtomicBoolean`/`AtomicInt` for proper synchronization.

Is it fine to delete `newLock()` completely? `Lock` is public but `newLock()` is internal, and `FileHandle` was the only place using it.

Closes #1778

### DIFF
--- a/okio/src/commonMain/kotlin/okio/FileHandle.kt
+++ b/okio/src/commonMain/kotlin/okio/FileHandle.kt
@@ -289,6 +289,7 @@ abstract class FileHandle(
       closed = true
       if (openStreamCount != 0) return
     }
+    lock.destroy()
     protectedClose()
   }
 
--- a/okio/src/commonMain/kotlin/okio/FileHandle.kt
+++ b/okio/src/commonMain/kotlin/okio/FileHandle.kt
@@ -15,12 +15,6 @@
  */
 package okio
 
-import kotlin.concurrent.atomics.AtomicBoolean
-import kotlin.concurrent.atomics.AtomicInt
-import kotlin.concurrent.atomics.ExperimentalAtomicApi
-import kotlin.concurrent.atomics.decrementAndFetch
-import kotlin.concurrent.atomics.plusAssign
-
 /**
  * An open file for reading and writing; using either streaming and random access.
  *
@@ -38,7 +32,6 @@ import kotlin.concurrent.atomics.plusAssign
  * File handles may be used by multiple threads concurrently. But the individual sources and sinks
  * produced by a file handle are not safe for concurrent use.
  */
-@OptIn(ExperimentalAtomicApi::class)
 abstract class FileHandle(
   /**
    * True if this handle supports both reading and writing. If this is false all write operations
@@ -51,13 +44,15 @@ abstract class FileHandle(
    * True once the file handle is closed. Resources should be released with [protectedClose] once
    * this is true and [openStreamCount] is 0.
    */
-  private var closed = AtomicBoolean(false)
+  private var closed = false
 
   /**
    * Reference count of the number of open sources and sinks on this file handle. Resources should
    * be released with [protectedClose] once this is 0 and [closed] is true.
    */
-  private var openStreamCount = AtomicInt(0)
+  private var openStreamCount = 0
+
+  val lock: Lock = newLock()
 
   /**
    * Reads at least 1, and up to [byteCount] bytes from this starting at [fileOffset] and copies
@@ -71,7 +66,9 @@ abstract class FileHandle(
     arrayOffset: Int,
     byteCount: Int,
   ): Int {
-    check(!closed.load()) { "closed" }
+    lock.withLock {
+      check(!closed) { "closed" }
+    }
     return protectedRead(fileOffset, array, arrayOffset, byteCount)
   }
 
@@ -81,7 +78,9 @@ abstract class FileHandle(
    */
   @Throws(IOException::class)
   fun read(fileOffset: Long, sink: Buffer, byteCount: Long): Long {
-    check(!closed.load()) { "closed" }
+    lock.withLock {
+      check(!closed) { "closed" }
+    }
     return readNoCloseCheck(fileOffset, sink, byteCount)
   }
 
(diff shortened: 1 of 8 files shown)


MohammedKHC:
Or should we just implement the `Lock` for native?

I think that Atomics should be more efficient for this use case.

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -32,6 +38,7 @@ package okio
  * File handles may be used by multiple threads concurrently. But the individual sources and sinks
  * produced by a file handle are not safe for concurrent use.
  */
+@OptIn(ExperimentalAtomicApi::class)
```
JakeWharton: We can't do this in the main artifact

okio/src/commonMain/kotlin/okio/FileHandle.kt
```diff
@@ -218,10 +207,8 @@ abstract class FileHandle(
   @Throws(IOException::class)
   fun sink(fileOffset: Long = 0L): Sink {
     check(readWrite) { "file handle is read-only" }
-    lock.withLock {
-      check(!closed) { "closed" }
-      openStreamCount++
-    }
+    check(!closed.load()) { "closed" }
+    openStreamCount += 1
```
JakeWharton: This creates a race:
* Thread 1: Close is false
* Thread 2: Close is flipped to true
* Thread 2: Stream count is checked to be 0
* Thread 1: Stream count is incremented to 1

MohammedKHC:
@JakeWharton Okay! instead of using atomics,
I implemented `Lock` for Unix and Windows.

Can you review it again?

okio/src/jsMain/kotlin/okio/JsPlatform.kt
```diff
@@ -50,3 +53,19 @@ private val path: dynamic
 
 internal val tmpdir: String
   get() = os?.tmpdir() as? String ?: "/tmp"
+
+actual class Lock {
```
JakeWharton: Would be curious if we could do `actual typealias Lock = Unit` here and in Wasm.

okio/src/unixMain/kotlin/okio/UnixPlatform.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+import kotlin.experimental.ExperimentalNativeApi
+import kotlin.native.ref.createCleaner
+import kotlinx.cinterop.alloc
+import kotlinx.cinterop.free
+import kotlinx.cinterop.nativeHeap
+import kotlinx.cinterop.ptr
+import platform.posix.*
```
JakeWharton: No star imports, please

okio/src/mingwX64Main/kotlin/okio/WindowsPlatform.kt
```diff
@@ -0,0 +1,55 @@
...
+
+actual class Lock {
+  val mutex = CreateMutexA(
+    null,
+    0,
+    null
+  ) ?: throw lastErrorToIOException()
+
+  @Suppress("unused")
+  @OptIn(ExperimentalNativeApi::class)
```
JakeWharton: Unfortunately we can't depend on unstable API

MohammedKHC: Hmm, but there’s no real alternative here.

`ByteString` already uses `@OptIn(ExperimentalNativeApi::class)` on Apple platforms for some method. Most native code already opts into `@OptIn(UnsafeNumber::class)`, and we also opt into `ExperimentalForeignApi` at the module level:

https://github.com/square/okio/blob/fc7aecb7f6f7a123f2024ab6397da04311546bf2/okio/build.gradle.kts#L61
https://github.com/square/okio/blob/fc7aecb7f6f7a123f2024ab6397da04311546bf2/okio/src/appleMain/kotlin/okio/ByteString.kt#L195

Given that, adding another opt-in here feels consistent with the existing approach.

What are your thoughts?

okio/src/jsMain/kotlin/okio/JsPlatform.kt
MohammedKHC: Will do!

okio/src/unixMain/kotlin/okio/UnixPlatform.kt
MohammedKHC: Okay!

okio/src/mingwX64Main/kotlin/okio/WindowsPlatform.kt
JakeWharton: Unsafe number has nothing to do with stability. It indicates that the size of the underlying integer varies across platforms (usually then requiring explicit conversion).

That usage of experimental native API is for `CName` only, not an actual runtime API.

As to the foreign API, we have no choice if we want to talk to anything other than pure Kotlin which we very much need to.

MohammedKHC: That makes sense. So what would you suggest we do here instead?
If opting in isn’t the right direction, what’s the alternative you’d prefer?

JakeWharton: We could make `Lock` an actual `Closeable` resource and move the responsibility of freeing to the caller for now, assuming the callers all have an appropriate lifecycle in which we can shut down the lock completely.

MohammedKHC: Done, Implemented `Closeable` for native lock.  And added an extension method `Lock.destroy`.
Does this seem appropriate?

PR #1799, opened 2026-04-10
Add constant-time `equals` for `ByteString`

### LINKED ISSUE #1797: Add `ByteString.equals(other: ByteString, constantTime: Boolean)` overload
Specifically for timing-safe cryptographic comparison of things like hashes.

Java has [`MessageDigest.isEqual(byte[], byte[])`](https://docs.oracle.com/en/java/javase/26/docs/api/java.base/java/security/MessageDigest.html#isEqual(byte%5B%5D,byte%5B%5D)). This is that but for `ByteString`, and on all targets.

This would xor all bytes (after a quick length check) no matter what so the number of CPU operations is constant (modulo the JVM's internal shenanigans).

For testing, we think you could do something like create a _very_ large random `ByteString` and then compare it against an exact copy and a copy with the first byte different. Do a few iterations of that in a loop (say 100) and then you should see a statistical probability that `equals` short-circuits and `equals(.., constantTime=true)` does not.

JakeWharton:
This is a best-effort comparison for use in cryptographic operations so as to not leak timing information.

Closes #1797

### DIFF
--- a/okio/api/okio.api
+++ b/okio/api/okio.api
@@ -310,6 +310,9 @@ public class okio/ByteString : java/io/Serializable, java/lang/Comparable {
 	public final fun endsWith (Lokio/ByteString;)Z
 	public final fun endsWith ([B)Z
 	public fun equals (Ljava/lang/Object;)Z
+	public final fun equals (Lokio/ByteString;)Z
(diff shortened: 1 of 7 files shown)


JakeWharton:
Wow BRUTAL variance on CI

Issue #1460, opened 2024-03-23
Issue with folders named like 'X: Something'


Ribesg:
Okio behaves weirdly with folders named like `X: Something` on at least MacOS Arm64 and Linux X64.

You can easily reproduce the issue in a Gradle project with native targets:
```kotlin
fun main() {
    FileSystem.SYSTEM.run {
        createDirectories("build/tmp/Working Folder".toPath())
        createDirectories("build/tmp/X: Broken Folder".toPath())
        list("build/tmp".toPath()).forEach(::println)
    }
}
```
```
tmp$ ./gradlew clean runDebugExecutableMacosArm64

> Task :runDebugExecutableMacosArm64
X: Broken Folder
build/tmp/Working Folder

BUILD SUCCESSFUL in 3s
10 actionable tasks: 6 executed, 4 up-to-date
```

swankjesse:
Oooh that’s for reporting this. I made a design decision to make path processing platform-independent, which is what’s causing grief here.

We’re incorrectly interpreting this as a Windows drive letter followed by a path name, rather than a single path name where the 2nd character is a colon. Ugh.

I’ll see what we can do to fix!

PR #1802, opened 2026-04-20
Better Windows' volume detection

oldergod:
Windows accepts both `/` and `\` !

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/Path.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/Path.kt
@@ -91,8 +91,10 @@ private fun Path.rootLength(): Int {
     return 1
   }
 
-  // Look for a root like `C:\`.
-  if (bytes.size > 2 && bytes[1] == ':'.code.toByte() && bytes[2] == '\\'.code.toByte()) {
+  // Look for a root like `C:\` or `C:/`.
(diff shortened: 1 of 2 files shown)


okio/src/commonMain/kotlin/okio/internal/Path.kt
```diff
@@ -113,7 +115,6 @@ internal inline fun Path.commonIsRelative(): Boolean {
 
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun Path.commonVolumeLetter(): Char? {
-  if (bytes.indexOf(SLASH) != -1) return null
```
oldergod, line 116: This was preventing forward slashes from being accepted

JakeWharton:
I'm not comfortable deciding if this is valid or not. Deferring to Jesse.

swankjesse (commented):
What’s the motivation here?

Unfortunately we’re in a small bind here because our API behaves the same regardless of what the host platform is, and treating `C:/Windows` as a drive letter also means we further aggravate this bug: https://github.com/square/okio/issues/1460

oldergod:
This came up through this ticket on Wire https://github.com/square/wire/pull/3215
If Windows does accept both slashes, I thought it'd make sense to support both

PR #1804, opened 2026-04-29
Add padding: Boolean = true to base64 functions

### LINKED ISSUE #1798: Add `padding: Boolean = true` to `base64` functions
So you can turn padding off. Sometimes you don't need it!

mattrob33:
Closes #1798

### DIFF
--- a/okio/src/commonMain/kotlin/okio/Base64.kt
+++ b/okio/src/commonMain/kotlin/okio/Base64.kt
@@ -113,8 +113,15 @@ internal fun String.decodeBase64ToArray(): ByteArray? {
   return out.copyOf(outCount)
 }
 
-internal fun ByteArray.encodeBase64(map: ByteArray = BASE64): String {
-  val length = (size + 2) / 3 * 4
+internal fun ByteArray.encodeBase64(
+  includePadding: Boolean,
+  map: ByteArray = BASE64,
+): String {
+  val length = if (includePadding) {
+    (size + 2) / 3 * 4
+  } else {
+    (size * 4 + 2) / 3
+  }
   val out = ByteArray(length)
   var index = 0
   val end = size - size % 3
@@ -133,8 +140,10 @@ internal fun ByteArray.encodeBase64(map: ByteArray = BASE64): String {
       val b0 = this[i].toInt()
       out[index++] = map[b0 and 0xff shr 2]
       out[index++] = map[b0 and 0x03 shl 4]
-      out[index++] = '='.code.toByte()
-      out[index] = '='.code.toByte()
+      if (includePadding) {
+        out[index++] = '='.code.toByte()
+        out[index] = '='.code.toByte()
+      }
     }
     2 -> {
       val b0 = this[i++].toInt()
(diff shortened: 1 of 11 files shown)


JakeWharton:
Are you able to add tests?

mattrob33:
Binary compatibility check is failing due to default values for args in expect/actual functions. I'm happy to make changes if needed but would need some guidance on the desired shape of this.

```
  -	public fun base64 ()Ljava/lang/String;
  +	public final fun base64 ()Ljava/lang/String;
  +	public fun base64 (Z)Ljava/lang/String;
  +	public static synthetic fun base64$default (Lokio/ByteString;ZILjava/lang/Object;)Ljava/lang/String;

  -	public fun base64Url ()Ljava/lang/String;
  +	public fun base64Url (Z)Ljava/lang/String;
  +	public static synthetic fun base64Url$default (Lokio/ByteString;ZILjava/lang/Object;)Ljava/lang/String;
```

okio/src/jvmMain/kotlin/okio/ByteString.kt
```diff
@@ -105,7 +106,7 @@ internal actual constructor(
     }
   }
 
-  actual open fun base64Url() = commonBase64Url()
+  actual open fun base64Url(includePadding: Boolean) = commonBase64Url(includePadding = includePadding)
```
MohammedKHC, line 110: Seems like you missed `@JvmOverloads` here.
Shouldn't adding it make the api backwards compatible?

JakeWharton:
That is a compatible change, and is fine to commit.

okio/src/jvmMain/kotlin/okio/ByteString.kt
mattrob33: Yes missed that one, just added

okio/src/commonMain/kotlin/okio/Base64.kt
```diff
@@ -113,8 +113,15 @@ internal fun String.decodeBase64ToArray(): ByteArray? {
...
-internal fun ByteArray.encodeBase64(map: ByteArray = BASE64): String {
-  val length = (size + 2) / 3 * 4
+internal fun ByteArray.encodeBase64(
+  includePadding: Boolean,
+  map: ByteArray = BASE64,
+): String {
+  val length = if (includePadding) {
+    (size + 2) / 3 * 4
+  } else {
+    (size * 4 + 2) / 3
```
swankjesse, line 123: checks out.

1 byte –> 2 base64 chars
2 bytes –> 3 base64 chars
3 bytes –> 4 base64 chars
4 bytes –> 6 base64 chars
...

swankjesse (approved):

swankjesse (commented):
I was expecting a new apiDump file

swankjesse:
Cherry-picking [this commit](https://github.com/square/okio/commit/466882f812dd41d66179965a37f42a11b8c17cc7) aught to do it.


```
git pull
git cherry-pick 466882f812dd41d66179965a37f42a11b8c17cc7
```

PR #1809, opened 2026-05-12
Generalize Sink/Source API to accept BufferedSource/BufferedSink

### LINKED ISSUE #1808: Source.read should accept a BufferedSink not Buffer, and Sink.write should accept a BufferedSource
I just implemented a Source. (I love how simple that was.) But I think a Source implementation should not require a Source parameter, which Buffer is. [Buffer][b] implements both `BufferedSource` and `BufferedSink`, and so it should not break anything (that isn't possibly already subtly broken) to change the method signature of [Source.read][r] to accept a BufferedSink. Buffers passed into Source.read will still work. But also, I can implement a Sink that I can pass into `read` without having to wrap it in a boilerplate abstraction that throws exceptions for unnecessary read implementations on a Sink. I get that everything kind of revolves around that fantastic Buffer implementation, but a Source shouldn't need to read from what it has been given to write to, should it?

Similarly, [Sink.write][w] shouldn't need to write to a parameter it has been given to read from, should it?

 [r]: https://square.github.io/okio/3.x/okio/okio/okio/-source/index.html#-934028336%2FFunctions%2F439404261
 [w]: https://square.github.io/okio/3.x/okio/okio/okio/-sink/index.html#143355683%2FFunctions%2F-1473893935
 [b]: https://square.github.io/okio/3.x/okio/okio/okio/-buffer/index.html

travisfw:
Change `Sink.write(source: Buffer, byteCount: Long)` to accept `BufferedSource` and `Source.read(sink: Buffer, byteCount: Long)` to accept `BufferedSink`. Since `Buffer` implements both interfaces this is backward-compatible, but it removes the unnecessary constraint that callers must use a `Buffer` specifically.

Fixes #1808

### DIFF
--- a/okio-nodefilesystem/src/commonMain/kotlin/okio/FileSink.kt
+++ b/okio-nodefilesystem/src/commonMain/kotlin/okio/FileSink.kt
@@ -20,9 +20,9 @@ internal class FileSink(
 ) : Sink {
   private var closed = false
 
-  override fun write(source: Buffer, byteCount: Long) {
+  override fun write(source: BufferedSource, byteCount: Long) {
     require(byteCount >= 0L) { "byteCount < 0: $byteCount" }
-    require(source.size >= byteCount) { "source.size=${source.size} < byteCount=$byteCount" }
+    require(source.buffer.size >= byteCount) { "source.size=${source.buffer.size} < byteCount=$byteCount" }
     check(!closed) { "closed" }
 
     val data = source.readByteArray(byteCount)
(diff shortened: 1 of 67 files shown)


travisfw:
oops. The `travisfw` branch should not be in this PR. I don't know how that got associated with it. :(
I made this PR from the `1808` branch. `travisfw` branch should not be showing up here. I changed the group in `travisfw` so I could use my downstream edits separate from upstream.
Edit: fixed. I had committed to the wrong branch. Sorry for the noise. 1808 is on the right commit at this time.

JakeWharton:
This would not be a binary-compatible change to make.

JakeWharton:
https://github.com/square/okio/issues/1808#issuecomment-4545501605

PR #1812, opened 2026-05-24
Add `ByteString.equals(other, constantTime)` overload

### LINKED ISSUE #1797: Add `ByteString.equals(other: ByteString, constantTime: Boolean)` overload

SAY-5:
Adds a constant-time `ByteString.equals(other, constantTime)` overload for timing-safe comparison of secrets like hashes and MACs, as requested in #1797. After a length check it XORs every byte and never short-circuits, so the running time does not depend on where the byte strings differ. Implemented in `commonMain` so it works on all targets, and tested across the basic and segmented byte string variants.

Closes #1797. Closes #1799

### DIFF
--- a/okio/src/commonMain/kotlin/okio/internal/ByteString.kt
+++ b/okio/src/commonMain/kotlin/okio/internal/ByteString.kt
@@ -234,6 +234,16 @@ internal inline fun ByteString.commonEquals(other: Any?): Boolean {
   }
 }
 
+internal fun ByteString.commonEqualsConstantTime(other: ByteString): Boolean {
+  if (other === this) return true
+  if (other.size != size) return false
+  var result = 0
+  for (i in 0 until size) {
+    result = result or (this[i].toInt() xor other[i].toInt())
+  }
+  return result == 0
+}
+
 @Suppress("NOTHING_TO_INLINE")
 internal inline fun ByteString.commonHashCode(): Int {
   val result = hashCode
--- a/okio/src/jvmMain/kotlin/okio/ByteString.kt
+++ b/okio/src/jvmMain/kotlin/okio/ByteString.kt
@@ -190,7 +190,8 @@ internal actual constructor(
 
   actual override fun equals(other: Any?) = commonEquals(other)
 
-  actual fun equals(other: ByteString, constantTime: Boolean) = commonEqualsConstantTime(other)
+  actual fun equals(other: ByteString, constantTime: Boolean) =
+    if (constantTime) commonEqualsConstantTime(other) else this == other
 
   actual override fun hashCode() = commonHashCode()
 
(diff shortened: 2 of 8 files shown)


okio/src/jvmMain/kotlin/okio/ByteString.kt
```diff
@@ -189,6 +190,8 @@ internal actual constructor(
 
   actual override fun equals(other: Any?) = commonEquals(other)
 
+  actual fun equals(other: ByteString, constantTime: Boolean) = commonEqualsConstantTime(other)
```
JakeWharton: This does not handle `constantTime=false` correctly.

SAY-5: Fixed in a4e8c40. The overload now branches on the flag and falls back to the regular equals for constantTime=false, with the false path covered in ByteStringTest.

okio/src/commonMain/kotlin/okio/internal/ByteString.kt
```diff
@@ -234,6 +234,16 @@ internal inline fun ByteString.commonEquals(other: Any?): Boolean {
   }
 }
 
+internal fun ByteString.commonEqualsConstantTime(other: ByteString): Boolean {
+  if (other === this) return true
+  if (other.size != size) return false
+  var result = 0
+  for (i in 0 until size) {
+    result = result or (this[i].toInt() xor other[i].toInt())
```
JakeWharton, line 242: Is there a reason you chose to use 0 as the true-y value and `or` the comparisons as opposed to using 1 and `and`? The result is the same, but this version is unintuitive to me because it's not how I think about the comparisons being done.

JakeWharton (approved):
Looks good! Just one question.

okio/src/commonMain/kotlin/okio/internal/ByteString.kt
SAY-5: It's the standard constant-time idiom: XOR each byte pair (0 when equal) and OR the results, so result stays 0 only if every pair matched and becomes non-zero the moment any byte differs, with a single comparison at the very end. Going with 1 and AND would need a per-byte equality test (this[i] == other[i]), which reintroduces a branch/comparison inside the loop that the OR-of-XOR form avoids. It mirrors crypto/subtle.ConstantTimeCompare and CRYPTO_memcmp. Happy to add a short comment to that effect if it helps readability.

JakeWharton: You don't need branching. I'm basically distributing a bitwise not to both sides of the final equality and reducing. But I can see that the resulting assembly operations contain the extra flip because xnor is only available in huge SIMD lanes and such.

PR #1821, opened 2026-06-28
Pass preopen-relative paths to WASI `path_*` functions

### LINKED ISSUE #1746: Latest NodeJS versions no longer accept absolute paths as WASI preview1 functions' arguments
After https://github.com/nodejs/node/pull/59791, NodeJS no longer accept absolute paths as arguments to WASI P1 functions.

`WasiFileSystem` does not convert paths from absolute to relative before passing them to WASI functions (https://github.com/square/okio/blob/83c8481334f3cae144f43cc7d7f256531b8c0340/okio-wasifilesystem/src/wasmWasiMain/kotlin/okio/WasiFileSystem.kt#L471), so it is prone to this problem.

eyupcanakman:
`WasiFileSystem` passes absolute paths to the WASI `path_*` functions, which resolve their argument against a preopened directory. After nodejs/node#59791, Node.js no longer accepts absolute paths there, so every file operation fails with `notcapable`.

Each path is now made relative to its preopen before the call, matching `createSymlink`.

okio's CI uses the Node bundled with Kotlin 2.2.21 (Node 22), which accepts absolute paths, so tests pass either way. On Node 26.3.0 the fix drops `notcapable` failures from 169 to 2. The other two are a separate Node restriction on `..` in symlink targets.

Closes #1746

### DIFF
--- a/okio-wasifilesystem/src/wasmWasiMain/kotlin/okio/WasiFileSystem.kt
+++ b/okio-wasifilesystem/src/wasmWasiMain/kotlin/okio/WasiFileSystem.kt
@@ -142,8 +142,8 @@ object WasiFileSystem : FileSystem() {
   override fun metadataOrNull(path: Path): FileMetadata? {
     withScopedMemoryAllocator { allocator ->
       val returnPointer = allocator.allocate(64)
-      val preopen = preopenForPath(path) ?: return null
-      val (pathAddress, pathSize) = allocator.write(preopen.relativePath(path))
+      val (preopen, relativePath) = preopenForPath(path) ?: return null
+      val (pathAddress, pathSize) = allocator.write(relativePath)
 
       val errno = path_filestat_get(
         fd = preopen.fd,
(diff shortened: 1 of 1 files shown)


swankjesse (commented):
Probably worth bumping our Node version to this new more-strict one.

okio-wasifilesystem/src/wasmWasiMain/kotlin/okio/WasiFileSystem.kt
swankjesse: I wonder if this function should return a `Pair<Preopen, String>?`  with the second value being the relative path. You always need to do both things together, right?

swankjesse (commented):
This is great.

eyupcanakman:
Done, `preopenForPath` returns the pair now. Before bumping Node though, note the stricter Node also rejects the `..` symlink targets (`canonicalizeReturnsShallowerPath`, `symlinkCanBeRelativeWithDotDots`) at `path_symlink`, so those two go red on it, separate from the absolute-path fix here.

swankjesse (approved):

PR #1823, opened 2026-07-19
Make FixedLengthSource public API

### LINKED ISSUE #1208: Make FixedLengthSource public API
I know it's trivially reimplementable, but it still feels frustrating to reinvent the wheel when there's a perfectly good wheel that's internal. (I actually just copied the class into my own project, so there are two now, one internal to okio, and one in my project.)

swankjesse:
Closes: https://github.com/lysine-dev/okio/issues/1208


okio/src/commonMain/kotlin/okio/Okio.kt
```diff
@@ -78,3 +79,76 @@ inline fun <T : Closeable?, R> T.use(block: (T) -> R): R {
...
+ *
+ * @param throwIfSourceIsLonger true to also throw if this has more than [byteCount] bytes.
+ *   This works by attempting to read more than [byteCount] bytes.
+ *
+ * @throws [EOFException] if this returns fewer than [byteCount] bytes.
+ */
+@JvmOverloads
+fun Source.limit(
+  byteCount: Long,
+  throwIfSourceIsLonger: Boolean = false,
```
JakeWharton, line 96: This is sufficient for OkHttp's needs? I thought it needed to know if you exhausted the limit or not.

okio/src/commonMain/kotlin/okio/Okio.kt
```diff
@@ -78,3 +79,76 @@ inline fun <T : Closeable?, R> T.use(block: (T) -> R): R {
...
+      if (remainingByteCount == 0L) return -1L
+      throw EOFException("expected ${this.byteCount} bytes but got $bytesReceived")
+    }
+
+    bytesReceived += result
+
+    val beyondLimitByteCount = bytesReceived - this.byteCount
+    if (beyondLimitByteCount > 0) {
+      // If we received bytes beyond the limit, don't return them to the caller.
+      sink.truncateToSize(sink.size - beyondLimitByteCount)
```
JakeWharton, line 141: This would only happen from a misbehaving source, right?

okio/src/commonMain/kotlin/okio/Okio.kt
```diff
@@ -78,3 +79,76 @@ inline fun <T : Closeable?, R> T.use(block: (T) -> R): R {
...
+      throw IOException("expected ${this.byteCount} bytes but got $bytesReceived")
+    }
+
+    return result
+  }
+
+  private fun Buffer.truncateToSize(newByteCount: Long) {
```
JakeWharton, line 148: You can't `UnsafeCursor` and walk back the counts/tail segments? I guess it'll never happen in a well behaved system so who cares the performance...

JakeWharton (approved):

okio/src/commonMain/kotlin/okio/Okio.kt
swankjesse: I looked into OkHttp's two best candidates...

HTTP/1 framing could use this, but it's overall worse because the framing also does other things like skip the remainder if closed prematurely. 

DNS messages are a good candidate and can use the simplest limit API.

okio/src/commonMain/kotlin/okio/Okio.kt
swankjesse: Kinda sorta 

In Okio zip file handling we expect a zip file's decompressed size to be consistent with the metadata's promised decompressed size. They could be inconsistent if the zip writer was buggy or malicious. 

In such cases we'll detect that the actual size was larger than the expected size and throw. 

But we won't return the excess data to the caller.

okio/src/commonMain/kotlin/okio/Okio.kt
swankjesse: Yeah, exactly. This is Okio itself, we could also just add a function to Buffer, trimToSize(). Not today though

PR #1825, opened 2026-07-21
New API, BufferedSink.utf8Appendable()

### LINKED ISSUE #1441: Integration with kotlin.text.Appendable
Should okio offer an `BufferedSink` -> `Appendable` adapter for use with `kotlinpoet` and other libraries that do not accept output streams?

```kotlin
fileSystem.write(path) {
  // Along the lines of BufferedSink#outputStream().
  fileSpec.writeTo(this.appendable())
}
```

### LINKED ISSUE #1615: Make BufferedSink implements Appendable
While `Writer` of Java IO implements `Appendable`, it will be good to have it in BufferedSink

swankjesse:
This is unreasonably painful because I'm committed to writing UTF-8 to the buffer, but the input string is UTF-16 and we might need to save a high surrogate for a follow-up call.

Closes: https://github.com/lysine-dev/okio/issues/1615
Closes: https://github.com/lysine-dev/okio/issues/1441

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Buffer.kt
+++ b/okio/src/jvmMain/kotlin/okio/Buffer.kt
@@ -27,6 +27,7 @@ import java.security.InvalidKeyException
 import java.security.MessageDigest
 import javax.crypto.Mac
 import javax.crypto.spec.SecretKeySpec
+import okio.internal.combineSurrogates
 import okio.internal.commonClear
 import okio.internal.commonClose
 import okio.internal.commonCompleteSegmentByteCount
@@ -72,6 +73,8 @@ import okio.internal.commonWriteLong
 import okio.internal.commonWriteShort
 import okio.internal.commonWriteUtf8
 import okio.internal.commonWriteUtf8CodePoint
+import okio.internal.isHighSurrogate
+import okio.internal.isLowSurrogate
 
 actual class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
   @JvmField internal actual var head: Segment? = null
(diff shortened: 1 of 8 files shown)


swankjesse:
Biggest potential hazard is that these two do different things:

```kotlin
val buffer = Buffer()
buffer.appendable().write(s1)
buffer.appendable().write(s2)
```


```kotlin
val buffer = Buffer()
buffer.appendable().apply {
  write(s1)
  write(s2)
}
```

This is because if `s1` ends with a high surrogate and `s2` starts with a low surrogate, the two can be glued back together in the second listing. But the first listing doesn’t know about the saved high surrogate so a character is lost.

okio/src/jvmMain/kotlin/okio/Buffer.kt
```diff
@@ -102,6 +105,70 @@ actual class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
+        if (savedHighSurrogate != 0) {
+          if (string[beginIndex].code.isLowSurrogate) {
+            writeUtf8CodePoint(combineSurrogates(savedHighSurrogate, string[beginIndex++].code))
+          } else {
+            writeByte('?'.code) // Our high surrogate was unmatched.
+          }
+          savedHighSurrogate = 0
+        }
+
+        // If the last character is a high surrogate, save it for later.
```
saket, line 139: What happens when the sink is flushed/closed before the matching low surrogate shows up? The saved value will be dropped. Is that something you want to explicitly support?

okio/src/jvmMain/kotlin/okio/Buffer.kt
```diff
@@ -102,6 +105,70 @@ actual class Buffer : BufferedSource, BufferedSink, Cloneable, ByteChannel {
...
+      override fun append(
+        charSequence: CharSequence?,
+        beginIndex: Int,
+        endIndex: Int,
+      ) = apply {
+        val string = charSequence?.toString() ?: "null"
+
+        require(beginIndex >= 0) { "beginIndex < 0: $beginIndex" }
+        require(endIndex >= beginIndex) { "endIndex < beginIndex: $endIndex < $beginIndex" }
+        require(endIndex <= string.length) { "endIndex > string.length: $endIndex > ${string.length}" }
```
saket, line 124: nit: do you want to use `IndexOutOfBoundsException` for these checks?

saket (approved):

okio/src/jvmMain/kotlin/okio/Buffer.kt
saket: Oh I see you've mentioned the same problem in your [other comment](https://github.com/lysine-dev/okio/pull/1825#issuecomment-5028740046):

> This is because if s1 ends with a high surrogate and s2 starts with a low surrogate, the two can be glued back together in the second listing.

okio/src/jvmMain/kotlin/okio/Buffer.kt
swankjesse: Yeah, it’s a good idea, but if I switch to it I should switch in 10 other places also. Follow up!

PR #1826, opened 2026-07-21
Optionally ignore whitespace when decoding hex

### LINKED ISSUE #1433: Consider `ignoreWhitespace=true` parameter to `decodeHex`
Hex often shows up with visual whitespace (spaces, newlines, tabs). I can preprocess it, and I do, but I've just copied that helper function to my third project.

I think this would be a pragmatic inclusion.

One caveat is that this would probably be a conditional that selected two mostly different implementations of the decoding due to how the current one works.

swankjesse:
Closes: https://github.com/lysine-dev/okio/issues/1433

### DIFF
--- a/okio/src/commonMain/kotlin/okio/ByteString.kt
+++ b/okio/src/commonMain/kotlin/okio/ByteString.kt
@@ -218,5 +218,14 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
     /** Decodes the hex-encoded bytes and returns their value a byte string. */
     @JvmStatic
     fun String.decodeHex(): ByteString
+
+    /**
+     * Decodes the hex-encoded bytes and returns their value a byte string.
+     *
+     * @param ignoreWhitespace true to skip ASCII whitespace characters (space, tab, carriage
+     *   return and line feed). Otherwise, this throws if whitespace is present otherwise.
+     */
+    @JvmStatic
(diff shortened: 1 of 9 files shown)


okio/src/commonTest/kotlin/okio/ByteStringHexTest.kt
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2018 Square, Inc.
```
swankjesse, line 2: This is split off ByteStringTest

okio/src/commonMain/kotlin/okio/ByteString.kt
```diff
@@ -218,5 +218,14 @@ internal constructor(data: ByteArray) : Comparable<ByteString> {
     /** Decodes the hex-encoded bytes and returns their value a byte string. */
     @JvmStatic
     fun String.decodeHex(): ByteString
+
+    /**
+     * Decodes the hex-encoded bytes and returns their value a byte string.
+     *
+     * @param ignoreWhitespace true to skip ASCII whitespace characters (space, tab, carriage
+     *   return and line feed). Otherwise, this throws if whitespace is present otherwise.
```
JakeWharton:
```suggestion
     *   return and line feed). Otherwise, this throws if whitespace is present.
```

JakeWharton (approved):

PR #1841, opened 2026-08-05
fix: apply Throttler.bytesPerSecond rate changes immediately (#1822)

### LINKED ISSUE #1822: [BUG] Throttler.bytesPerSecond() does not apply rate changes immediately
When increasing the rate limit using `Throttler.bytesPerSecond()`, the new limit is not applied immediately if bytes have already been allocated into the future.

For example, if the throttler is configured with a very low rate (e.g. 1 KB/s) and the limit is then increased, reads continue to wait according to the previous allocation schedule before the higher rate takes effect.

This makes `Throttler` effectively unusable for applications such as download managers, where users expect speed limit changes to take effect immediately.

Is the current behavior intentional?

If preserving the current behavior is important, would you consider making it optional to reset the allocation schedule when changing the rate? I'd be happy to submit a PR if that's something you'd be interested in.

arimu1:
## Summary

`Throttler.bytesPerSecond()` updated the configured rate but left any allocation already scheduled into the future under the previous rate. After a slow rate had pushed `allocatedUntil` far ahead, increasing the rate still waited on that old schedule — so throughput changes did not take effect promptly (e.g. download managers).

On rate reconfiguration, mark the allocation schedule for reset. The next `byteCountOrWaitNanos` / `take` restarts at the current time so the new throughput applies immediately, then normal sustained throttling continues at the new rate.

Fixes #1822

## Test plan

- [x] Added `ThrottlerTakeTest.bytesPerSecondIncreaseAppliesImmediately` (deterministic fake clock, same style as existing take tests)
- [x] `./gradlew :okio:jvmTest --tests 'okio.ThrottlerTakeTest' --tests 'okio.ThrottlerTest'` (Temurin 21): all `ThrottlerTakeTest` cases **PASSED** (6/6 including new); `ThrottlerTest` skipped (`@Ignore` flaky)

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/Throttler.kt
+++ b/okio/src/jvmMain/kotlin/okio/Throttler.kt
@@ -49,6 +49,13 @@ class Throttler internal constructor(
   private var waitByteCount: Long = 8 * 1024 // 8 KiB.
   private var maxByteCount: Long = 256 * 1024 // 256 KiB.
 
+  /**
+   * When true, the next call to [byteCountOrWaitNanos] discards any future allocation and
+   * restarts the schedule at the supplied `now`. This makes [bytesPerSecond] rate changes take
+   * effect immediately instead of waiting out bytes allocated under a previous rate.
+   */
+  private var resetAllocation = false
+
   val lock: ReentrantLock = ReentrantLock()
   val condition: Condition = lock.newCondition()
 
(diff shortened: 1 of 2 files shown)


JakeWharton:
Hi! Thanks for the PR, but we are imminently about to ban LLM descriptions and comments, as well as LLM-authored code. Your PR description is pretty clearly LLM authored and that makes me think the code is too (I didn't review the diff).

You are welcome to use whatever tools you want for investing problems, but we expect all code to be written by hand along with comments and descriptions. If you can meet these requirements, you're welcome to make a new PR with a fix.

PR #1842, opened 2026-08-06
Enforce that Socket.cancel() does not throw

swankjesse:
It doesn't throw in practice for java.net.Socket sockets, but that type is documented to throw IOException.

Also recover from a Conscrypt bug, as currently implemented in OkHttp.

### DIFF
--- a/okio/src/jvmMain/kotlin/okio/internal/DefaultSocket.kt
+++ b/okio/src/jvmMain/kotlin/okio/internal/DefaultSocket.kt
@@ -42,7 +42,17 @@ internal class DefaultSocket(val socket: JavaNetSocket) : Socket {
   override val sink: Sink = SocketSink()
 
   override fun cancel() {
-    socket.close()
+    try {
+      socket.close()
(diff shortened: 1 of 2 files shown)


okio/src/jvmMain/kotlin/okio/internal/DefaultSocket.kt
```diff
@@ -42,7 +42,17 @@ internal class DefaultSocket(val socket: JavaNetSocket) : Socket {
   override val sink: Sink = SocketSink()
 
   override fun cancel() {
-    socket.close()
+    try {
+      socket.close()
+    } catch (rethrown: RuntimeException) {
+      if (rethrown.message == "bio == null") {
```
swankjesse, line 48: https://github.com/lysine-dev/okhttp/blob/main/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/-UtilJvm.kt#L285

JakeWharton (approved):

PR #1845, opened 2026-08-12
Fix `FileHandle.resize(0L)` on mingwX64

### LINKED ISSUE #1844: FileHandle.resize(0L) throws after successful SetFilePointer on mingwX64
## Description

Okio 3.18.1 consistently throws from `FileHandle.resize(0L)` on the Kotlin/Native `mingwX64` target when resizing a non-empty file:

```text
okio.IOException: The operation completed successfully.
```

This is not a transient file-system or concurrency failure. `WindowsFileHandle.protectedResize()` treats a `SetFilePointer()` return value of `0` as failure, but Windows documents that a successful move to file position `0` returns `0` because the return value is the low-order `DWORD` of the new position.

## Minimal reproduction

https://github.com/czp3009/okio-mingw-file-handle-resize-reproduction

Core code:

```kotlin
fun main() {
    val fileSystem = FileSystem.SYSTEM
    val path = "test".toPath()

    fileSystem.sink(path).buffer().use { sink ->
        sink.writeUtf8("seed")
    }

    fileSystem.openReadWrite(path, mustExist = true).use { handle ->
        handle.resize(0L)
    }

    check(fileSystem.metadata(path).size == 0L)
}
```

The Windows native process terminates with an uncaught exception at `handle.resize(0L)`:

```text
Uncaught Kotlin exception: okio.IOException: The operation completed successfully.
```

The final size assertion is not reached. The exact formatted message can depend on the thread's last-error value and the Windows language, but the erroneous exception is deterministic.

## Root cause

The affected code is [`WindowsFileHandle.protectedResize()`](https://github.com/square/okio/blob/parent-3.18.1/okio/src/mingwX64Main/kotlin/okio/WindowsFileHandle.kt#L139-L154):

```kotlin
val movePointerResult = SetFilePointer(
    hFile = file,
    lDistanceToMove = size.toInt(),
    lpDistanceToMoveHigh = distanceToMoveHigh.ptr,
    dwMoveMethod = FILE_BEGIN.toUInt(),
)
if (movePointerResult == 0U) {
    throw lastErrorToIOException()
}
```

According to Microsoft's [`SetFilePointer` documentation](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-setfilepointer):

- On success, the return value is the low-order `DWORD` of the new file-pointer position.
- On failure, the return value is `INVALID_SET_FILE_POINTER`, not zero.
- Because `INVALID_SET_FILE_POINTER` can itself be a valid low-order position, failure with a non-null high-order pointer must be checked as `result == INVALID_SET_FILE_POINTER && GetLastError() != NO_ERROR`.

For `resize(0L)`, the call succeeds and returns the new low-order position, which is `0`. The `movePointerResult == 0U` condition therefore enters the error branch after a successful API call. Okio throws before reaching `SetEndOfFile`.

The observed message is also explained by the Windows API contract:

- [`GetLastError`](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror) should only be read when the preceding function's return value indicates that error information is valid. A successful function does not necessarily clear the thread's last-error value.
- [`ERROR_SUCCESS`](https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499-) is value `0`, with the system message `The operation completed successfully.`
- Okio's [`lastErrorToIOException()`](https://github.com/square/okio/blob/parent-3.18.1/okio/src/mingwX64Main/kotlin/okio/Windows.kt#L32-L53) formats that value with `FormatMessageW` and wraps the result in `IOException`.

The same incorrect condition also affects target positions whose low-order 32 bits are zero. `resize(0L)` is the smallest stable reproducer.

## Suggested minimal fix

The failure condition should follow the documented contract instead of comparing the result with zero:

```kotlin
if (movePointerResult == 0xffffffffU && GetLastError() != 0U) {
    throw lastErrorToIOException()
}
```

czp3009:
`SetFilePointer()` returns the low-order file position, so zero is valid when resizing a file to zero bytes. Treat only `INVALID_SET_FILE_POINTER` with a non-`NO_ERROR` last error as failure.

Resolve https://github.com/lysine-dev/okio/issues/1844

### DIFF
--- a/okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
+++ b/okio-testing-support/src/commonMain/kotlin/okio/AbstractFileSystemTest.kt
@@ -1736,6 +1736,20 @@ abstract class AbstractFileSystemTest(
     }
   }
 
+  @Test fun fileHandleResizeToZero() {
+    val path = base / "file-handle-resize-to-zero"
+    fileSystem.openReadWrite(path).use { handle ->
+
+      handle.sink().buffer().use { sink ->
+        sink.writeUtf8("abcde")
+      }
+
+      handle.resize(0L)
+
+      assertEquals(0L, handle.size())
+    }
+  }
+
   @Test fun fileHandleResizeLarger() {
     val path = base / "file-handle-resize-larger"
     fileSystem.openReadWrite(path).use { handle ->
--- a/okio/src/mingwX64Main/kotlin/okio/WindowsFileHandle.kt
+++ b/okio/src/mingwX64Main/kotlin/okio/WindowsFileHandle.kt
@@ -30,7 +30,9 @@ import platform.windows.FlushFileBuffers
 import platform.windows.GetFileSizeEx
 import platform.windows.GetLastError
 import platform.windows.HANDLE
+import platform.windows.INVALID_SET_FILE_POINTER
 import platform.windows.LARGE_INTEGER
+import platform.windows.NO_ERROR
 import platform.windows.ReadFile
 import platform.windows.SetEndOfFile
 import platform.windows.SetFilePointer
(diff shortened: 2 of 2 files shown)


czp3009:
Is there a project manager who can review and merge my PR? If there are any prerequisites for submitting a PR, please let me know—I couldn’t find any specific requirements on the project website. okio is a great project, but this bug is currently holding me up, and I’d like to start using the fixed version as soon as possible.

Issue #1846, opened 2026-08-14
Test Balloon migration


JakeWharton:
We have decided to migrate from Burst to Test Balloon.

This will be done in multiple steps. PRs for individual steps are welcome.

* [x] Switch from JUnit 4 to JUnit 5 engine for JVM
  * #1862 
* [ ] Port tests from Burst to Test Balloon

PR #1865, opened 2026-08-30
Migrate Burst to TestBalloon

wilburx9:
Migrated Burst to TestBalloon

Part 2 of #1846


gradle/libs.versions.toml
```diff
@@ -5,6 +5,7 @@ kotlin = "2.2.21"
 # Set to lower version than KGP version to maximize compatibility
 kotlinCoreLibrariesVersion = "2.1.21"
 ktlint = "0.48.2"
+testBalloon = "1.0.1-K2.2.21"
```
JakeWharton: We can use the 1.1 rc

okio-testing-support/src/commonMain/kotlin/okio/TestDirectory.kt
```diff
@@ -15,22 +15,28 @@
...
-class TestDirectory(
-  val fileSystem: FileSystem,
-  val temporaryDirectory: Path = FileSystem.SYSTEM_TEMPORARY_DIRECTORY,
-) : TestInterceptor {
-  lateinit var path: Path
-    private set
+fun TestSuiteScope.testDirectory(
+  fileSystem: FileSystem,
+  temporaryDirectory: Path = FileSystem.SYSTEM_TEMPORARY_DIRECTORY,
+) = testFixture { fileSystem.createTestPath(temporaryDirectory) }
```
OliverO2, line 26: Does this get deleted after use? If not and if that makes sense, [here is an example strategy](https://infix-de.github.io/testBalloon/latest/how-to/effective-testing/#use-a-temporary-directory-and-keep-it-on-failures).

wilburx9: @OliverO2, thanks for the suggestion.  I believe I tried that approach, but [some tests were failing because of the recursive deletion](https://github.com/lysine-dev/okio/actions/runs/33416583038/job/99568564363), so I had to remove it. Or do you mean only deleting for non-CI builds?

OliverO2: I was thinking of this as a general practice: I'd expect resources like temporary files/directories to have a lifetime tied to the scope (test suite) they are used in. A test fixture automatically takes care of this, closing its resource at the end of its `TestSuiteScope`.

This should work regardless of CI. The CI special case was just an example, where the contents of resources would only be examined locally (non-CI), and only for failing tests.

One thing I could imagine is: If you are using concurrent invocation and the same file path is used in concurrently executing suites, these could collide. Could this be the case?

OliverO2: Seems like a bug in `WasiFileSystem`:

If I clean up the temporary directory like this:

```kotlin
fun TestSuiteScope.testDirectory(
  fileSystem: FileSystem,
  temporaryDirectory: Path = FileSystem.SYSTEM_TEMPORARY_DIRECTORY,
) = testFixture { fileSystem.createTestPath(temporaryDirectory) } closeWith {
    fileSystem.deleteRecursively(this)
}
```

the failing test is `okio.WasiTest↘canonicalizeAbsolutePathWithSymlinkCycle`.

This test creates a cycle of three symlinks. When trying to delete the parent directory via `fileSystem.deleteRecursively()`, the function `WasiFileSystem.pathOpen` gets called and tries to open one of the symlinks. This fails with errno 32 (loop).

Seems like there is no "follow symlinks" option in the functions that list directory entries. Maybe either `pathOpen` or the `list` function calling it should (conditionally) check for a symlink first before trying to open it. Otherwise, with a symlink cycle it is bound to fail as expected and `fileSystem.deleteRecursively()` can never succeed in this case.

Println debugging trace:
```
>>> commonDeleteRecursively on /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4
>>> collectRecursively start: /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4
>>> collectRecursively before children
>>> listOrNull(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) start
>>> list(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) start
>>> pathOpen(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) start
>>> pathOpen(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) path_open errno=0
>>> list(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) after pathOpen
>>> list(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4) start
>>> list(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4): errno=0
>>> collectRecursively children: [/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper, /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/rock, /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/scissors]
>>> collectRecursively symlink: /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4
>>> collectRecursively child: /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper
>>> collectRecursively start: /tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper
>>> collectRecursively before children
>>> listOrNull(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper) start
>>> list(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper) start
>>> pathOpen(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper) start
>>> pathOpen(/tmp/test-f4e97d1653c0ae50da5d749cf8e8bcf4/paper) path_open errno=32
loop
ErrnoException: loop
```

wilburx9: Interesting. I'd suggest we open a separate issue to investigate this. In the interim, I can bring back `closeWith` but disable folder deletion with a flag passed to `testDirectory()`. Then explicitly disable it for the failing test.

Thoughts? @JakeWharton @OliverO2

OliverO2: Created #1874 and provided a fix in https://github.com/wilburx9/okio/pull/1.

Feel free to cherry-pick and/or change as required.

JakeWharton: A flag for now is fine, if only to get it landed sooner rather than later!

PR #1869, opened 2026-09-04
Fix an unintended behavior change with base64 padding

### LINKED ISSUE #1868: Binary compatibility overload introduced in 3.18.1 for base64() in appleMain changes the previous behavior
Prior to version 3.18.0, the `base64()` function in appleMain used [commonBase64()](https://github.com/lysine-dev/okio/blob/parent-3.17.0/okio/src/appleMain/kotlin/okio/ByteString.kt#L77) which always applied padding.

As of version 3.18.1, there is a binary-compatibility overload that [defaults the padding to "false"](https://github.com/lysine-dev/okio/blob/parent-3.18.1/okio/src/appleMain/kotlin/okio/ByteString.kt#L82), but that seems like the opposite of the previous behavior.

My issue is that, while I don't use okio directly, it's a dependency of another library I use, [apollo-kotlin](https://github.com/apollographql/apollo-kotlin). Since the default behavior changed, some GraphQL operations no longer work.

I can make do with 3.17.0 for now but is there a reason the compatible overload was defaulted to `false`?

swankjesse:
I inadvertently changed behavior in 3.18.1.

Closes: https://github.com/lysine-dev/okio/issues/1868

### DIFF
--- a/okio/src/appleMain/kotlin/okio/ByteString.kt
+++ b/okio/src/appleMain/kotlin/okio/ByteString.kt
@@ -79,12 +79,12 @@ internal actual constructor(
   actual open fun base64(includePadding: Boolean): String = commonBase64(includePadding = includePadding)
 
   @Deprecated(message = "for binary compatibility", level = DeprecationLevel.HIDDEN)
-  fun base64(): String = commonBase64(includePadding = false)
+  fun base64(): String = commonBase64(includePadding = true)
 
   actual open fun base64Url(includePadding: Boolean): String = commonBase64Url(includePadding = includePadding)
 
   @Deprecated(message = "for binary compatibility", level = DeprecationLevel.HIDDEN)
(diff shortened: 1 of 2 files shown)


okio/src/appleMain/kotlin/okio/ByteString.kt
```diff
@@ -79,12 +79,12 @@ internal actual constructor(
   actual open fun base64(includePadding: Boolean): String = commonBase64(includePadding = includePadding)
 
   @Deprecated(message = "for binary compatibility", level = DeprecationLevel.HIDDEN)
-  fun base64(): String = commonBase64(includePadding = false)
+  fun base64(): String = commonBase64(includePadding = true)
```
swankjesse, line 82: Unfortunately it’s particularly annoying to write a test case for this, which is why we didn’t catch it the first time around.

JakeWharton (approved):

